PR #26, opened 2019-11-28
revived Blueprint plugin and added globals push to js

fefanto:
# problem : 
 - Blueprint plugin example does not compile with latest master : assuming it's been left behind if compared to gain plugin.
 - getDuktapeContext API (used to send globals from juce to js) got scrapped in last commit. Need it restored (unless there's a new/better way coming , but I'd love this even temporarily for retrocompatibility).

# Main Changes
- getDuktapeContext restored (in the EcmaScriptEngine class) - help view changed to test it: instead of remnant help shows the version sent from juce.
- lambda signature of registerNativeMethod aligned to latest flavor in Blueprint plugin example
- commented out the //appRoot.enableHotkeyReloading(); in the Bluepring plugin example editor : no idea if this has to come back or not.

### DIFF
--- a/blueprint/core/blueprint_EcmascriptEngine.cpp
+++ b/blueprint/core/blueprint_EcmascriptEngine.cpp
@@ -347,9 +347,4 @@ namespace blueprint
         return result;
     }
 
-    duk_context* EcmascriptEngine::getDuktapeContext()
-    {
-        return ctx;
-    }
-
 }
--- a/blueprint/core/blueprint_EcmascriptEngine.cpp
+++ b/blueprint/core/blueprint_EcmascriptEngine.cpp
@@ -347,4 +347,9 @@ namespace blueprint
         return result;
     }
 
+    duk_context* EcmascriptEngine::getDuktapeContext()
+    {
+        return ctx;
+    }
+
 }
--- a/examples/BlueprintPlugin/Source/PluginEditor.cpp
+++ b/examples/BlueprintPlugin/Source/PluginEditor.cpp
@@ -19,7 +19,6 @@ BlueprintPluginAudioProcessorEditor::BlueprintPluginAudioProcessorEditor (Bluepr
 
     addAndMakeVisible(appRoot);
     appRoot.evaluate(sourceDir.getChildFile("ui/build/js/main.js").loadFileAsString());
-    //appRoot.enableHotkeyReloading();
     
     appRoot.engine.registerNativeMethod(
         "setParameterValueNotifyingHost",
@@ -27,7 +26,7 @@ BlueprintPluginAudioProcessorEditor::BlueprintPluginAudioProcessorEditor (Bluepr
             auto* self = reinterpret_cast<BlueprintPluginAudioProcessorEditor*>(stash);
             const juce::String& paramId = args.arguments[0].toString();
             const double value = args.arguments[1];
-           
+
             DBG("Setting " << paramId << " to " << value);
 //            if (auto* parameter = self->processor.getValueTreeState().getParameter(paramId))
 //                parameter->setValueNotifyingHost(value);
@@ -39,10 +38,7 @@ BlueprintPluginAudioProcessorEditor::BlueprintPluginAudioProcessorEditor (Bluepr
     
     
     // Globals in the js env
-    auto* ctx = appRoot.engine.getDuktapeContext();
-    duk_push_global_object(ctx);
-    duk_push_string(ctx, JucePlugin_VersionString);
-    duk_put_prop_string(ctx, -2, "__VERSION__");
+    appRoot.engine.registerNativeProperty("__VERSION__", JucePlugin_VersionString);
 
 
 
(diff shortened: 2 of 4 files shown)


nick-thompson:
Thanks for this @fefanto! Sorry I'm just getting around to it, I'll review the code in detail now

examples/BlueprintPlugin/Source/ui/src/App.js
```diff
@@ -79,8 +79,11 @@ class App extends Component {
           <Chrome
             {...styles.chrome}
             onToggleSettings={this._onToggleSettings}
-            onToggleHelp={this._onToggleHelp}>
-            <Help {...styles.contentContainer} />
+            onToggleHelp={this._onToggleHelp}
```
nick-thompson: I'm not sure this part needs to change?

examples/BlueprintPlugin/Source/PluginEditor.cpp
```diff
@@ -18,26 +18,34 @@ BlueprintPluginAudioProcessorEditor::BlueprintPluginAudioProcessorEditor (Bluepr
     File sourceDir = (File (__FILE__)).getParentDirectory();
 
     addAndMakeVisible(appRoot);
-    appRoot.evalScript(sourceDir.getChildFile("ui/build/js/main.js").loadFileAsString());
-    appRoot.enableHotkeyReloading();
-
-    appRoot.registerNativeMethod(
+    appRoot.evaluate(sourceDir.getChildFile("ui/build/js/main.js").loadFileAsString());
```
nick-thompson: :+1: perfect, thanks for this. You're right about the `enableHotKeyReloading` method, that was removed. Let's just delete that line here, and if you don't mind it looks like there's some extra whitespace on line 30

blueprint/core/blueprint_EcmascriptEngine.cpp
```diff
@@ -347,4 +347,9 @@ namespace blueprint
         return result;
     }
 
+    duk_context* EcmascriptEngine::getDuktapeContext()
```
nick-thompson: I'd like to keep this out of the `EcmascriptEngine` interface moving forward. My goal with that refactor was to create a generic interface for interacting with a js engine that could be implemented by any of various third party libraries. For example, I'm using Duktape here now, but I'd love to add another implementation via V8 that you could use with an optional switch.

So for that reason, if the details of the underlying impllementation leak into the interface, I think the interface should be reconsidered. For this case I don't think we need to do that, I'll comment below

examples/BlueprintPlugin/Source/PluginEditor.cpp
```diff
@@ -18,26 +18,34 @@ BlueprintPluginAudioProcessorEditor::BlueprintPluginAudioProcessorEditor (Bluepr
...
-        }
+            return juce::var::undefined();
+        },
+        (void *) this
     );
-
+    
+    
+    // Globals in the js env
+    auto* ctx = appRoot.engine.getDuktapeContext();
```
nick-thompson: So for this we can replace this chunk with just this:

```cpp
appRoot.engine.registerNativeProperty("__VERSION__", JucePlugin_VersionString);
```

examples/BlueprintPlugin/Source/PluginEditor.cpp
fefanto: ok should be done

fefanto: killed the duktape method as well

examples/BlueprintPlugin/Source/ui/src/App.js
fefanto: ok was replacing the Help tag to test the version global, would make total sense to host both the help and the version print in that same view ...but if I put  _<Help {...styles.contentContainer} />_ back I get this error (which I got also prior to my change)

```
**Blueprint Fatal Error:** uncaught: 'ScrollView must have only one child, and that child must be a ScrollView.ContentView.'
It is now unsafe to execute further code in the EcmascriptEngine.
libc++abi.dylib: terminating with uncaught exception of type std::runtime_error: uncaught: 'ScrollView must have only one child, and that child must be a ScrollView.ContentView.'


```

nick-thompson: Ah, yea, it depends where you're putting these things: `<Help>` is just a big `<View>` with a `<ScrollView>` inside of it. The `ScrollView` can only ever have one child, which is internally handled by setting the component of a `juce::Viewport`.

Maybe we just put the `<Text color={Colors.STROKE}>{"v" + global.__VERSION__}</Text>` inside `Help.js` as a `<SectionTitle>`?

fefanto: Anything goes for me... Just saying that that view was crashing in me also on master, does it work on yours. if it does work on your machine this means it's maybe coming from a difference in our Blueprint plugin specific package.json ,?

nick-thompson: Hm, that's odd, I mean the only `ScrollView` is inside `Help.js`, and that `Help` component doesn't care about any children coming in on props.

Seems a little bit like it's a problem unrelated to this PR, so I'll go ahead and merge the PR and then look into it separately. I'll comment back if I find it!

nick-thompson:
Ok figured out the Help screen issue– it was just a totally unrelated breakage from a change I made earlier 🙄 Sorry about that! I fixed it in b33615b2e1a836c7fa06ebe48c401a498e9a749d on master if you're curious.

Thanks for your work!

PR #35, opened 2019-12-28
Added a HotReloader for refreshing ReactApplicationRoot

dikadk:
It's pretty much same code as you have in GenericEditor. I wasn't sure what is the best way to refresh UI so decided to pass reference to juce::Component. Let me know if you had something else on your mind.

I think whole HotReloader will need to be disabled for release target. But this PR doesn't have code which checks for it.

### DIFF
--- a/blueprint/core/blueprint_HotReloader.cpp
+++ b/blueprint/core/blueprint_HotReloader.cpp
@@ -1,0 +1,69 @@
+/*
+  ==============================================================================
+
+    blueprint_HotRealoader.cpp
+    Created: 11 Dec 2019 2:20:29pm
+
+  ==============================================================================
+*/
+
+namespace blueprint
+{
+
+//==============================================================================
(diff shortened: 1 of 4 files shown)


blueprint/core/blueprint_HotReloader.cpp
```diff
@@ -0,0 +1,69 @@
...
+
+        lastModifiedTime = lmt;
+    }
+}
+
+void HotReloader::assignNewAppRoot(const juce::String &code)
+{
+
+    root = std::make_unique<blueprint::ReactApplicationRoot>();
+    root->engine.onUncaughtError = [this](const juce::String &msg, const juce::String &trace) {
```
nick-thompson: The only thing I see here is that it would be nice for the consumer of a `HotReloader` to provide this callback so that, for example, if my PluginEditor is using a HotReloader I can teardown and show a red screen with the error otuput in the PluginEditor

nick-thompson:
Thanks for your work @dikadk! I'm going to go ahead and close this as inactive and resolved by #54 in the meantime.

PR #46, opened 2020-05-21
View manager refactor

JoshMarler:
@nick-thompson this still needs a couple minor things tidying. Mainly documentation and I've added a few TODO comments as quick discussion points before this is merged. Will clean up and remove based on your feedback. Have tested this with hot reloading running and all appears to work. 

If you're OK with the basic shape I'll clean up all the documentation etc. The hot reloading as part of ReactApplicationRoot can follow this as next PR. Pretty much have that implemented. 

I can rebase and squash this all into a single commit to keep your history a little nicer once we're done if you like.

### DIFF
--- a/blueprint/core/blueprint_ReactApplicationRoot.h
+++ b/blueprint/core/blueprint_ReactApplicationRoot.h
@@ -21,7 +21,7 @@
 #include "blueprint_TextShadowView.h"
 #include "blueprint_TextView.h"
 #include "blueprint_View.h"
-
+#include "blueprint_ViewManager.h"
 
 namespace blueprint
 {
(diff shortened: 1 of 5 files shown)


blueprint/core/blueprint_ReactApplicationRoot.h
```diff
@@ -1,5 +1,5 @@
 /*
-  ==============================================================================
+  s=============================================================================
```
JoshMarler, line 2: Opps. Will clean that up in another commit

blueprint/core/blueprint_ReactApplicationRoot.h
```diff
@@ -34,238 +34,25 @@ namespace blueprint
...
-        typedef std::pair<std::unique_ptr<View>, std::unique_ptr<ShadowView>> ViewPair;
-        typedef std::function<ViewPair()> ViewFactory;
-
         //==============================================================================
         ReactApplicationRoot()
         {
             JUCE_ASSERT_MESSAGE_THREAD
 
+            //TODO: We used to construct the root ShadowView after calling registerNativeRenderingHooks.
+            //      Is there any issue moving the construction of the root ShadowView here?
```
JoshMarler, line 43: This appears to be all good from my testing but worth you sanity checking @nick-thompson

blueprint/core/blueprint_ViewManager.cpp
```diff
@@ -0,0 +1,239 @@
...
+
+    void ViewManager::registerViewType(const juce::String &typeId, ViewFactory f)
+    {
+        // If you hit this jassert, you're trying to register a type which
+        // has already been registered!
+        jassert (viewFactories.find(typeId) == viewFactories.end());
+        viewFactories[typeId] = f;
+    }
+
+    /** Creates a new view instance and registers it with the view table. */
```
JoshMarler, line 39: Oops. Will move this doc string to header in another commit

blueprint/core/blueprint_ViewManager.cpp
```diff
@@ -0,0 +1,239 @@
...
+    Created: 21 May 2020 15:32:00am
+
+  ==============================================================================
+*/
+
+#include "blueprint_ViewManager.h"
+
+namespace blueprint
+{
+    //TODO: Have the root ShadowView owned by ViewManager or have ReactApplicationRoot own this
```
JoshMarler, line 14: @nick-thompson  This is probably the key discussion point.

blueprint/core/blueprint_ViewManager.cpp
```diff
@@ -0,0 +1,239 @@
...
+    ViewManager::ViewManager(View* rootView)
+        : rootShadowView(std::make_unique<ShadowView>(rootView))
+    {
+        jassert(rootView);
+        rootView->addComponentListener(this);
+    }
+
+    ViewManager::~ViewManager()
+    {
+        View* rootView = rootShadowView->getAssociatedView();
```
JoshMarler, line 27: Needs an assertion.

blueprint/core/blueprint_ViewManager.cpp
```diff
@@ -0,0 +1,239 @@
...
+            if (auto* childView = dynamic_cast<View*>(child))
+            {
+                enumerateChildViewIds(ids, childView);
+            }
+        }
+
+        ids.push_back(v->getViewId());
+    }
+
+    void ViewManager::performRootShadowTreeLayout()
```
JoshMarler, line 193: Methods from this point are those with changes to use rootShadowView

blueprint/core/blueprint_ReactApplicationRoot.h
```diff
@@ -295,6 +82,19 @@ namespace blueprint
             engine.invoke("__BlueprintNative__.dispatchEvent", eventType, std::forward<T>(args)...);
         }
 
+        //==============================================================================
+        void registerViewType(const juce::String& typeId, ViewManager::ViewFactory f)
+        {
+            jassert(_viewManager);
```
JoshMarler, line 88: Add JUCE_ASSERT_MESSAGE_THREAD probably

blueprint/core/blueprint_ViewManager.h
```diff
@@ -0,0 +1,87 @@
...
+
+        /** Walks the view table, returning the first view with a `refId`
+         *  whose value equals the provided id.
+         */
+         //TODO: Currently unused. Keep?
+        View* getViewByRefId (const juce::Identifier& refId);
+
+        std::unique_ptr<ShadowView>  rootShadowView;
+
+        std::map<ViewId, std::unique_ptr<View>> viewTable;
```
JoshMarler, line 83: Not sure how you feel about style @nick-thompson but be nice to align all these member var names

blueprint/core/blueprint_ViewManager.cpp
nick-thompson: I like how you have it now– ViewManager seems like the right spot to own all shadow views

blueprint/core/blueprint_ViewManager.cpp
```diff
@@ -0,0 +1,239 @@
...
+{
+    //TODO: Have the root ShadowView owned by ViewManager or have ReactApplicationRoot own this
+    //      and pass a ViewPair like struct to ViewManager constructor?
+    //      Plausible passing a ViewPair fits better with the existing registerViewType
+    //      pattern?
+    ViewManager::ViewManager(View* rootView)
+        : rootShadowView(std::make_unique<ShadowView>(rootView))
+    {
+        jassert(rootView);
+        rootView->addComponentListener(this);
```
nick-thompson, line 22: I think I'd prefer to leave out the ComponentListener here; I think `ReactApplicationRoot` can maintain this line:

```
void resized() override {
  _viewManager->performShadowTreeLayout();
}
```

and then we don't have the need for the listener here. I lean that way because the component listener creates a bit of a circular dependency: the appRoot depends on the view manager, and the view manager depends backwards on the appRoot. (Of course in some sense that dep is always going to be there because the view manager needs to lay out the appRoot and everything else, but this feels like a way to minimize that circular dependency)

blueprint/core/blueprint_ViewManager.cpp
```diff
@@ -0,0 +1,239 @@
...
+#include "blueprint_ViewManager.h"
+
+namespace blueprint
+{
+    //TODO: Have the root ShadowView owned by ViewManager or have ReactApplicationRoot own this
+    //      and pass a ViewPair like struct to ViewManager constructor?
+    //      Plausible passing a ViewPair fits better with the existing registerViewType
+    //      pattern?
+    ViewManager::ViewManager(View* rootView)
+        : rootShadowView(std::make_unique<ShadowView>(rootView))
```
nick-thompson, line 19: What do you think about the `shadowViewTable` owning this unique_ptr? Maybe in the constructor here we make the root shadow view, put it in the map, and hang onto the `rootViewId` as a member instead of the pointer. That way the methods below that depend on lookups in the shadow view table can also find the root shadow view if they ever need to

blueprint/core/blueprint_ViewManager.h
```diff
@@ -0,0 +1,87 @@
...
+    private:
+        void enumerateChildViewIds (std::vector<ViewId>& ids, View* v);
+
+        /** Recursively computes the shadow tree layout on the root ShadowView, then traverses the tree
+            flushing new layout bounds to the associated view components.
+         */
+        void performRootShadowTreeLayout();
+
+        /** Returns a pointer pair to the view associated to the given id. */
+        //TODO: Private for now as only used internally but any desire for this to be made public?
```
nick-thompson, line 71: Good question. I think as long as it's used only internally we should keep it private. If and when we have a good case to make it public we can do it then

blueprint/core/blueprint_ViewManager.h
```diff
@@ -0,0 +1,87 @@
...
+        //      so classes like ReactApplicationRoot can fetch a View?
+        std::pair<View*, ShadowView*> getViewHandle (ViewId viewId);
+
+        /** Walks the view table, returning the first view with a `refId`
+         *  whose value equals the provided id.
+         */
+         //TODO: Currently unused. Keep?
+        View* getViewByRefId (const juce::Identifier& refId);
```
nick-thompson, line 79: Yea let's keep this one, it's useful for client code. For example, in this plugin WOV by K-Devices (https://k-devices.com/products/wov/), which is written in blueprint, those lines drawn between the knobs are done in the PluginEditor.h `paint()` method, and in that paint routine the editor asks the appRoot "hey do you have a view tagged by the refId `gainSlider`"? And with the `View*` the paint() routine is able to identify the exact location of that view and draw a line to it.

Then from the React code we can just label that view with the refId via props:

```
<Slider refId="gainSlider" />
```

It's neat when you want to kind of interoperate between react and native paint routines. Of course that was before we started talking about the Canvas and drawing stuff :) But I think this still makes sense here for now

blueprint/core/blueprint_ViewManager.h
nick-thompson: Sure! What do you suggest?

blueprint/core/blueprint_ViewManager.cpp
JoshMarler: Yeah I think that's nicer. Agreed. Fill fix up in my next set of changes

blueprint/core/blueprint_ViewManager.cpp
JoshMarler: Yeah cool. I wasn't too sure if only the ViewManager should have the responsibility of triggering performShadowTreeLayout. Will get this fixed up in next set of changes.

blueprint/core/blueprint_ViewManager.h
JoshMarler: Will kill the TODO in next set of changes.

blueprint/core/blueprint_ViewManager.h
JoshMarler: Ahh nice. I was actually wondering how obtaining views by the equivalent of a component I'd should work for a separate problem in my app. Cool, makes perfect sense. Will kill the TODO

blueprint/core/blueprint_ViewManager.h
JoshMarler: @nick-thompson cleaned up in https://github.com/nick-thompson/blueprint/pull/46/commits/9aba516339d06f6ff9a3d03a3bb5ce9bba592c7d

blueprint/core/blueprint_ViewManager.cpp
JoshMarler: @nick-thompson resolved in https://github.com/nick-thompson/blueprint/pull/46/commits/3500dd553b88731732f2d6d08f9563c38305aa5b

blueprint/core/blueprint_ViewManager.cpp
JoshMarler: @nick-thompson, changed made in https://github.com/nick-thompson/blueprint/pull/46/commits/71ed1a32f3c01f094cacf15b242800f33ffd2ce7

JoshMarler:
@nick-thompson hopefully this is ready for final review now. If you're happy with it I can rebase and squash everything down into a single commit?

blueprint/core/blueprint_ReactApplicationRoot.h
```diff
@@ -295,6 +84,21 @@ namespace blueprint
             engine.invoke("__BlueprintNative__.dispatchEvent", eventType, std::forward<T>(args)...);
         }
 
+        //==============================================================================
+        void registerViewType(const juce::String& typeId, ViewManager::ViewFactory f)
```
JoshMarler, line 88: @nick-thompson after this gets merged I think registerViewType could do with a doc comment that gives a clear example of usage.

nick-thompson: 👍 totally agree!

nick-thompson:
@JoshMarler this is excellent, thanks so much for the fixes. I'd love a rebase + squash, then happy to merge!

JoshMarler:
Awesome I'll get that sorted in a second.

JoshMarler:
@nick-thompson, All squashed up into single commit https://github.com/nick-thompson/blueprint/pull/46/commits/0e17dde01c444750ff1104ea2034e2619d5d7a8a

nick-thompson:
🙌 merged! Thanks again :) I'll check out the canvas PR now as well

PR #47, opened 2020-05-21
Fix gcc errors

JoshMarler:
@nick-thompson these changes fix some issues I had building on linux with GCC. It requires an update of yoga to fix an obscure enum-bitfield warning in GCC. We build with warnings as errors and this caused problems.

I've had these in my fork for a few weeks so have forgotten the exact errors but I can trigger said errors again if you'd like the compiler output.

### DIFF
--- a/blueprint/blueprint.cpp
+++ b/blueprint/blueprint.cpp
@@ -47,6 +47,7 @@
 #include "blueprint.h"
 
 #include "yoga/yoga/log.cpp"
+#include "yoga/yoga/event/event.cpp"
 #include "yoga/yoga/Utils.cpp"
 #include "yoga/yoga/YGConfig.cpp"
 #include "yoga/yoga/YGEnums.cpp"
--- a/blueprint/blueprint.h
+++ b/blueprint/blueprint.h
@@ -41,6 +41,7 @@
 #endif
 
 #include "yoga/yoga/log.h"
+#include "yoga/yoga/event/event.h"
 #include "yoga/yoga/Utils.h"
 #include "yoga/yoga/YGConfig.h"
 #include "yoga/yoga/YGEnums.h"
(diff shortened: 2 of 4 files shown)


nick-thompson:
Awesome, thanks @JoshMarler . I'm reviewing now, but when I build this branch I get an issue

```
yoga/YGEnums.h not found
```
at
```
#include <yoga/YGEnums.h>
```
in yoga/event/event.h

Did you see that? I'm able to resolve it just adding yoga to the search paths in the module definition:

```
diff --git a/blueprint/blueprint.h b/blueprint/blueprint.h
index 3ddfc42..59fd1e8 100644
--- a/blueprint/blueprint.h
+++ b/blueprint/blueprint.h
@@ -12,7 +12,7 @@
     description:        A React.js render backend targeting JUCE Components.
     minimumCppStandard: 17
     dependencies:       juce_core, juce_graphics, juce_gui_basics
-    searchpaths:        ./duktape/src-noline/
+    searchpaths:        ./duktape/src-noline/ ./yoga
    END_JUCE_MODULE_DECLARATION
 *******************************************************************************/
```

That work for you? If so, I'm ready to merge with that little fix (I'm happy to make the fix as I merge in as well if that's helpful)

JoshMarler:
Ah I have to confess I haven't been building from the Projucer as I'm working in CMake. That is a very good point. I'll sanity check everything builds with the Projucer shortly. Apologies I have to shoot out for a while now but I'll take a look at this all again tomorrow I expect. I should probably try to get some Vitamin D!

If it builds for you though @nick-thompson I think it's probably safe as that means good in both systems.

nick-thompson:
No rush :D

I think it's safe too– I can confirm this works with that change in Projucer/MacoS builds, and surely that little change won't break Cmake builds. I'll just go ahead and merge and make that little fix on the way in!

PR #49, opened 2020-05-22
Move Canvas API funtion bindings into a CanvasContext object

JoshMarler:
@nick-thompson, I've started messing with adding more of the Canvas API.  Thought I'd drop this PR over even if just as a discussion starter. Tried to encapsulate all the registering of native functions and properties into a little CanvasContext class so it's easily used from CavasView::paint().

It would be nice to add a few little drawing examples to the README maybe. The README could list each of the CanvasRenderingContext2D API functions currently supported and we could grow the list incrementally. 

In this PR the following CanvasRenderingContext2D props and functions have been added:

. strokeStyle 
. fillStyle
. lineWidth 

. rect()
. fillRect()
. strokeRect()

. beginPath()
. lineTo()
. moveTo()
. arc()
. quadraticCurveTo()
. closePath()

. stroke()
. fill()

. rotate()
. translate()
. setTransform()

If you're happy with the shape of this I thought I could drop another PR over next which has the following:

. fillText()
. strokeText()
. font

. drawImage() - drawImage will probably just support an svg string initially. 

I think with this set of API calls there is a nice starter set of functions for drawing. Perhaps nice to add gradients and stroke styles after that.

### DIFF
--- a/blueprint/core/blueprint_CanvasView.h
+++ b/blueprint/core/blueprint_CanvasView.h
@@ -11,19 +11,411 @@
 
 #include "blueprint_View.h"
 
-
 namespace blueprint
 {
+    /** The CanvasContext class is a C++ implementation/binding for the
+     *  HTML Canvas API's CanvasRenderingContext2D object.
+     *
+     *  https://developer.mozilla.org/en-US/docs/Web/API/CanvasRenderingContext2D
+     *
+     *  CanvasContext is used by the CanvasView class to support HTML <canvas> like rendering in React/Blueprint
+     *  components. CanvasContext could also be used outside of the CanvasView element in things like juce::LookAndFeel
+     *  overrides.
+     *
+     *  CanvasContext draws to a juce::Image internally.
+     *
+     *  To use CanvasContext you simply need to instantiate a CanvasContext instance, call init() and then pass the
+     *  CanvasContext instance to your JS drawing routine. Your JS drawing routine will then render to the
+     *  CanvasContext's juce::Image when invoked, this juce::Image instance can then be rendered anywhere using
+     *  CanvasContext::getImage() and juce::Graphics::drawImageAt(). See the CanvasView::paint() implementation
+     *  for an example of this.
+     *
+     *  Note that is is important to call CanvasContext::setSize() before invoking your JS drawing routine.
+     *
+     *  A typical JS drawing routine that uses this implementation of the CanvasRenderingContext2D may look like the
+     *  below:
+     *
+     * @code
+     *      function drawCircle(ctx, x, y, radius, filled) {
+     *          ctx.beginPath();
+     *          ctx.moveTo(x + radius, y);
+     *          ctx.arc(x, y, radius, 0, 2 * Math.PI);
+     *
+     *          if (filled)
+     *              ctx.fill();
+     *          else
+     *              ctx.stroke();
+     *      }
+     *
+     *      function doDraw(ctx) {
+     *          ctx.strokeColour = `ff884848`;
+     *          ctx.fillColour   = `ffffffaa`;
+     *
+     *          drawCircle(ctx, x, y, radius, true);
+     *      }
+     * @endcode
+     *
+     */
+    class CanvasContext : public juce::DynamicObject
+    {
+    public:
+        using FillStyle   = juce::FillType;
+        using StrokeStyle = juce::FillType;
+
+        struct Properties
+        {
+            FillStyle      fillStyle{};
+            StrokeStyle    strokeStyle{};
+            int            lineWidth = 1;
+        };
+
+        explicit CanvasContext()
+            : image(juce::Image::ARGB, 1, 1, true)
+            , graphics(std::make_unique<juce::Graphics>(image))
+        {
+            registerNativeProperties();
+            registerNativeFunctions();
+        }
+
+        ~CanvasContext() = default;
+
+        /** Set the size of the CanvasContext. This is akin to setting the width and height attributes
+         *  on a HTML <canvas> element. If you have a blueprint::View or a juce::Component that wished
+         *  to render a CanvasContext you should ensure CanvasContext::setSize() is called from the
+         *  View/Component's resized function.
+         */
+        void setSize(int width, int height)
+        {
+            image = juce::Image(juce::Image::ARGB, width, height, true);
+        }
+
+        /** Get the juce::Image instance that has been rendered to. This can be drawn to the screen
+         *  using juce::Graphics::drawImageAt().
+         */
+        const juce::Image& getImage() const
+        {
+            return image;
+        }
+
+        /**  init should be called before passing the CanvasContext instance to a JS drawing function. */
+        void init()
+        {
+            graphics = std::make_unique<juce::Graphics>(image);
+        }
+
+    private:
+        void registerNativeProperties()
+        {
+            setProperty("__setFillStyle", juce::var::NativeFunction {
+                    [=](const juce::var::NativeFunctionArgs& args) -> juce::var {
+                        jassert(graphics);
+                        jassert(args.numArguments == 1);
+
+                        //TODO: Implement fillStyle pattern
+                        //TODO: Implement fillStyle gradient
+
+                        const auto colourHex = args.arguments[0].toString();
+                        properties.fillStyle.setColour(juce::Colour::fromString(colourHex));
+
+                        graphics->setFillType(properties.fillStyle);
+
+                        return juce::var();
+                    }
+            });
+
+            setProperty("__setStrokeStyle", juce::var::NativeFunction {
+                    [=](const juce::var::NativeFunctionArgs& args) -> juce::var {
+                        jassert(graphics);
+                        jassert(args.numArguments == 1);
+
+                        //TODO: Implement strokeStyle pattern
+                        //TODO: Implement stokeStyle gradient
+
+                        const auto colourHex = args.arguments[0].toString();
+                        properties.strokeStyle.setColour(juce::Colour::fromString(colourHex));
+
+                        return juce::var();
+                    }
+            });
+
+            setProperty("__setLineWidth", juce::var::NativeFunction {
+                    [=](const juce::var::NativeFunctionArgs& args) -> juce::var {
+                        jassert(graphics);
+                        jassert(args.numArguments == 1);
+
+                        const int lineWidth = args.arguments[0];
+                        properties.lineWidth = lineWidth;
+
+                        return juce::var();
+                    }
+            });
+        }
+
+        void registerNativeFunctions()
+        {
+//            setProperty("rect", juce::var::NativeFunction {
+//                    [=](const juce::var::NativeFunctionArgs& args) -> juce::var {
+//                        jassert(args.numArguments == 4);
+//
+//                        const float x = args.arguments[0];
+//                        const float y = args.arguments[1];
+//                        const float width = args.arguments[2];
+//                        const float height = args.arguments[3];
+//
+//                        // TODO: For some reason this operation closes the current path.
+//                        //       The API docs suggest addRectangle should be added as a new sub-path
+//                        //       leaving the previous path open.
+//                        //       Fix.
+//                        path.addRectangle(x, y, width, height);
+//
+//                        return juce::var();
+//                    }
+//            });
+
+            setProperty("fillRect", juce::var::NativeFunction {
+                    [=](const juce::var::NativeFunctionArgs& args) -> juce::var {
+                        jassert(graphics);
+                        jassert(args.numArguments == 4);
+
+                        const int x = args.arguments[0];
+                        const int y = args.arguments[1];
+                        const int width = args.arguments[2];
+                        const int height = args.arguments[3];
+
+                        graphics->fillRect(x, y, width, height);
+
+                        return juce::var();
+                    }
+            });
+
+            setProperty("strokeRect", juce::var::NativeFunction {
+                    [=](const juce::var::NativeFunctionArgs& args) -> juce::var {
+                        jassert(graphics);
+                        jassert(args.numArguments == 4);
+
+                        const int x = args.arguments[0];
+                        const int y = args.arguments[1];
+                        const int width = args.arguments[2];
+                        const int height = args.arguments[3];
+
+                        graphics->drawRect(x, y, width, height, properties.lineWidth);
+
+                        return juce::var();
+                    }
+            });
+
+            //TODO: Add clearRect. Use View background color and g.fillRect?
+
+            // Path functions
+            setProperty("beginPath", juce::var::NativeFunction {
+                    [=](const juce::var::NativeFunctionArgs& args) -> juce::var {
+                        jassert(graphics);
+                        jassert(args.numArguments == 0);
+
+                        // Reset the current path on a call to beginPath on the context.
+                        path = juce::Path();
+
+                        return juce::var();
+                    }
+            });
+
+            setProperty("lineTo", juce::var::NativeFunction {
+                    [=](const juce::var::NativeFunctionArgs& args) -> juce::var {
+                        jassert(graphics);
+                        jassert(args.numArguments == 2);
+
+                        const float x = args.arguments[0];
+                        const float y = args.arguments[1];
+
+                        path.lineTo(x, y);
+
+                        return juce::var();
+                    }
+            });
+
+            setProperty("moveTo", juce::var::NativeFunction {
+                    [=](const juce::var::NativeFunctionArgs& args) -> juce::var {
+                        jassert(graphics);
+                        jassert(args.numArguments == 2);
+
+                        const float x = args.arguments[0];
+                        const float y = args.arguments[1];
+
+                        path.startNewSubPath(x, y);
+
+                        return juce::var();
+                    }
+            });
+
+            setProperty("arc", juce::var::NativeFunction {
+                    [=](const juce::var::NativeFunctionArgs& args) -> juce::var {
+                        jassert(graphics);
+
+                        // CanvasRenderingContext2D.arc() uses default antiClockWise false arg
+                        jassert(args.numArguments >= 5 && args.numArguments <= 6);
+
+                        const float x             = args.arguments[0];
+                        const float y             = args.arguments[1];
+                        const float radius        = args.arguments[2];
+                        const float startAngle    = args.arguments[3];
+                        const float endAngle      = args.arguments[4];
+                        //TODO; Handle antiClockWise
+                        //bool        antiClockWise = args.arguments[5];
+
+                        const float width  = radius * 2;
+                        const float height = radius * 2;
+
+                        path.addArc(x, y, width, height, startAngle, endAngle);
+
+                        return juce::var();
+                    }
+            });
+
+            setProperty("quadraticCurveTo", juce::var::NativeFunction {
+                    [=](const juce::var::NativeFunctionArgs& args) -> juce::var {
+                        jassert(graphics);
+                        jassert(args.numArguments == 4);
+
+                        const float controlPointX = args.arguments[0];
+                        const float controlPointY = args.arguments[1];
+                        const float endPointX     = args.arguments[2];
+                        const float endPointY     = args.arguments[3];
+
+                        path.quadraticTo(controlPointX, controlPointY, endPointX, endPointY);
+
+                        return juce::var();
+                    }
+            });
+
+            //TODO: Implement CanvasRenderingContext2D.bezierCurveTo(). Will use juce::Path::cubicTo
+
+            setProperty("closePath", juce::var::NativeFunction {
+                    [=](const juce::var::NativeFunctionArgs& args) -> juce::var {
+                        jassert(graphics);
+                        jassert(args.numArguments == 0);
+
+                        path.closeSubPath();
+
+                        return juce::var();
+                    }
+            });
+
+            setProperty("stroke", juce::var::NativeFunction {
+                    [=](const juce::var::NativeFunctionArgs& args) -> juce::var {
+                        jassert(graphics);
+                        jassert(args.numArguments == 0);
+
+                        graphics->setColour(properties.strokeStyle.colour);
+                        graphics->strokePath(path, juce::PathStrokeType((float)properties.lineWidth));
+
+                        return juce::var();
+                    }
+            });
+
+            setProperty("fill", juce::var::NativeFunction {
+                    [=](const juce::var::NativeFunctionArgs& args) -> juce::var {
+                        jassert(graphics);
+                        jassert(args.numArguments == 0);
+
+                        graphics->setColour(properties.strokeStyle.colour);
+                        graphics->strokePath(path, juce::PathStrokeType((float)properties.lineWidth));
+                        graphics->setColour(properties.fillStyle.colour);
+                        graphics->fillPath(path);
+
+                        return juce::var();
+                    }
+            });
+
+            // Transforms
+            setProperty("rotate", juce::var::NativeFunction {
+                    [=](const juce::var::NativeFunctionArgs& args) -> juce::var {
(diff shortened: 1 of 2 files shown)


JoshMarler:
One interesting idea around Canvas API support is augmenting the standard CanvasRenderingContext2D API with some of the additional functions in juce such as drawRoundedRect (not part of the html Canvas API). 

It would be nice to offer some additional drawing functions familiar to juce users without breaking the CanvasRenderingContext2D API and potential web based UIs.

blueprint/core/blueprint_CanvasView.h
```diff
@@ -21,6 +21,190 @@ namespace blueprint
...
+                            jassert(args.numArguments == 0);
+
+                            // Reset the current path on a call to beginPath on the context.
+                            path = juce::Path();
+
+                            return juce::var();
+                        }
+                });
+
+                setProperty("lineTo", juce::var::NativeFunction {
```
JoshMarler, line 154: I need to test this further before anything gets merged. Just pushed to illustrate how some initial Path support might look.

blueprint/core/blueprint_CanvasView.h
```diff
@@ -21,6 +21,180 @@ namespace blueprint
...
+                            g.drawRect(x, y, width, height, properties.lineWidth);
+
+                            return juce::var();
+                        }
+                });
+
+                //TODO: Add clearRect. Use View background color and g.fillRect?
+
+                // Path functions
+                setProperty("beginPath", juce::var::NativeFunction {
```
JoshMarler, line 133: The following example works as you would expect with these changes
```
function doCanvasDraw(ctx) {
  ctx.strokeStyle = `ff884848`;
  ctx.lineWidth = 5;

  ctx.beginPath();
  ctx.lineTo(150, 0);
  ctx.lineTo(150, 50);
  ctx.lineTo(0, 50);
  ctx.closePath();
  ctx.stroke();
}
```

blueprint/core/blueprint_CanvasView.h
```diff
@@ -21,6 +21,201 @@ namespace blueprint
...
+                            const float x = args.arguments[0];
+                            const float y = args.arguments[1];
+
+                            path.lineTo(x, y);
+
+                            return juce::var();
+                        }
+                });
+
+                setProperty("arc", juce::var::NativeFunction {
```
JoshMarler, line 157: Added simple arc() support which works with the following example:

```
function drawCircle(ctx, x, y, radius) {
  ctx.beginPath();
  ctx.moveTo(x + radius, y);
  ctx.arc(x, y, radius, 0, 2 * Math.PI);
  ctx.stroke();
}
```

blueprint/core/blueprint_CanvasView.h
```diff
@@ -21,6 +21,216 @@ namespace blueprint
...
+                            const float width  = radius * 2;
+                            const float height = radius * 2;
+
+                            path.addArc(x, y, width, height, startAngle, endAngle);
+
+                            return juce::var();
+                        }
+                });
+
+                setProperty("quadraticCurveTo", juce::var::NativeFunction {
```
JoshMarler, line 178: quadraticCurveTo tested and working with below example:
```
function drawRoundedRect(ctx, x, y, width, height, radius) {
  ctx.beginPath();

  ctx.moveTo(x + radius, y);
  ctx.lineTo(x + width - radius, y);
  ctx.quadraticCurveTo(x + width, y, x + width, y + radius);
  ctx.lineTo(x + width, y + height - radius);
  ctx.quadraticCurveTo(x + width, y + height, x + width - radius, y + height);
  ctx.lineTo(x + radius, y + height);
  ctx.quadraticCurveTo(x, y + height, x, y + height - radius);
  ctx.lineTo(x, y + radius);
  ctx.quadraticCurveTo(x, y, x + radius, y);

  ctx.closePath();
  ctx.stroke();
}

function doCanvasDraw(ctx) {
  ctx.strokeStyle = `ff884848`;
  ctx.lineWidth = 5;

  drawRoundedRect(ctx, 100, 10, 50, 130, 20);
}
```

JoshMarler:
Dropping comment here whilst I think about this.

Its actually probably not necessary to register methods and properties in every paint call. The CanvasContext class could simply have a pointer to a juce::Graphics instance as a data member.

Then in CanvasView::paint we just call CanvasContext::draw(&g) which reassigns the pointer uses in each of the native methods and callbacks and invokes onDraw etc. That could get rid of the overhead of registering methods in every single call to paint(). I'll experiment with it later when I return to my desk.

blueprint/core/blueprint_CanvasView.h
```diff
@@ -11,9 +11,276 @@
...
+
+                        graphics->setColour(properties.strokeStyle.colour);
+                        //TODO: Consider how to map HTML Canvas API properties to juce::PathStrokeType
+                        graphics->strokePath(path, juce::PathStrokeType((float)properties.lineWidth));
+
+                        return juce::var();
+                    }
+            });
+
+            setProperty("fill", juce::var::NativeFunction {
```
JoshMarler, line 264: Separate stroke and fill colors appear to be working based on below example:

```
function drawRoundedRect(ctx, x, y, width, height, radius, filled) {
  ctx.beginPath();

  ctx.moveTo(x + radius, y);
  ctx.lineTo(x + width - radius, y);
  ctx.quadraticCurveTo(x + width, y, x + width, y + radius);
  ctx.lineTo(x + width, y + height - radius);
  ctx.quadraticCurveTo(x + width, y + height, x + width - radius, y + height);
  ctx.lineTo(x + radius, y + height);
  ctx.quadraticCurveTo(x, y + height, x, y + height - radius);
  ctx.lineTo(x, y + radius);
  ctx.quadraticCurveTo(x, y, x + radius, y);

  ctx.closePath();

  if (filled)
    ctx.fill();
  else
    ctx.stroke();
}

function drawFilledAndStrokedRect(ctx, x, y, width, height, radius, strokeColour, fillColour) {
  ctx.strokeStyle = strokeColour;
  ctx.fillStyle = fillColour;

  drawRoundedRect(ctx, x, y, width, height, radius, true);
}
```

blueprint/core/blueprint_CanvasView.h
```diff
@@ -11,9 +11,321 @@
...
+                        graphics->strokePath(path, juce::PathStrokeType((float)properties.lineWidth));
+                        graphics->setColour(properties.fillStyle.colour);
+                        graphics->fillPath(path);
+
+                        return juce::var();
+                    }
+            });
+
+            // Transforms
+            setProperty("rotate", juce::var::NativeFunction {
```
JoshMarler, line 278: Simple transform and rotation support allowed drawing of below slider like thing.

image

```
function drawCircle(ctx, x, y, radius, filled) {
  ctx.beginPath();
  ctx.moveTo(x + radius, y);
  ctx.arc(x, y, radius, 0, 2 * Math.PI);

  if (filled)
    ctx.fill();
  else
    ctx.stroke();
}

function drawRotaryKnob(ctx, x, y, radius) {

  // Draw slider thumb
  drawCircle(ctx, x, y, radius, true);

  ctx.moveTo(x + radius, y);
  ctx.lineTo(x + radius, y + (radius / 1.5));
  ctx.stroke();

  // Draw slider track
  const pad     = 20;
  const arcSize = 0.7;

  ctx.beginPath();
  ctx.moveTo((x - pad) + (radius + pad), y - pad);
  ctx.arc(x - pad, y - pad, (radius + pad), 0,  2 * Math.PI * arcSize);

  const rotateAngle = (180 + (180 * (1 - arcSize)))  * Math.PI / 180;

  const deltaX = x + radius;
  const deltaY = y + radius;

  ctx.translate(deltaX, deltaY);
  ctx.rotate(rotateAngle);
  ctx.translate(-deltaX, -deltaY);

  ctx.stroke();
}

function doCanvasDraw(ctx) {
  ctx.strokeColour = `ff884848`;
  ctx.fillColour   = `ffffffaa`;
  drawRotaryKnob(ctx, 400, 30, 40);
}
```

nick-thompson:
Woah, @JoshMarler you're amazing! Just seeing all these notifications now, I'll try to get through as much as I can this morning

JoshMarler:
@nick-thompson, No problem! there's no rush. Apologies if I've spammed you a little bit! I blame the lock-down and having little else to do!

JoshMarler:
@nick-thompson just sitting down to fixup the transform stuff and give this all a more thorough testing. I think having a transform member in CanvasContext will make the most sense based on the expected behaviour of the CanvasRenderingContext2D API. Be cool to hear your thoughts on the overall shape of this in the meantime. Still a slight work in progress. I'll aim to get this existing API calls all working flawlessly and won't add any further elements of the CanvasRenderingContext2D API until this PR goes in.

blueprint/core/blueprint_CanvasView.h
JoshMarler: Got a resetTransform() function working with the above approach. Still a couple kinks to workout.

JoshMarler:
> @JoshMarler overall, _love_ this! Inheriting the `DynamicObject` is such a good idea, I never thought of it. Also you covered so much of the API already 
> 
> I have two kind of high level thoughts– one of them is inline about the transform stuff, the other is: instead of receiving a pointer to a graphics context during paint, and a `View` ref at construction, what if the `CanvasContext` owned a `juce::Image` and all of the drawing routines happened on a `juce::Graphics g(memberImage)` pointing at that image? One on hand, that makes me feel a little better about lifetime stuff (referencing a raw pointer always makes me want to `jassert` everywhere and stuff), and on another I think that might leave room for some like easy draw routine caching. Then the `CanvasView`'s `paint()` routine could basically be:
> 
> ```
> if (props["onDraw"] && props["onDraw"].isMethod()) {
>     std::invoke(props["onDraw"].getNativeFunction(), args);
> }
> g.drawImageAt(canvasCtx->getImage(), 0, 0);
> ```
> 
> I think we would want to resize that image during `CanvasView`'s `resize()` so that it always matches perfectly.
> 
> Besides that, I think instead of taking a `View&` at construction I'd rather implement a method on the CanvasContext that creates the js args for you. This way we avoid a circular dependency and the CanvasContext isn't necessarily coupled to a particular `View` as well. I could imagine somethign like this:
> 
> ```
> // This is CanvasView's `paint()` method
> if (props["onDraw"] && props["onDraw"].isMethod()) {
>     std::invoke(props["onDraw"].getNativeFunction(), canvasCtx->getNativeFunctionArgs());
> }
> ```
> 
> What do you think? Besides those two ideas, I think this is basically ready to merge, and then we can tackle other paint methods in another PR as you suggested

@nick-thompson Ohh man yeah. I hadn't even thought about using an image. That could open up a really nice API for the look and feel override thing. Being able image states/UI snapshots etc. could open up all kinds of stuff. You clearly know the juce Graphics API better than me!

I'll have a go at this now.

blueprint/core/blueprint_CanvasView.h
JoshMarler: @nick-thompson, yeah a stack is nice idea. I effectively have this in a less elegant shape at the moment. Hitting a couple of issues with it around invert() not applying to some transforms etc. This sounds like the right track to head down though. I'll take a stab at this now and let you know how it goes.

JoshMarler:
@nick-thompson. Yeah nice the transform stack does the trick.

blueprint/core/blueprint_CanvasView.h
JoshMarler: @nick-thompson, resolved in https://github.com/nick-thompson/blueprint/pull/49/commits/0dfb056c504738c28a7b553ae979174e47ee52f8. Tested and works great! Thanks for the tip.

JoshMarler:
@nick-thompson, woop woop. Drawing to an image all working nicely. Will push up shortly. If it's cool with you I'd like to just give this whole PR a bit of a battle test. I'll try drawing various things just to make sure I haven't introduced any odd behaviour.

JoshMarler:
@nick-thompson, OK. So I have this all working with the transform stack and drawing to an Image in CanvasContext. However, resizing the image held by the CanvasContext isn't 100% straight forward.

We need to store a juce::Graphics instance in CanvasContext to preserve progressive transform state changes etc. (As per the current implementation). However, we cannot directly resize the image held by CanvasContext's Graphics instance so instead we need to replace/reconstruct the Graphics instance when we resize. 

I can push the current working state up but basically in CanvasView::resize() we call the following CanvasContext::setSize() function:

```
        void CanvasContext::setSize(int width, int height)
        {
            image = juce::Image(juce::Image::ARGB, width, height, true);
            graphics = std::make_unique<juce::Graphics>(image);
        }

        void CanvasView::resized() override
        {
            ctx->setSize(getWidth(), getHeight());
        }

        void CanvasView::paint (juce::Graphics& g) override
        {
            View::paint(g);

            jassert(ctx);

            if (props.contains("onDraw") && props["onDraw"].isMethod())
            {
                std::vector<juce::var> jsArgs {{ctx.get()}};
                juce::var::NativeFunctionArgs nfargs (juce::var(), jsArgs.data(), jsArgs.size());

                std::invoke(props["onDraw"].getNativeFunction(), nfargs);
            }

            g.drawImageAt(ctx->getImage(), 0, 0);
        }       
```

This all seems to be working fine and I thiiink this is OK but do you feel there is any more elegant way to do this? Drawing to an Image is really powerful and perfectly fits in with my look and feel requirement so I'd like to take that approach as you've suggested.

blueprint/core/blueprint_CanvasView.h
JoshMarler: @nick-thompson, my only slight reservation with this approach is it might mean a bit of acrobatics to support getTransform(): https://developer.mozilla.org/en-US/docs/Web/API/CanvasRenderingContext2D/getTransform

But maybe that isn't something we need to worry about right at this second and we can deal with in a future change.

nick-thompson: Ahh interesting. Yea, I don't think we need to support `getTransform` right now, that can be a "deal with it later" kind of thing. If our implementation supports _most_ of the Canvas Web API then I think we're in a reallly good spot.

That said, thinking about it, `getTransform`, as far as I understand it, returns a single transform that reflects the current state of the canvas context. So if you apply 5 rotations and 3 translations, they're all represented in a single affine transform matrix that is returned here. To do that we just need matrix mults of the underlying matrix coefficients. So, instead of storing the matrix inverses in our `transformStack`, we could store them as is so that we can do this:

```cpp
setProperty("rotate", juce::var::NativeFunction {
  /// etc etc
  transformStack.push_back(transform);
});

setProperty("getTransform", juce::var::NativeFunction {
    // etc etc
    Matrix<3, 3> a = Matrix<3, 3>::Identity();
    for (auto& t : transformStack) {
        a = a * makeMatrixFromAffineTransform(t);
    }
    return makeVarFromMatrix(a);
}

setProperty("resetTransform", juce::var::NativeFunction {
  /// etc etc
  while (transformStack.size() > 0) {
     g.addTransform(transformStack.pop_back().inverted();
  }
  return juce::var();
});
```

I got a little hand-wavy around the matrix multiplication, but hopefully you see what I mean. Let me know what you think! And yea I'm totally happy to defer this to a later date. If you think this sounds like a good approach we can write a new issue to track implementing the `getTransform` stuff

JoshMarler: Ha yeah that was the exact "acrobatics" I had in mind. It's easily sorted when we come to it.

nick-thompson:
> @nick-thompson, OK. So I have this all working with the transform stack and drawing to an Image in CanvasContext. However, resizing the image held by the CanvasContext isn't 100% straight forward.
> 
> We need to store a juce::Graphics instance in CanvasContext to preserve progressive transform state changes etc. (As per the current implementation). However, we cannot directly resize the image held by CanvasContext's Graphics instance, so instead we need to replace/reconstruct the Graphics instance when we resize.
> 
> I can push the current working state up but basically in CanvasView::resize() we call the following CanvasContext::setSize() function:

Ahhh right right, of course. Yea, I think your idea is great. The only thing I can think of is that if we're hanging onto the `Graphics` instance as a member variable, we'll probably want to make sure to reset the graphics context (https://docs.juce.com/master/classGraphics.html#ab8b7cd49bf1ff738c5ff848727e3bc75) at the beginning of `CanvasView::paint()`, otherwise context state changes (such as transformations) will accumulate with each successive call to `onDraw`

Besides that, yea, happy to go ahead with this plan, then squash and merge! And I'd love your help with the README afterwards. Thanks @JoshMarler

blueprint/core/blueprint_CanvasView.h
nick-thompson: Yea, acrobatics for a later time 😆 

Ok great, I'll document this with a new issue

JoshMarler:
Ahh yeah, good shout on the reset. Not clear if resetToDefaultState clears transforms but failing that it's easy enough to just add a reset function to CanvasContext which resets the Graphics unique_ptr. Awesome. Will finish this all off first thing tomorrow

JoshMarler:
@nick-thompson. One other question I've ignored until now before we merge this. 

Juce has slightly different path behaviour to the Canvas if we take a circle drawing example:

https://www.w3schools.com/html/tryit.asp?filename=tryhtml5_canvas_tut_path2

In the above example there is no need for a call to moveTo() before the call to arc(). Juce on the other hand will stroke the path from 0,0 to the start of the arc position.

I.e the following code doesn't product the same thing as the linked browser example:

```
  const radius = 40;
  ctx.beginPath();
  ctx.arc(100, 50, radius, 0, 2 * Math.PI);
  ctx.stroke();
```

The above produces this:

image

To render the desired shape we need to change the drawing code to use a moveTo()

```
  const radius = 40;
  ctx.beginPath();
  ctx.moveTo(100 + radius, 50);
  ctx.arc(100, 50, radius, 0, 2 * Math.PI);
  ctx.stroke();
```

Should this be considered an issue? We could add a startNewSubPath() call to the C++ implementation of things like arc() but part of me thinks this is undesirable and users will just have to be aware of the need to call moveTo()

```
            setProperty("arc", juce::var::NativeFunction {
                    [=](const juce::var::NativeFunctionArgs& args) -> juce::var {
                        jassert(args.numArguments >= 5 && args.numArguments <= 6);

                        const float x             = args.arguments[0];
                        const float y             = args.arguments[1];
                        const float radius        = args.arguments[2];
                        const float startAngle    = args.arguments[3];
                        const float endAngle      = args.arguments[4];
                        //TODO; Handle antiClockWise
                        //bool        antiClockWise = args.arguments[5];

                        const float width  = radius * 2;
                        const float height = radius * 2;

                        path.startNewSubPath(x, y);
                        path.addArc(x, y, width, height, startAngle, endAngle);

                        return juce::var();
                    }
            });

```

blueprint/core/blueprint_CanvasView.h
```diff
@@ -59,44 +451,34 @@ namespace blueprint
...
-                        const int y = args.arguments[1];
-                        const int width = args.arguments[2];
-                        const int height = args.arguments[3];
-
-                        g.fillRect(x, y, width, height);
+                std::invoke(props["onDraw"].getNativeFunction(), nfArgs);
+            }
+            else
+            {
+                jassertfalse;
```
JoshMarler, line 466: Having TypeScript interfaces for component properties would really help in avoiding these kinds of errors. Make it a strongly types constrains that the <Canvas> component requires an "onDraw" prop.

blueprint/core/blueprint_CanvasView.h
```diff
@@ -11,19 +11,411 @@
...
+            , graphics(std::make_unique<juce::Graphics>(image))
+        {
+            registerNativeProperties();
+            registerNativeFunctions();
+        }
+
+        ~CanvasContext() = default;
+
+        /** Set the size of the CanvasContext. This is akin to setting the width and height attributes
+         *  on a HTML <canvas> attribute. If you have a blueprint::View or a juce::Component that wished
```
JoshMarler, line 83: Will change to "on a HTML <canvas> element"

JoshMarler:
@nick-thompson, I've squashed this all down to a single commit and added some initial comments/documentation. If you're happy with the shape of everything then I think this should be good to merge now.

blueprint/core/blueprint_CanvasView.h
nick-thompson: Totally agree. I think this particular case makes more sense as a warning, not an assert. Technically your app can still run and function correctly without this prop, you just won't get your draw callbacks

blueprint/core/blueprint_CanvasView.h
```diff
@@ -11,19 +11,411 @@
...
+//                        const float x = args.arguments[0];
+//                        const float y = args.arguments[1];
+//                        const float width = args.arguments[2];
+//                        const float height = args.arguments[3];
+//
+//                        // TODO: For some reason this operation closes the current path.
+//                        //       The API docs suggest addRectangle should be added as a new sub-path
+//                        //       leaving the previous path open.
+//                        //       Fix.
+//                        path.addRectangle(x, y, width, height);
```
nick-thompson, line 169: :+1: we can manually do some `lineTo`'s here I think, right?

blueprint/core/blueprint_View.cpp
```diff
@@ -53,6 +53,11 @@ namespace blueprint
             _refId = juce::Identifier(value.toString());
     }
 
+    const juce::var& View::getProperty(const juce::Identifier& name) const
```
nick-thompson, line 56: I think we can omit this method now, right?

blueprint/core/blueprint_CanvasView.h
JoshMarler: I tried a move to before the addRectangle call bit it appeared to close the previous path. If you'd prefer I can just remove the commented out block of code and fix this in a commit after this. I don't think rect() is essential at this point.

blueprint/core/blueprint_View.cpp
JoshMarler: Yeah fair point. Will remove.

nick-thompson:
> @nick-thompson, I've squashed this all down to a single commit and added some initial comments/documentation. If you're happy with the shape of everything then I think this should be good to merge now.

Awesome @JoshMarler this looks great! Two more small asks from me in there– the `getProperty` method and the `jassertfalse` becoming a warning, otherwise I'm ready to merge :)

To your point above about the `moveTo`, yea, that's tricky. I think you're right that it makes the most sense for users to just be aware of the need to call this `moveTo` here. Especially as you've put it here:

> Calling moveTo() does not break the desired behaviour/output in a browser so the code is cross-compatible

I think that definitely answers the question. Maybe the takeaway here is that as we build out the documentation for CanvasView we should have a "things you should know" section or something where we can highlight little details like this

blueprint/core/blueprint_CanvasView.h
nick-thompson: Hmm. No worries, let's leave the commented block in. I think having the bones in place there will be helpful for our own remembering, and helpful if we have a new contributor who wants a good, small first task

JoshMarler:
@nick-thompson, final two points actioned. Squashed changes again so good to go.

nick-thompson:
Merged! This is super cool, thanks for all your help Josh. As we continue building out the remaining API, I'd love to get a cool `CanvasExample` project going to show this feature off too :)

JoshMarler:
@nick-thompson, Yeah a canvas example project would be awesome. Totally up for assisting with that once some more of the API is implemented and DOCS added. 

I have a little TODO list compiling. The below are my next tasks but I'll probably complete these after I've finished the HotReload refactor.
                                                                                                                                                  
. Get fillText() strokeText(), font and textAlign working.                                                                                                                                                                                                                                                                                                                                                                            
. Get drawImage() working with <svg> string.                                                                                                                                                                                                                                                                                                                                                                                  
. Get gradients and stroke patterns (i.e. line - dash) working.

Issue #27, opened 2019-12-07
Refactor a HotReloader out of BlueprintGenericEditor


nick-thompson:
The `BlueprintGenericEditor` implements hot reloading by polling the file system to see if the js bundle has changed. It's super simple, and there's no reason that it has to be tied to a notion of a plugin editor (it should be usable no matter what type of app you're using). So I think just pulling that out into a separate class which can be composed wherever you need it makes sense.

```
HotReloader(std::unique_ptr<ReactApplicationRoot>& appRoot, const juce::File& bundle)
```

Just run the timer, check the bundle, and `reset()` the `appRoot` there. Then BlueprintGenericEditor just composes the HotReloader for this functionality

JoshMarler:
Hey @nick-thompson, This framework is so flipping good. I'm actually working on this as part of a project I'm working on at my current shop. If you're still after this I can drop over a few pull request with some of my changes when I'm satisfied things are working well. 

I also have a few GCC fixes as I've been compiling on Linux

JoshMarler:
As an aside. Presumably we need to notify clients of HotReloader at two points. Going off of the example code in BlueprintGenericEditor::assignNewAppRoot:

1. When resetting the appRoot unique_ptr so that clients can then call registerNativeMethod etc.
2. Once appRoot->evaluate(bundleFile) has been called so clients can addAndMakeVisible etc.

i.e. Something along the lines of the below:
`    void HotReloader::timerCallback()
    {
        const auto lmt = bundleFile.getLastModificationTime();

        if (lmt > bundleLastModifiedTime)
        {
            // Sanity check
            jassert(bundleFile.existsAsFile());

            // Assign a fresh appRoot
            appRoot = std::make_unique<ReactApplicationRoot>();
            listeners.call([=](Listener &l) { l.ReloadingAppRootBundle(appRoot.get(), bundleFile); });

            appRoot->evaluate(bundleFile.getFullPathName());
            listeners.call([=](Listener &l) { l.AppRootBundleReloaded(appRoot.get(), bundleFile); });

            // At this point the appRoot may have been removed due to an error evaluating
            // the bundle, so we check for that case and halt if necessary
            if (!appRoot)
                return;

            bundleLastModifiedTime = lmt;
        }
    }`

Where the pointer passed to AppRootBundleReloaded is potentially nullptr in the event of an error (Specified in the Documentation/Contract of the class)

[...]

Issue #24, opened 2019-09-09
Style Guide / Code Formatting


tobanteAudio:
For future contributions, it would probably be very helpful to have a common style guide for source files (JavaScript, C++ & Markdown). 

For C++ the only real option is `clang-format`, which is really nice once you find a config that you like.

For JavaScript and Markdown, there are more options. The one that I'm familiar with is `Prettier`, which does both JS & MD.

We could add a script which runs both tools on the source files.

Is this something you are interested in?

nick-thompson:
💯 absolutely. I think that would be helpful even just for me, but also to lower the overhead for any interested contributors.

To be honest, I'm not very well versed in either Prettier or clang-format, but I've seen them both around a bunch. I've tried to keep pretty close to the JUCE code style in my C++ but my JavaScript formatting has just been habitual. If we can roughly formalize the existing style into a config setting for clang-format and prettier then I think that'd be perfect. Do you want to give it a try?

JoshMarler:
@nick-thompson , definitely needed for beta?

nick-thompson:
100%, let's add it

nick-thompson:
Quick update here, @olidacombe has some work with Prettier going on the js side in #191.

This morning, I'm looking at clang-tidy and clang-format to cover the C++ side. Should be pretty easy to integrate these things into our workflow, but a prerequisite for clang-format seems to be picking and adhering to a .clang-format style. I'm not very well-versed in these, so figured I'd ask here: anybody familiar with some .clang-format options that land pretty close to the way the code is generally formatted in our codebase already?

JoshMarler:
Not sure if you've seen this Nick but worth a look: https://forum.juce.com/t/automatic-juce-like-code-formatting-with-clang-format/31624/7

Granted we might not want to match juce's coding style exactly.

nick-thompson:
Excellent, I'll poke at that later this week! Thanks

olidacombe:
I am finding myself thinking about style guidance again in the context of props defined for core js components. `TextInputProps` is a good example, where there's a mixture of cases for the props (some are camelCase, some kebab-case).  Although it's perhaps not worth losing sleep over, I think having consistency in the interface exposed to clients has value.  What we've done with css props is convert to kebab-case when crossing to C++, so clients have a choice of case, and we use camelCase in the examples.  If we want to "go consistent", 2 options are:

[...]

PR #54, opened 2020-05-27
Initial refactor of hot reloading functionality.

JoshMarler:
@nick-thompson, here is some initial changes to move hot-reload into ReactApplicationRoot. 

Couple of questions. 

1. Do we need the evaluate(const juce::String& bundle) overload? I can see it could be useful to 
   evaluate inline JS code in things like unit tests etc but is there a strong need for this over using 
   juce::File ?

2. Should ReactApplicationRoot handle painting errors by default, only in debug or not at all? i.e. 
    should client code handle this? I think it's nice for ReactApplicationRoot to handle this.

Also worth mentioning, despite our original convo here: https://github.com/nick-thompson/blueprint/issues/27

I don't think it's actually useful  for us to have a before and after bundle eval callback. We just need after. This is because client code will no longer be tearing down the entire appRoot in the event of a reload/error. 

If you're happy with the overall shape of this I will refactor GenericEditor and the example apps to better demonstrate usage of the new shape. 

FYI I've tried to allow for multiple bundles as per our earlier conversation. I haven't tested this in earnest yet but will do so. I need to find some sensible polyfill to test with or something. Everything else appears to be working, hot-reload and error reporting/painting are both good.

### DIFF
--- a/blueprint/core/blueprint_GenericEditor.cpp
+++ b/blueprint/core/blueprint_GenericEditor.cpp
@@ -21,14 +21,14 @@ namespace blueprint
         for (auto& p : processor->getParameters())
             p->addListener(this);
 
-        // Now we can provision the app root
-        assignNewAppRoot(code);
+        // Setup the ReactApplicationRoot callbacks and evaluate the supplied JS code/bundle
+        registerAppRootCallbacks();
+        appRoot.evaluate(code);
 
-        // If an error showed up, our appRoot is already gone
-        if (appRoot)
-            addAndMakeVisible(appRoot.get());
+        // Add ReactApplicationRoot as child component
+        addAndMakeVisible(appRoot);
 
-        // Set an arbitrary size, should be overriden from outside the constructor
+        // Set an arbitrary size, should be overridden from outside the constructor
         setSize(400, 200);
     }
 
@@ -38,30 +38,24 @@ namespace blueprint
         // Sanity check
         jassert (bundle.existsAsFile());
         bundleFile = bundle;
-        lastModifiedTime = bundleFile.getLastModificationTime();
 
         // Bind parameter listeners
         for (auto& p : processor->getParameters())
             p->addListener(this);
 
-        // Now we can provision the app root
-        assignNewAppRoot(bundle.loadFileAsString());
+        // Setup the ReactApplicationRoot callbacks and evaluate the supplied JS code/bundle
+        registerAppRootCallbacks();
+        appRoot.evaluate(bundleFile);
 
-        // If an error showed up, our appRoot is already gone
-        if (appRoot)
-            addAndMakeVisible(appRoot.get());
+        // Add ReactApplicationRoot as child component
+        addAndMakeVisible(appRoot);
 
-        // Kick off the timer that polls for file changes
-        startTimer(50);
-
-        // Set an arbitrary size, should be overriden from outside the constructor
+        // Set an arbitrary size, should be overridden from outside the constructor
         setSize(400, 200);
     }
 
     BlueprintGenericEditor::~BlueprintGenericEditor()
     {
-        stopTimer();
-
         for (auto& p : processor.getParameters())
         {
             p->removeListener(this);
@@ -71,8 +65,6 @@ namespace blueprint
     //==============================================================================
     void BlueprintGenericEditor::parameterValueChanged (int parameterIndex, float newValue)
     {
-        Component::SafePointer<blueprint::ReactApplicationRoot> safeAppRoot (appRoot.get());
-
         // Collect some information about the parameter to push into the engine
         const auto& p = processor.getParameters()[parameterIndex];
         juce::String id = p->getName(100);
@@ -86,17 +78,14 @@ namespace blueprint
         // Dispatch parameter value updates to the javascript engine at 30Hz
         throttleMap.throttle(parameterIndex, 1000.0 / 30.0, [=]() mutable {
             juce::MessageManager::callAsync([=]() mutable {
-                if (safeAppRoot)
-                {
-                    safeAppRoot->dispatchEvent(
-                        "parameterValueChange",
-                        parameterIndex,
-                        id,
-                        defaultValue,
-                        newValue,
-                        stringValue
-                    );
-                }
+                appRoot.dispatchEvent(
+                    "parameterValueChange",
+                    parameterIndex,
+                    id,
+                    defaultValue,
+                    newValue,
+                    stringValue
+                );
             });
         });
     }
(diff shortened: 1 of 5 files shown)


blueprint/core/blueprint_ReactApplicationRoot.h
```diff
@@ -56,12 +118,69 @@ namespace blueprint
...
+
+            if (hotReloadEnabled)
+            {
+                jassert(bundleWatcher);
+                bundleWatcher->watch(bundle);
+            }
+
+            auto result = evaluate(bundle.loadFileAsString());
+
+            //TODO: Should we guard against users calling evaluate on public engine member directly and missing
```
JoshMarler, line 178: @nick-thompson, this is a TODO that is probably worth you commenting on.

blueprint/core/blueprint_ReactApplicationRoot.h
```diff
@@ -44,8 +95,18 @@ namespace blueprint
             // Install React.js backend rendering methods
             registerNativeRenderingHooks();
 
-            // And install view types
-            installNativeViewTypes();
+            // Initialise the ViewManager
+            initViewManager();
+
+#if JUCE_DEBUG
+            //TODO: Keep error handler here in JUCE_DEBUG builds only or should
```
nick-thompson, line 102: Good question. I'd say let's always install this handler, because it's always up to the user if they want to override it in their release builds or something

blueprint/core/blueprint_ReactApplicationRoot.h
```diff
@@ -44,8 +95,18 @@ namespace blueprint
...
+            //      client code be responsible for handling any errors?
+            engine.onUncaughtError = [this](const juce::String& msg, const juce::String& trace) {
+                handleBundleError(trace);
+            };
+
+            enableHotReloading(true);
```
nick-thompson, line 108: Enabling this in `JUCE_DEBUG` :+1:

blueprint/core/blueprint_ReactApplicationRoot.h
```diff
@@ -56,12 +117,69 @@ namespace blueprint
...
         {
             JUCE_ASSERT_MESSAGE_THREAD
-            return engine.evaluate(bundle);
+
+            // Clear error state from previous js evals
+            errorText.reset();
+            auto result = engine.evaluate(bundle);
+
+            if (onBundleEvaluated)
+                onBundleEvaluated(std::nullopt);
```
nick-thompson, line 150: :+1:

Yea we definitiely want this method around still. In develeopment builds it's always preferable to just read the the bundle file from disk, which enables things like hot reloading and enables skipping a compile step to get webpack's latest output. But for shipping a product, what I've been doing is injecting the production bundle as a Binary Resource in the projucer, and reading it from memory as a string at runtime

blueprint/core/blueprint_ReactApplicationRoot.h
nick-thompson: Hm... good question. I think it's probably a "things you should know" section in the docs. I could imagine cases where you might want to evaluate something but don't care for any callbacks, especially in apps where almost all of the logic is done in js-land and callbacks to native aren't really important

blueprint/core/blueprint_ReactApplicationRoot.h
```diff
@@ -99,10 +217,108 @@ namespace blueprint
             return _viewManager.get();
         }
 
+        //==============================================================================
+        /**
+         * Enables "hot-reload" functionality so that ReactApplicationRoot watches JS bundle files
+         * added via evaluate() for changes. When one of these bundle files changes (i.e. from a Webpack watch/rebuild)
+         * ReactApplicationRoot will reload the bundle and update the View/ShadowView tree.
+         **/
+        void enableHotReloading(bool enable)
```
nick-thompson, line 226: Forgive me for maybe being nitpicky, but rather than boolean arguments in this type of situation I prefer a second method: `void enableHotReloading()` and `void disableHotReloading()`. To me it just makes the callsite so much more clear about what your intention is. Then, as I think about it, would we ever really even want a `disableHotReloading`? I can imagine in release mode not starting hot reloading, but I don't think there would be too many cases where you would start, and then want to stop, the hot reloader? What do you think?

So maybe then my ask here would be just remove the boolean argument and

```
void enableHotReloading()
{
    bundleWatcher = std::make_unique<etc>(etc);
    hotReloadingEnabled = true;
}
```

blueprint/core/blueprint_ReactApplicationRoot.h
```diff
@@ -99,10 +217,108 @@ namespace blueprint
...
+         *             }
+         *         }
+         *     };
+         *
+         *     appRoot->evaluate(myAppBundle);
+         * }
+         *
+         * @endcode
+         **/
+        BundleEvalCallback onBundleEvaluated;
```
nick-thompson, line 286: :+1: Love the inline documentation

blueprint/core/blueprint_ReactApplicationRoot.h
JoshMarler: Yeah sounds like a good plan.

blueprint/core/blueprint_ReactApplicationRoot.h
JoshMarler: Right cool. My current use case installs the bundle on the target system along with our other application resources so I was missing this use-case. Cool. Will leave this as is

blueprint/core/blueprint_ReactApplicationRoot.h
JoshMarler: Yep fair point. I don't really think it is necessary either.

JoshMarler:
> Excellent work, I love the shape this is taking. As I'm reading over the code it makes me think that after we merge this work, we could just delete the `BlueprintGenericEditor`, right? The only added value remaining will be the parameter handling stuff, which people will probably want to do their own way anyway
> 
> Anyways, a couple quick ideas/asks inline, and then I have two more thoughts:
> 
> * I think that during the hot-reload "reset" we'll also need to throw away the existing `EcmascriptEngine` and make a new one. My understanding of Duktape's error case is that once we hit it, any further executing of javascript code can potentially result in undefined behavior. So I think we'll probably want to play a game similar to what we're doing here for the ViewManager with a `std::unique_ptr<EcmascriptEngine>` and resetting it on hot reload
> * As a follow-up I think this makes a case for having a `beforeBundleEval` callback: pretty much only as an opportunity to install custom error handlers on the freshly-made EcmascriptEngine before evaluating your app bundle
> 
> Let me know what you think, and thanks as always for your work @JoshMarler!

Right OK thanks @nick-thompson. I wasn't too sure whether we could get away with leaving the EcmaScriptEngine initialised. That makes sense to me. I'll rework things to fit with this shape.

JoshMarler:
> Excellent work, I love the shape this is taking. As I'm reading over the code it makes me think that after we merge this work, we could just delete the `BlueprintGenericEditor`, right? The only added value remaining will be the parameter handling stuff, which people will probably want to do their own way anyway
> 
> Anyways, a couple quick ideas/asks inline, and then I have two more thoughts:
> 
> * I think that during the hot-reload "reset" we'll also need to throw away the existing `EcmascriptEngine` and make a new one. My understanding of Duktape's error case is that once we hit it, any further executing of javascript code can potentially result in undefined behavior. So I think we'll probably want to play a game similar to what we're doing here for the ViewManager with a `std::unique_ptr<EcmascriptEngine>` and resetting it on hot reload
> * As a follow-up I think this makes a case for having a `beforeBundleEval` callback: pretty much only as an opportunity to install custom error handlers on the freshly-made EcmascriptEngine before evaluating your app bundle
> 
> Let me know what you think, and thanks as always for your work @JoshMarler!

@nick-thompson, I'm just finishing this off. I wonder if it actually makes sense to keep BlueprintGenericEditor? I think there is some value in a generic example/helper for binding parameter changes to an AudioProcessorValueTreeState instance to get people up and running quickly. It also seems like a good place to provide a concrete example of using the beforeBundleEval and afterBundleEval callbacks. i.e. register the param changed methods in beforeBundleEval and dispatch the initial/current parameter values via BlueprintGenericEditor::parameterValueChanged in afterBundleEval.

JoshMarler:
@nick-thompson, just need to fixup the sample apps and then this should be good to go subject to your review points.

examples/GainPlugin/Source/PluginEditor.cpp
```diff
@@ -81,35 +81,35 @@ GainPluginAudioProcessorEditor::GainPluginAudioProcessorEditor (GainPluginAudioP
 
     // Lastly, start our timer for reporting meter values
     startTimerHz(30);
-
+
```
JoshMarler, line 84: Apologies @nick-thompson, not sure what happened here. I jumped to mac to check example apps were working (have been working on Linux) and I think my editor must have converted some line endings.

JoshMarler:
@nick-thompson, OK hopefully this is ready for a final review. Both the example apps appear to be working.

blueprint/core/blueprint_GenericEditor.cpp
```diff
@@ -21,14 +21,14 @@ namespace blueprint
...
+        appRoot.evaluate(code);
 
-        // If an error showed up, our appRoot is already gone
-        if (appRoot)
-            addAndMakeVisible(appRoot.get());
+        // Add ReactApplicationRoot as child component
+        addAndMakeVisible(appRoot);
 
-        // Set an arbitrary size, should be overriden from outside the constructor
+        // Set an arbitrary size, should be overridden from outside the constructor
```
nick-thompson, line 31: :)

blueprint/core/blueprint_GenericEditor.cpp
```diff
@@ -86,17 +78,14 @@ namespace blueprint
         // Dispatch parameter value updates to the javascript engine at 30Hz
         throttleMap.throttle(parameterIndex, 1000.0 / 30.0, [=]() mutable {
             juce::MessageManager::callAsync([=]() mutable {
-                if (safeAppRoot)
```
nick-thompson, line 89: Ahh interesting, don't need this anymore because the appRoot doesn't get swapped out itself anymore. Nice!

examples/GainPlugin/Source/PluginEditor.cpp
nick-thompson: No worries! This kinda thing would be solved by #24 anyway... getting some kind of clang-format thing going so that anybody who wants to contribute doesn't have to worry about manally upholding any style guide or manually change their editor line endings or anything. Just clang-format before merge && :+1:. But until we have that I'm quite relaxed about things like this!

blueprint/core/blueprint_GenericEditor.cpp
JoshMarler: Ha Jetbrains IDE's do yellow squiggly thinks for spell correct. It's actually a bit annoying ...

PR #55, opened 2020-05-28
For CanvasContext drawing operations use highest sampling quality

JoshMarler:
@nick-thompson, I hit some issues with rendering quality. Given CanvasContext is responsible for drawing UI elements I think it makes sense for it's Graphics instance to use the highest image sampling quality possible.

### DIFF
--- a/blueprint/core/blueprint_CanvasView.h
+++ b/blueprint/core/blueprint_CanvasView.h
@@ -101,6 +101,7 @@ namespace blueprint
         void init()
         {
             graphics = std::make_unique<juce::Graphics>(image);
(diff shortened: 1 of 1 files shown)


JoshMarler:
It may be worth comparing this again direct drawing to the View's graphics context still. Slightly concerned using an Image may introduce rendering quality issues.

JoshMarler:
@nick-thompson. In fact, maybe ignore this for a moment. I think something else is going on in my app.

nick-thompson:
Ah ok, interesting. Let me know if you find anything!

PR #56, opened 2020-06-01
Fix canvas render issues

JoshMarler:
@nick-thompson, the replacement/reset of the juce::Image instance used in CanvasContext::init() resolved the rendering issues I have been having. Hopefully we can get this in ASAP!

### DIFF
--- a/blueprint/core/blueprint_CanvasView.h
+++ b/blueprint/core/blueprint_CanvasView.h
@@ -117,8 +117,6 @@ namespace blueprint
                         const auto colourHex = args.arguments[0].toString();
                         properties.fillStyle.setColour(juce::Colour::fromString(colourHex));
 
-                        graphics->setFillType(properties.fillStyle);
-
                         return juce::var();
                     }
             });
@@ -182,6 +180,7 @@ namespace blueprint
                         const int width = args.arguments[2];
                         const int height = args.arguments[3];
 
+                        graphics->setFillType(properties.fillStyle);
                         graphics->fillRect(x, y, width, height);
 
                         return juce::var();
@@ -198,6 +197,7 @@ namespace blueprint
                         const int width = args.arguments[2];
                         const int height = args.arguments[3];
 
+                        graphics->setColour(properties.strokeStyle.colour);
                         graphics->drawRect(x, y, width, height, properties.lineWidth);
 
                         return juce::var();
@@ -317,8 +317,6 @@ namespace blueprint
                         jassert(graphics);
                         jassert(args.numArguments == 0);
 
-                        graphics->setColour(properties.strokeStyle.colour);
-                        graphics->strokePath(path, juce::PathStrokeType((float)properties.lineWidth));
                         graphics->setColour(properties.fillStyle.colour);
                         graphics->fillPath(path);
 
--- a/blueprint/core/blueprint_CanvasView.h
+++ b/blueprint/core/blueprint_CanvasView.h
@@ -70,8 +70,9 @@ namespace blueprint
         };
 
         explicit CanvasContext()
-            : image(juce::Image::ARGB, 1, 1, true)
-            , graphics(std::make_unique<juce::Graphics>(image))
+            : ctxWidth(1)
+            , ctxHeight(1)
+            , image(juce::Image::ARGB, ctxWidth, ctxHeight, true)
         {
             registerNativeProperties();
             registerNativeFunctions();
@@ -86,7 +87,9 @@ namespace blueprint
          */
         void setSize(int width, int height)
         {
-            image = juce::Image(juce::Image::ARGB, width, height, true);
+            ctxWidth  = std::max(width, 1);
+            ctxHeight = std::max(height, 1);
+            image     = image.rescaled(ctxWidth, ctxHeight, juce::Graphics::highResamplingQuality);
         }
 
         /** Get the juce::Image instance that has been rendered to. This can be drawn to the screen
@@ -100,7 +103,11 @@ namespace blueprint
         /**  init should be called before passing the CanvasContext instance to a JS drawing function. */
         void init()
         {
-            graphics = std::make_unique<juce::Graphics>(image);
+            graphics   = std::make_unique<juce::Graphics>(image);
+            properties = Properties{};
+            path       = juce::Path{};
+
+            transformStack.clear();
         }
 
     private:
@@ -204,7 +211,22 @@ namespace blueprint
                     }
             });
 
-            //TODO: Add clearRect. Use View background color and g.fillRect?
+            setProperty("clearRect", juce::var::NativeFunction {
+                    [=](const juce::var::NativeFunctionArgs& args) -> juce::var {
+                        jassert(graphics);
+                        jassert(args.numArguments == 4);
+
+                        const int x = args.arguments[0];
+                        const int y = args.arguments[1];
+                        const int width = args.arguments[2];
+                        const int height = args.arguments[3];
+
+                        juce::Rectangle<int> area (x, y, width, height);
+                        image.clear(area);
+
+                        return juce::var();
+                    }
+            });
 
             // Path functions
             setProperty("beginPath", juce::var::NativeFunction {
@@ -436,6 +458,9 @@ namespace blueprint
         }
 
         //==============================================================================
+        int                                ctxWidth;
+        int                                ctxHeight;
+
         juce::Image                        image;
         std::unique_ptr<juce::Graphics>    graphics;
         juce::Path                         path;
@@ -510,6 +535,7 @@ namespace blueprint
 
         void resized() override
         {
+            View::resized();
             ctx->setSize(getWidth(), getHeight());
         }
 
--- a/packages/juce-blueprint/esm/Blueprint.js
+++ b/packages/juce-blueprint/esm/Blueprint.js
@@ -22,52 +22,86 @@ export function Image(props) {
   return React.createElement('Image', props, props.children);
 }
 
-export function Canvas(props) {
-  if (props.hasOwnProperty('onDraw')) {
-    let userOnDraw = props.onDraw;
-    let wrappedDraw = function(ctx) {
-      Object.defineProperty(ctx, 'fillStyle', {
-        enumerable: false,
-        configurable: false,
-        get: function() {
-          return 'Not Supported';
-        },
-        set: function(value) {
-          this.__setFillStyle(value);
-        }
-      });
+export function bindCanvasContextProperties(ctx) {
+  Object.defineProperty(ctx, 'fillStyle', {
+    enumerable: false,
+    configurable: false,
+    get: function() {
+      return 'Not Supported';
+    },
+    set: function(value) {
+      this.__setFillStyle(value);
+    }
+  });
 
-      Object.defineProperty(ctx, 'strokeStyle', {
-        enumerable: false,
-        configurable: false,
-        get: function() {
-          return 'Not Supported';
-        },
-        set: function(value) {
-          this.__setStrokeStyle(value);
-        }
-      });
+  Object.defineProperty(ctx, 'strokeStyle', {
+    enumerable: false,
+    configurable: false,
+    get: function() {
+      return 'Not Supported';
+    },
+    set: function(value) {
+      this.__setStrokeStyle(value);
+    }
+  });
 
-      Object.defineProperty(ctx, 'lineWidth', {
-        enumerable: false,
-        configurable: false,
-        get: function() {
-          return 'Not Supported';
-        },
-        set: function(value) {
-          this.__setLineWidth(value);
-        }
-      });
+  Object.defineProperty(ctx, 'lineWidth', {
+    enumerable: false,
+    configurable: false,
+    get: function() {
+      return 'Not Supported';
+    },
+    set: function(value) {
+      this.__setLineWidth(value);
+    }
+  });
+}
 
-      return userOnDraw(ctx);
+export class Canvas extends Component {
+  constructor(props) {
+    super(props);
+
+    this._onMeasure = this._onMeasure.bind(this);
+    this._onDraw = this._onDraw.bind(this);
+
+    this.state = {
+      width: 0,
+      height: 0
     };
-
-    return React.createElement('CanvasView', Object.assign({}, props, {
-      onDraw: wrappedDraw,
-    }), props.children);
   }
 
-  return React.createElement('CanvasView', props, props.children);
+  _onMeasure(width, height) {
+    this.setState({
+      width: width,
+      height: height
+    });
+
+    if (typeof this.props.onMeasure === 'function') {
+      this.props.onMeasure(width, height);
+    }
+  }
+
+  _onDraw(ctx) {
+    if (typeof this.props.onDraw === 'function') {
+      bindCanvasContextProperties(ctx);
+
+      if (this.props.autoclear) {
+        ctx.clearRect(0, 0, this.state.width, this.state.height);
+      }
+
+      this.props.onDraw(ctx);
+    }
+  }
+
+  render() {
+    //TODO: Check whether need to use below arrow function for "this" binding
+    //      is a bug in duktape. Possible this only occurs on linux. Does not 
+    //      appear to occur on mac.
+    return React.createElement('CanvasView', Object.assign({}, this.props, {
+      onDraw: (ctx) => { this._onDraw(ctx) },
+      onMeasure: (width, height) => { this._onMeasure(width, height )}
+    }), this.props.children);
+  }
 }
 
 function ScrollViewContentView(props) {


nick-thompson:
Ahh of course, makes total sense. One thing I'm thinking though is that when you're working with a `<canvas>` in web and using a `window.requestAnimationFrame` callback, for example, the canvas state is maintained across invocations of your callback. So you might like take a screenshot of the canvas state, move it left 1px, then erase and redraw the rightmost 1px if you're doing a scrolling waveform view or something. Here, if we use this `init` before invoking the callback, state is always lost, right?

I'm wondering if the more "API parity" approach would be to implement `clearRect` (https://developer.mozilla.org/en-US/docs/Web/API/CanvasRenderingContext2D/clearRect) either by resetting the internal image (or more likely using this guy https://docs.juce.com/master/classImage.html#a2e8dc6de9ca6b4181658d0144579fb2f).

Then in your js callback you just `ctx.clearRect(0, 0, 100, 100);` at the top of your routine and I believe it would resolve the same issue you're having while being more in line with how you would interact with a web <canvas>. What do you think?

JoshMarler:
> Ahh of course, makes total sense. One thing I'm thinking though is that when you're working with a `<canvas>` in web and using a `window.requestAnimationFrame` callback, for example, the canvas state is maintained across invocations of your callback. So you might like take a screenshot of the canvas state, move it left 1px, then erase and redraw the rightmost 1px if you're doing a scrolling waveform view or something. Here, if we use this `init` before invoking the callback, state is always lost, right?
> 
> I'm wondering if the more "API parity" approach would be to implement `clearRect` (https://developer.mozilla.org/en-US/docs/Web/API/CanvasRenderingContext2D/clearRect) either by resetting the internal image (or more likely using this guy https://docs.juce.com/master/classImage.html#a2e8dc6de9ca6b4181658d0144579fb2f).
> 
> Then in your js callback you just `ctx.clearRect(0, 0, 100, 100);` at the top of your routine and I believe it would resolve the same issue you're having while being more in line with how you would interact with a web . What do you think?

So, I had juce::Image::clear() in init() initially and it works fine. Totally get what you're getting at, my only reservation would be that this requires devs (who may not care about web compatibility) to remember to clearRect() in each of their draw routines to ensure rendering quality (A side-effect of the fact that we paint to an Image internally). This comes for free with CanvasView. Is this a big issue though?

Part of me thinks the fewer "remember this thing" requirements the better, is there any sensible way we can give this to user code "for free" without breaking state in the manner you've discussed?

nick-thompson:
Hmm yea, totally hear you. The way I've been thinking about it was that people coming from javascript/React probably have some familiarity with web stuff, and the more Blueprint feels and acts like that web stuff, the fewer "remember this thing" requirements we have. I'd be interested to know if there's more people coming the other direction– dipping their toes in React because they want a better story for GUI development coming from JUCE.

How about we do this: implement `CanvasView` and the `CanvasContext` such that we only clear the image when you call `clearRect` from javascript. This way we're maximally compliant with the web API at that level. Then, in javascript, we support a prop "autoclear" or "disableautoclear" that will automatically clear the canvas for you:

```
export function Canvas(props) {
  if (props.hasOwnProperty('onDraw')) {
    let userOnDraw = props.onDraw;
    let wrappedDraw = function(ctx) {
      Object.defineProperty(ctx, 'fillStyle', {
        enumerable: false,
        configurable: false,
        get: function() {
          return 'Not Supported';
        },
        set: function(value) {
          this.__setFillStyle(value);
        }
      });

      // Right here we can automagically clear the canvas
      if (props.autoclear) {
        ctx.clearRect(0, 0, w, h);
      }

      return userOnDraw(ctx);
    };

    return React.createElement('CanvasView', Object.assign({}, props, {
      onDraw: wrappedDraw,
    }), props.children);
  }

  return React.createElement('CanvasView', props, props.children);
}
```

This way automatic clearing behavior is pretty harmlessly opt-out (or opt-in) based on the way the code is structured. If you want web compatibility, turn that flag on. Or if you don't care about web compat but you're trying to do a scrolling waveform view, you'll care about that flag

The only detail here is that `Canvas` at the js layer will have to catch its own width/height to use for the call to clearRect, similar to how GainPlugin does: https://github.com/nick-thompson/blueprint/blob/master/examples/GainPlugin/Source/jsui/src/Slider.js#L50

What do you think?

JoshMarler:
" I'd be interested to know if there's more people coming the other direction– dipping their toes in React because they want a better story for GUI development coming from JUCE."

That is certainly the case on our end and I suspect would be so for many others because C++ GUI is such a pain and Blueprint is just SO much better!

I think using clearRect in the Canvas wrapped draw code is totally fine and we should definitely have a clearRect implementation regardless. The bit that has me curious is the compatibility back and forth between a web/dom backed react app and a juce backed app. 

So in the juce case, rendering is basically broken without clearing/resetting the image between each paint/onDraw call (For now at least, it has to be cleared or the UI looks a right mess!). So, if we take your ```window.requestAnimationFrame``` example,  the concern is loosing state, i.e. people familiar with <canvas> on the web may be surprised by their code's behaviour rendering via juce, totally understand the desire to avoid/minimise this! However, if people have code/use-cases expecting preserving of the image state between draw functions/callbacks, how would this be expected to work without hitting the rendering problems under juce (I'm very likely missing part of your point btw!)? 

i.e. If I performed some animation/snapshot which shifted the existing juce::Image left by 1 pixel and then draws 1 new pixel at the rightmost edge am I going to hit quantisation errors still? 

I'm still trying to wrap my head around how the same react app code could target both web/dom and juce renderers with minimal "special case" code.... I've been away from web dev for too long :-(

JoshMarler:
@nick-thompson, more than happy to get your idea implemented by the way! Just curious to know how you see someone using the Canvas View/Element in a way that would be cross compatible between Web and Juce with the way I have it implemented now.

nick-thompson:
Thinking about this a bunch @JoshMarler, thanks for the discussion.

I'm wondering, specifically what causes the quantization errors. I saw the quote of the post you mentioned from Jules on the forum, but if I understand that correctly, you would only see these rendering artifacts if you're repeatedly applying the same draw routine to the same image over and over again. So, in the example of writing a draw function that snapshots the canvas state, translates it left 1px, and redraws the rightmost 1px, the image state itself should never hit this rendering issue _if_ you clearRect() after taking the snapshot. Right? So then I'm wondering if the rendering artifacts you're seeing come from the `paint` method of `CanvasView` always pasting the same image to the same rect on screen. Maybe our `paint()` method should `fillAll(getLocalBounds())` with a `0x00000000` transparent before drawing the image?

I totally agree with you that there are kind of two separate issues here: (1) preserving the quality of the draw operation and (2) parity with the web API. I think (1) takes priority over (2), and if we can find a way to implement (2) without sacrificing (1) then it's ideal

JoshMarler:
> Thinking about this a bunch @JoshMarler, thanks for the discussion.
> 
> I'm wondering, specifically what causes the quantization errors. I saw the quote of the post you mentioned from Jules on the forum, but if I understand that correctly, you would only see these rendering artifacts if you're repeatedly applying the same draw routine to the same image over and over again. So, in the example of writing a draw function that snapshots the canvas state, translates it left 1px, and redraws the rightmost 1px, the image state itself should never hit this rendering issue _if_ you clearRect() after taking the snapshot. Right? So then I'm wondering if the rendering artifacts you're seeing come from the `paint` method of `CanvasView` always pasting the same image to the same rect on screen. Maybe our `paint()` method should `fillAll(getLocalBounds())` with a `0x00000000` transparent before drawing the image?
> 
> I totally agree with you that there are kind of two separate issues here: (1) preserving the quality of the draw operation and (2) parity with the web API. I think (1) takes priority over (2), and if we can find a way to implement (2) without sacrificing (1) then it's ideal

Hey @nick-thompson !

Yeah, I wouldn't expect the issue to occur in the snapshot example if you were calling clearRect. I'd have to test clearing the rectangle we render the image to in Canvas::paint(). 

My understanding was that we are currently persisting the Image instance between calls to the various draw functions (strokePath, lineTo etc.), so we are effectively repeatedly drawing the same lines/paths to the Image instance and quantisation errors occur which then results in slight pixel variations (noticeable mainly on strokes/outlines etc.). 

Filling / Replacing the contents of the target rect may work but if we're still rendering to the same Image instance in CanvasContext I would have thought we would still see the problem. I may well be misunderstanding how painting to an Image via juce::Graphics works so I'll give this a test now and let you know the result!

nick-thompson:
Right, ok. No I think we're both on the same page! Let me try something, I'll post back in a bit

JoshMarler:
@nick-thompson, yeah so clearing the target rectangle we render the image to doesn't appear to make any odds which sort of makes sense to me.

nick-thompson:
Aha, update wiith circles: https://jsfiddle.net/ncthom91/s32pdzye/19/

Is that similar to what you're seeing?

JoshMarler:
@nick-thompson, Yep! This is what I'm seeing and incidentally with arc functions drawing rounded rectangles etc.

nick-thompson:
> @nick-thompson, Yep! This is what I'm seeing and incidentally with arc functions drawing rounded rectangles etc. Thanks for that, very informative. I have learnt a thing!

Great, same for me! :)

> So based off of this example are we saying clearRect() is indeed the preferred approach as we maintain parity with the actual result in the browser? i.e. both rendering engines/approaches would suffer the same issue. In which case your original autoclear property idea is possibly a good fit?

Yea, I think so. And we'll write in the documentation that the supported properties of the CanvasView include this "autoclear" prop and we can write a bit about our findings here to explain why it's useful. Hopefully that makes it feel less like a "things you should know" and more like a "ah neat that's helpful." Sound good?

JoshMarler:
Yep awesome sounds good to me. I'll take a crack at getting this all finished tomorrow then. Cheers.

JoshMarler:
@nick-thompson, just working on this now. Having a little property adding the onMeasure prop/callback to the Canvas/CanvasView. Any ideas on the best way to implement that part?

JoshMarler:
@nick-thompson, just pushed the whole change up so you have a bit more context: https://github.com/nick-thompson/blueprint/commit/63d080025f19bb4c361cb79fa4ed70810a561cdb

nick-thompson:
Awesome, looking over it all now!

blueprint/core/blueprint_CanvasView.h
```diff
@@ -86,7 +87,9 @@ namespace blueprint
          */
         void setSize(int width, int height)
         {
-            image = juce::Image(juce::Image::ARGB, width, height, true);
+            ctxWidth  = width;
+            ctxHeight = height;
+            image     = image.rescaled(ctxWidth, ctxHeight, juce::Graphics::highResamplingQuality);
```
nick-thompson, line 92: Ohhh wow, great catch :+1:

packages/juce-blueprint/esm/Blueprint.js
```diff
@@ -22,52 +22,85 @@ export function Image(props) {
...
-        get: function() {
-          return 'Not Supported';
-        },
-        set: function(value) {
-          this.__setLineWidth(value);
-        }
-      });
-
-      return userOnDraw(ctx);
+export function bindCanvasContextProperties(ctx) {
```
nick-thompson, line 25: :+1: definitely. This `export` here marks this method as part of the "public" blueprint js API, and I'm not sure we want this to be publicly accessible, but we definitely want it to be shared and shareable internally.

But, let's not worry about that for now. Once the native side of blueprint is more complete (i.e. as we approach a "beta" release or something) we will want to take a pass at what our js API should look like

packages/juce-blueprint/esm/Blueprint.js
```diff
@@ -22,52 +22,85 @@ export function Image(props) {
...
+        return userOnDraw(ctx);
+      }
+
+      canvas = React.createElement('CanvasView', Object.assign({}, props, {
+        onDraw: wrappedDraw
+      }), props.children);
+    }
+
+    return (
+        <View {...this.props} onMeasure={this._onMeasure}>
```
nick-thompson, line 99: Hmm... you shoudn't need to wrap this with a View. As long as you put the `onMeasure` prop on the canvas you should get the callback.

nick-thompson: Ah, got it! Check out the `CanvasView::resized()` method– that override should call the base class `View::resized()` and then do its `ctx->setSize` (as we do in the CanvasView::paint). The View base class does all the usual business of calling the `onMeasure` callback and painting things like borders and background colors

packages/juce-blueprint/esm/Blueprint.js
JoshMarler: Yeah. I basically have an initial concept I'm messing around with for the look and feel override stuff. It's not a pattern I would advocate for so much now I've dug into it but it's something my current shop wants me to investigate to allow us to get up and running fast. 

```
function ReactBasedSlider(props) {
    if (props.hasOwnProperty('drawLinearSlider')) {
        let userDrawLinearSlider = props.drawLinearSlider;
        let wrappedDraw = function(ctx, width, height, sliderPos, sliderStyle) {
            bindCanvasContextProperties(ctx);
            return userDrawLinearSlider(ctx, width, height, sliderPos, sliderStyle);
        };
        return React.createElement('ReactBasedSlider', Object.assign({}, props, {
            drawLinearSlider: wrappedDraw,
        }), props.children);
    }

    return React.createElement('ReactBasedSlider', props, props.children);
}

export default ReactBasedSlider;
```

packages/juce-blueprint/esm/Blueprint.js
JoshMarler: Ahhhhhaaaa. Ha of course. I was confused as to why the CanvasView's onMeasure prop wasn't working. Let me try fixing that up now.

blueprint/core/blueprint_CanvasView.h
JoshMarler: Heh yeah, wasn't exactly preserving state!

packages/juce-blueprint/esm/Blueprint.js
JoshMarler: Hmm. This doesn't appear to be working yet when trying to bind the outer component onMeasure. I'm missing how to make this work. 

```
     let onMeasure = this._onMeasure;

      canvas = React.createElement('CanvasView', Object.assign({}, props, {
        onDraw: wrappedDraw,
        onMeasure: onMeasure
      }), props.children);
```

nick-thompson: Hmm, I think you'll basically want this?

```js
class CanvasView extends React.Component {
  constructor(props) {
    super(props);

    this._onMeasure = this._onMeasure.bind(this);
    this._onDraw = this._onDraw.bind(this);

    this.state = {
      width: 0,
      height: 0,
    };
  }

  _onMeasure(w, h) {
    this.setState({ width: w, height: h });
    if (typeof this.props.onMeasure === 'function')
      this.props.onMeasure(w, h);
  }

  _onDraw(ctx) {
    if (typeof this.props.onDraw === 'function') {
      bindCtx(ctx);
      if (this.props.autoclear)
        ctx.clearRect(0, 0, this.state.width, this.state.height);
      this.props.onDraw(ctx);
    }
  }
  render() {
    return React.createElement('CanvasView', Object.assign({}, this.props, {
      onDraw: this._onDraw,
      onMeasure: this._onMeasure,
    }, this.props.children);
  }
}
```

JoshMarler: @nick-thompson, maybe ignore me ... think I've sorted

JoshMarler: @nick-thompson, ha yeah OK I fixed but yours is nicer than mine. I need a little js refresh session at some point I think ...

JoshMarler: Curses no that isn't working. Not getting any repaints now. Will investigate.

nick-thompson: 😕 confusing. Yea, I'd have to dig into it... that _should_ work.. I think

JoshMarler: Yeah no worries. I'll sort it, something daft I expect

JoshMarler: Works with this but I'll try to make it fit your structure more as it's nicer:

```
export class Canvas extends Component {
  constructor(props) {
    super(props);
    this._onMeasure = this._onMeasure.bind(this);

    this.state = {
      width: 0,
      height: 0
    };
  }

  _onMeasure(width, height) {
    console.log('canvas on measure');
    this.setState({
      width: width,
      height: height
    });
  }

  render() {
    let props = this.props;
    let userOnDraw = props.onDraw;

    const {width, height} = this.state

    let canvas = React.createElement('CanvasView', props, props.children);

    if (userOnDraw) {
      let wrappedDraw = (ctx) => {
        bindCanvasContextProperties(ctx);
        if (props.autoclear) ctx.clearRect(0, 0, width, height);
        return userOnDraw(ctx);
      }

      canvas = React.createElement('CanvasView', Object.assign({}, props, {
        onDraw: wrappedDraw,
        onMeasure: this._onMeasure
      }), props.children);
    }

    return canvas;
  }
}
```

nick-thompson: That's super odd... they feel basically identical. I'm cool with how you have it there though honestly! The only thing I would add is in the _onMeasure hook, check if the user tried to supply their own `onMeasure` and if so, call it there

JoshMarler: Right ... I'm determined to workout why it isn't working

JoshMarler: Pfft. Stumped. Possible it's anything to do with my usage of the Canvas component?

```
          <View
              {...this.props}
              onMeasure={this._onMeasure}
              onMouseDown={this._onMouseDown}
              onMouseDrag={this._onMouseDrag}>
            <Canvas autoclear={true} {...styles.canvas} onDraw={this._drawSlider} />
            {this.props.children}
          </View>
```

nick-thompson: I doubt it. You're not getting any errors right? The paint's just not happening?

JoshMarler: Nope nothing. This works fine:

```
export class Canvas extends Component {
  constructor(props) {
    super(props);
    this._onMeasure = this._onMeasure.bind(this);

    this.state = {
      width: 0,
      height: 0
    };
  }

  _onMeasure(width, height) {
    this.setState({
      width: width,
      height: height
    });
  }

  render() {
    let props = this.props;
    const {width, height} = this.state

    let wrappedDraw = (ctx) => {
      bindCanvasContextProperties(ctx);
      if (props.autoclear) ctx.clearRect(0, 0, width, height);
      if (typeof props.onDraw === 'function') props.onDraw(ctx);
    }

    let wrappedOnMeasure = (width, height) => {
      this._onMeasure(width, height);
      if (typeof props.onMeasure === 'function') props.onMeasure(width, height);
    }

    return React.createElement('CanvasView', Object.assign({}, props, {
      onDraw: wrappedDraw,
      onMeasure: wrappedOnMeasure
    }), props.children);
  }
}
```

How strange.

JoshMarler: @nick-thompson, I'm going to call it quits for the day so I've pushed up the only working version I can get in https://github.com/nick-thompson/blueprint/pull/56/commits/1e351ae97ae5b4785f035d5d836982c1b4b32c60 incase you want to take a look and see if you can workout why we can't get your implementation to work. I don't understand it myself.

nick-thompson: Took a pass this morning, I've got it working on my end! Only needed a few small changes: https://gist.github.com/nick-thompson/d28a7fcd334fe0c2eb47ff4636649da6 (the unique_ptr stuff is just because I'm on an older version of juce, ignore those)

You should be able to apply that diff to your branch locally and give it a run. Let me know if that works for you

JoshMarler: @nick-thompson, heh. Good catch with std::max!

Strangely still the same issue for me so must be something my end. Were you testing this in an example project?

JoshMarler: Nice catch with std::max. 

So I must be missing something but I can only make your gist work if I wrap this._onDraw and this._onMeasure. Some JS issue I'm ignorant of here. I might test on mac with my project just in case this is some total weird platform specific thing. 

This works for me:

```
export class Canvas extends Component {
  constructor(props) {
    super(props);

    this._onMeasure = this._onMeasure.bind(this);
    this._onDraw    = this._onDraw.bind(this);

    this.state = {
      width: 0,
      height: 0
    };
  }

  _onMeasure(width, height) {
    this.setState({
      width: width,
      height: height
    });

    if (typeof this.props.onMeasure === 'function')
      this.props.onMeasure(width, height);
  }

  _onDraw(ctx) {
    if (typeof this.props.onDraw === 'function') {
      bindCanvasContextProperties(ctx);

      if (this.props.autoclear) {
        ctx.clearRect(0, 0, this.state.width, this.state.height);
      }

      this.props.onDraw(ctx);
    }
  }

  render() {
    let wrappedDraw = (ctx) => {
      this._onDraw(ctx);
    }

    let wrappedOnMeasure = (width, height) => {
      this._onMeasure(width, height);
    }

    return React.createElement('CanvasView', Object.assign({}, this.props, {
      onDraw: wrappedDraw,
      onMeasure: wrappedOnMeasure
    }), this.props.children);
  }
}
```

JoshMarler: Does seem to be an issue binding this in Object.assign() as the below also works. Not clear on why I'm the only one seeing this though. 

```
  render() {
    return React.createElement('CanvasView', Object.assign({}, this.props, {
      onDraw: (ctx) => { this._onDraw(ctx) },
      onMeasure: (width, height) => { this._onMeasure(width, height) }
    }), this.props.children);
  }

```

nick-thompson: Wow, no kidding. Ok, thanks for the update. This sounds like a bug in duktape to be honest with you. Wondering if we'll fix the issue just by updating our duktape installation...

Let's use the code from your comment:

```
  render() {
    return React.createElement('CanvasView', Object.assign({}, this.props, {
      onDraw: (ctx) => { this._onDraw(ctx) },
      onMeasure: (width, height) => { this._onMeasure(width, height) }
    }), this.props.children);
  }
```
And can you also add a todo comment?
```
// TODO: Check this for bug in duktape
```

That at least will unblock this PR from going in, and we can revisit later if we find anything in duktape. Sound good?

JoshMarler: Yeah that's all I can boil it down to, I'm testing this on Ubuntu so maybe a duktape on linux quirk. Will add the TODO. Thanks Nick!

JoshMarler: @nick-thompson, have rebased and pushed up changes with you advised comment in https://github.com/nick-thompson/blueprint/commit/9540539efc1763be4bf18024768abeae0b09606c. Once this is in I will fix up https://github.com/nick-thompson/blueprint/pull/58 and then I think we're at a good point to start adding some Canvas/drawing docs and example apps etc.

 I think I'll raise an issue for the Canvas docs/examples which we can use to track to work and discuss ideas. Sound good?

nick-thompson: Awesome, sounds great! I'll take one last look here now and merge if all's good

Issue #14, opened 2019-06-26
Open question: Image/BinaryData Support


nick-thompson:
Handling image data is an interesting challenge, and there are many different ways to do it.

Currently, Blueprint implements an `<Image>` element that accepts a `source` property expecting a value that is an SVG string, which on the backend is parsed by a `juce::Drawable` and drawn at paint time. This works nicely with Webpack's SVG file loader because it automatically reads `require('path/to/svg.svg')` statements and satisfies them by embedding the raw svg string from the file.

However, this maybe won't scale for larger, more complicated SVGs, and it totally ignores the *other* common image use case, which is big PNG files either read from disk or carried in BinaryData. One possible path forward is to support BinaryData directly with some kind of C++ interface that lets you register a key/value pair which is a name and a pointer to some specific binary data, and allow that pair to be referenced by the key on the javascript side. `<BinaryDataImage sourceKey={'background'} />`. I'm not convinced this is a great idea though.

On the flip side, this might be a good case for a problem that Blueprint intentionally does not solve, and leaves open to the user. For example, if you have a big PNG strip for drawing a knob and you use the juce ImageCache to speed that up, maybe the best approach for interfacing with that is to write a custom KnobView and register that view with the ReactApplicationRoot (#8). Then there's no reason for Blueprint to concern itself with binary data, and you can still take full advantage of the techniques that JUCE already has for interacting with it.

JoshMarler:
@nick-thompson, You game for me to have a crack at this? Following a URI pattern seems like it shouldn't be too hard. If it's a file path like URI load via file system, if it's URL follow a different model. Can start with file path and go from there. 

Binary assets seem like they should be reasonably straight forward too. Maybe go for local URI's first and then support remote file loading as second step?

nick-thompson:
Definitely! Yea, let's go for local first and defer remote. So first step, then, is supporting `data:`, `file:`, and a custom `bindat:`? Or maybe just start with `data:` and `file:`.

`data:` should roughly match the Data URL spec https://developer.mozilla.org/en-US/docs/Web/HTTP/Basics_of_HTTP/Data_URIs, and will have to edit the GainPlugin example to adapt the way it currently sends svg strings (I think can just prepend the data: with the mime type)
`file:<filepath>` maybe we only support absolute file paths for now? What would a relative file path be relative to? Wondering about things like iOS bundles.... maybe a "todo later" on that
`bindat:<name>` would be just a name-based lookup on BinaryData.

[...]

PR #57, opened 2020-06-01
Add initial support for canvas API drawImage() function.

JoshMarler:
This implementation of the CanvasRenderingContext2D.drawImage()
  function currently just supports svg strings which can be loaded via
  require() and the webpack svg-inline-loader.

@nick-thompson, 

To get this to work properly for my specific use case I actually needed to interrogate the SVG string to get the image/viewBox width and height. A function like this (slightly less hacky) may or may not be of use in Blueprint.

```
function getSvgDimensions(svg) {
    let parseString = require('xml2js').parseString;
    let svgDimensions = {};

    parseString(svg, (err, result) => {
        if (result) {
            const viewBox = result.svg.$.viewBox;
            if (viewBox) {
                const dimensions = viewBox.split(' ');
                const width  = parseInt(dimensions[2]);
                const height = parseInt(dimensions[3]);

                svgDimensions = { width, height };
            }
        }
        else {
            console.log('Bad SVG string passed: ' + err);
        }
    });

    return svgDimensions;
}
```

### DIFF
--- a/blueprint/core/blueprint_CanvasView.h
+++ b/blueprint/core/blueprint_CanvasView.h
@@ -394,6 +394,45 @@ namespace blueprint
                         return juce::var();
                     }
             });
+
+            // drawImage support
+            setProperty("drawImage", juce::var::NativeFunction {
+                    [=](const juce::var::NativeFunctionArgs& args) -> juce::var {
+                        jassert(graphics);
+                        jassert(args.numArguments >= 3 && args.numArguments <= 5);
+
+                        const juce::String svg  = args.arguments[0].toString();
+                        const float        xPos = args.arguments[1];
+                        const float        yPos = args.arguments[2];
+
+                        //TODO: Add support for drawimage source width and source height to draw sub rect of an image.
+                        //      ctx.drawImage(image, sx, sy, sWidth, sHeight, dx, dy, dWidth, dHeight);
+
+                        std::unique_ptr<juce::XmlElement> svgElement  = juce::XmlDocument::parse(svg);
+                        jassert(svgElement);
+
+                        if (svgElement)
+                        {
+                            std::unique_ptr<juce::Drawable> svgDrawable = juce::Drawable::createFromSVG(*svgElement);
+
+                            if (args.numArguments == 5)
+                            {
+                                const float destWidth  = args.arguments[3];
+                                const float destHeight = args.arguments[4];
+                                const auto  bounds     = juce::Rectangle<float>(xPos, yPos, destWidth, destHeight);
+
+                                svgDrawable->setTransformToFit(bounds, juce::RectanglePlacement::stretchToFit);
+                                svgDrawable->draw(*graphics, 1.0f);
+                            }
+                            else
+                            {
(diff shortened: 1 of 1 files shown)


blueprint/core/blueprint_CanvasView.h
```diff
@@ -394,6 +394,45 @@ namespace blueprint
...
+
+                        const juce::String svg  = args.arguments[0].toString();
+                        const float        xPos = args.arguments[1];
+                        const float        yPos = args.arguments[2];
+
+                        //TODO: Add support for drawimage source width and source height to draw sub rect of an image.
+                        //      ctx.drawImage(image, sx, sy, sWidth, sHeight, dx, dy, dWidth, dHeight);
+
+                        std::unique_ptr<juce::XmlElement> svgElement  = juce::XmlDocument::parse(svg);
+                        jassert(svgElement);
```
nick-thompson, line 412: Here I think we can be slightly more forgiving– we have the `if (svgElement)` below and the `jassert` here. I think we can just nudge that to:

```
if (!svgElement) {
    DBG("WARNING: Invalid SVG string supplied to `drawImage`.");
    return juce::var();
}
```

I think favoring warnings + noops when we can just helps out the hot reload stuff. It'd be marginally annoying if I had to recompile just because I put a typo in my javascript that calls `drawImage` :)

JoshMarler:
> Looks great, just have one kind of nitpick in-line before we merge please!
> 
> Also this PR feels related to #14. I'd love to hear your thoughts on that... For example we could treat all images as Data URIs and load them as strings into the js bundle, but then we're passing around and parsing huge strings to draw images. Maybe instead we could have an `appRoot.registerNativeImage` method where you pass some BinaryData or something and we can inject this javascript object into Duktape that carries pointers to that data in its capture group. It's probably a long discussion, not meant for here, but if you're interested, would love your discussion in #14!

Ooo yeah that's an interesting approach. I haven't used the juce BinaryData asset stuff via the Projucer for a long time. My first knee jerk reaction without looking into this all properly is if we're treating images as URIs can juce play the "server" role and just load an asset based on an identifier. I've literally not thought about it properly at all though. My current project is a completely "vector based" UI without any png assets and the like currently, so all just svg file and juce::Path etc. I've not needed to worry too much about this yet but it is almost certainly a problem I'm going to hit very soon. I will have a good think about this one!

blueprint/core/blueprint_CanvasView.h
JoshMarler: Yeah cool makes sense! Will get that sorted now.

JoshMarler: @nick-thompson, tidied up in https://github.com/nick-thompson/blueprint/pull/57/commits/d52953cedb899331b8059ee178e6c85493f89221

nick-thompson:
Great, thanks! Good to merge now? Or should we wait for #56 as well?

JoshMarler:
> Great, thanks! Good to merge now? Or should we wait for #56 as well?

I reckon it should be good to go in as doesn't make things anymore broken than they already are heh.

PR #59, opened 2020-06-04
Fixup a hot reload error in ReactApplicationRoot.

JoshMarler:
Ensure we call then juce::File overload of evaluate when bundle
  has changed and avoid duplicate watchers.

@nick-thompson, realised there is an error in this bit of hot reload code. Could have used an unordered_map or something to avoid duplicates but I think this is nicer to be honest as the timerCallback is going to be called more often and iterating over a map with pairs is uglier imho.

### DIFF
--- a/blueprint/core/blueprint_ReactApplicationRoot.h
+++ b/blueprint/core/blueprint_ReactApplicationRoot.h
@@ -52,6 +52,16 @@ namespace blueprint
             watchedBundles.push_back({ bundle, bundle.getLastModificationTime() });
         }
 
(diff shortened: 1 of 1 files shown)


nick-thompson:
Ah, nice catch. So you were getting multiple "hot reloads" after a single bundle file change?

LGTM, I'll go ahead and merge. Thanks!

JoshMarler:
Yeah exactly. I hadn't hit an issue before whilst testing but I started the whole "juce wrapper" component idea where I use the CanvasContext in look and feel overrides and it reared its head there.

Issue #22, opened 2019-09-09
Add continuous integration to project


tobanteAudio:
During my first tests with `blueprint`, `npm` had a couple of errors which I could solve by deleting the `package-lock.json` files & running the `npm install` & `npm run build` commands with `yarn`:

```sh
yarn
yarn build
```

The problem could have probably been solved otherwise, but I thought this is a perfect example were CI could help and catch these kinds of errors early.

The project maintainer has to enable the services, but I can create the `appveyor.yml` (Windows) and `.travis.yml` (macOS & Linux).

This would also help with automatic documentation generation. See Issue #12 

Toby

tobanteAudio:
I added Travis-ci support on my fork of `blueprint`: [tobanteAudio/blueprint](https://github.com/tobanteAudio/blueprint)

Trying to build it with the `xcode10.1` image I get the same error as explained above (in Windows). Here is the [output](https://travis-ci.org/tobanteAudio/blueprint/builds/582612104) of the build.

tobanteAudio:
The problem seems to be the `npm install` command. It should be replaced by `npm ci`. 
See [here for an explanation](https://medium.com/better-programming/npm-ci-vs-npm-install-which-should-you-use-in-your-node-js-projects-51e07cb71e26). 

Although I'm not exactly sure why it works with one and not the other.

Maybe this should be in the README?

nick-thompson:
Awesome! I love it, this looks great. A quick couple thoughts:

* Interesting about `npm install` vs `npm ci`. It makes sense as to why `yarn` works but `npm install` doesn't... I feel like `npm ci` should be npm's default (which would match `yarn`'s behavior). Let's use `npm ci` in the CI builds and also yea update the README. I suppose users who just want to use Blueprint should use `npm ci` but people who want to contribute and might consider updating the dependencies would want `npm install`
* Love that you have `pluginval` in there, that's a great idea
* I'm a little unsure about the VST3 build– did you have to add the SDK or is JUCE ok without it? I had originally unchecked that build just thinking that it would complicate the examples process for anybody who doesn't have the VST3 SDK locally. But if JUCE is cool without that, then let's keep it

tobanteAudio:
The VST3 SDK is built into JUCE, so no extra dependencies.

nick-thompson:
Ok perfect! I didn't know that.

[...]

PR #61, opened 2020-06-11
Initial spike for duktape debugger support. Currently a work in

JoshMarler:
progress.

@nick-thompson, this is still in progress but thought I'd open up the PR so we can begin to discuss. I'll follow up with a comment on the PR explaining the current setup/workflow with the vscode extension using source maps. 

Without source-maps it should still be possible to debug the main.js/bundle file in the duktape browser debugger I believe. Would need to switch off minification in webpack when building bundles to make this usable. I will test this shortly. 

I've only been testing on Linux in the context of my app thus far so I'm about the pull this branch down on the mac and test with the example GainPlugin. 

I need to make a PR to the vscode extension as I have had to fix/hack one thing to get this all working. I'll fork the extension and drop you a link over so you can play with it. 

Note that I've had to edit duk_config.h as the debug defines need to be toggled at that level rather than in blueprint.cpp. This may prompt discussions about the build system and custom duktape headers but perhaps we can get away with this change to the default config for now?

### DIFF
--- a/blueprint/core/blueprint_EcmascriptEngine.cpp
+++ b/blueprint/core/blueprint_EcmascriptEngine.cpp
@@ -74,6 +74,10 @@ namespace blueprint
             throw std::runtime_error(msg);
         }
 
+        //TODO: Keep safeEvalString and safeCompileFile or make EcmascripEngine member funcs
+        //      now that we have out errorHandle wrapper in EcmascriptEngine? Passing
+        //      callback here feels a bit ambiguous in regards to popping the error
+        //      off of the stack after invoking the error callback. Whose responsibility is it etc.
         template <typename T>
         static bool safeEvalString(duk_context* ctx, const juce::String& s, T&& callback)
         {
@@ -81,13 +85,8 @@ namespace blueprint
             {
                 if (callback != nullptr)
                 {
-                    const juce::String trace = duk_safe_to_stacktrace(ctx, -1);
-                    const juce::String msg = duk_safe_to_string(ctx, -1);
-
-                    // Call the user provided error handler
-                    std::invoke(callback, msg, trace);
-
-                    // Indicate failure
+                    // Call the user provided error handler and indicate failure
+                    std::invoke(callback);
                     return false;
                 }
 
@@ -108,13 +107,8 @@ namespace blueprint
             {
                 if (callback != nullptr)
                 {
-                    const juce::String trace = duk_safe_to_stacktrace(ctx, -1);
-                    const juce::String msg = duk_safe_to_string(ctx, -1);
-
-                    // Call the user provided error handler.js
-                    std::invoke(callback, msg, trace);
-
-                    // Indicate failure
+                    // Call the user provided error handler.js and indicate failure
+                    std::invoke(callback);
                     return false;
                 }
 
@@ -123,6 +117,13 @@ namespace blueprint
 
             // Indicate success
             return true;
+       }
+
+       static void dumpContext(duk_context* ctx)
+       {
+           duk_push_context_dump(ctx);
+           DBG(duk_to_string(ctx, -1));
+           duk_pop(ctx);
        }
     }
 
@@ -134,6 +135,30 @@ namespace blueprint
 
         // Add console.log support
         duk_console_init(ctx, DUK_CONSOLE_FLUSH);
+
+        // Create our error handler wrapping a user supplied error callback.
+        errorHandler = [=]
+        {
+            if (onUncaughtError == nullptr)
+                duk_throw_raw(ctx);
+
+            const juce::String trace = duk_safe_to_stacktrace(ctx, -1);
+            const juce::String msg = duk_safe_to_string(ctx, -1);
+
+            // Call the user provided error handler
+            std::invoke(onUncaughtError, msg, trace);
+
+            // Clear out the stack so we can re-register native functions
+            // after we clear out the lambda release pool etc.
+            while (duk_get_top(ctx))
+            {
+                duk_remove(ctx, duk_get_top_index(ctx));
+            }
+
+            // Clear the LambdaHelper release pool as duktape does not call object
+            // finalizers in the event of an evaluation error or duk_pcall failure.
+            lambdaReleasePool.clear();
+        };
     }
 
     EcmascriptEngine::~EcmascriptEngine()
@@ -144,37 +169,32 @@ namespace blueprint
     //==============================================================================
     juce::var EcmascriptEngine::evaluate (const juce::String& code)
     {
-        if (!detail::safeEvalString(ctx, code, onUncaughtError))
-            return juce::var::undefined();
+        jassert(code.isNotEmpty());
+
+        if (!detail::safeEvalString(ctx, code, errorHandler))
+            return juce::var(EvaluationError);
 
         auto result = readVarFromDukStack(ctx, -1);
-
         duk_pop(ctx);
+
         return result;
     }
 
     juce::var EcmascriptEngine::evaluate (const juce::File& code)
     {
+        jassert(code.existsAsFile());
+        jassert(code.loadFileAsString().isNotEmpty());
+
         juce::var result = juce::var::undefined();
 
-        if (detail::safeCompileFile(ctx, code, onUncaughtError))
+        if (!detail::safeCompileFile(ctx, code, errorHandler))
+            return juce::var(EvaluationError);
+
+        // Call compiled function
+        if (duk_pcall(ctx, 0) != DUK_EXEC_SUCCESS)
         {
-            // Call compiled function
-            if (duk_pcall(ctx, 0) != DUK_EXEC_SUCCESS)
-            {
-                if (onUncaughtError != nullptr)
-                {
-                    const juce::String trace = duk_safe_to_stacktrace(ctx, -1);
-                    const juce::String msg = duk_safe_to_string(ctx, -1);
-
-                    // Call the user provided error handler
-                    std::invoke(onUncaughtError, msg, trace);
-
-                    return juce::var::undefined();
-                }
-
-                duk_throw_raw(ctx);
-            }
+            errorHandler();
+            return juce::var(EvaluationError);
         }
 
         // Collect the return value
@@ -208,7 +228,7 @@ namespace blueprint
     void EcmascriptEngine::registerNativeMethod (const juce::String& target, const juce::String& name, NativeFunction fn, void* stash)
     {
         // Evaluate the target string on the context, leaving the result on the stack
-        if (!detail::safeEvalString(ctx, target, onUncaughtError))
+        if (!detail::safeEvalString(ctx, target, errorHandler))
             return;
 
         // We wrap the native function to provide a helper layer storing and retrieving the
@@ -238,7 +258,8 @@ namespace blueprint
     void EcmascriptEngine::registerNativeProperty (const juce::String& target, const juce::String& name, const juce::var& value)
     {
         // Evaluate the target string on the context, leaving the result on the stack
-        if (!detail::safeEvalString(ctx, target, onUncaughtError))
+        // TODO: Return specific error code if target not in stack?
+        if (!detail::safeEvalString(ctx, target, errorHandler))
             return;
 
         // Then assign the property
@@ -250,13 +271,15 @@ namespace blueprint
     juce::var EcmascriptEngine::invoke (const juce::String& name, const std::vector<juce::var>& vargs)
     {
         // Evaluate the target string on the context, leaving the result on the stack
-        if (!detail::safeEvalString(ctx, name, onUncaughtError))
+        // TODO: Return specific error code if name not in stack?
+        if (!detail::safeEvalString(ctx, name, errorHandler))
             return juce::var::undefined();
 
+        // Ensure requested function exists on the stack
         duk_require_function(ctx, -1);
 
         // Push the args to the duktape stack
-        auto nargs = static_cast<duk_idx_t>(vargs.size());
+        const auto nargs = static_cast<duk_idx_t>(vargs.size());
         duk_require_stack_top(ctx, nargs);
 
         for (auto& p : vargs)
@@ -265,24 +288,16 @@ namespace blueprint
         // Invocation
         if (duk_pcall(ctx, nargs) != DUK_EXEC_SUCCESS)
         {
-            if (onUncaughtError != nullptr)
-            {
-                const juce::String trace = duk_safe_to_stacktrace(ctx, -1);
-                const juce::String msg = duk_safe_to_string(ctx, -1);
+            errorHandler();
 
-                // Call the user provided error handler
-                std::invoke(onUncaughtError, msg, trace);
-
-                return juce::var::undefined();
-            }
-
-            duk_throw_raw(ctx);
+            // TODO: Return specific error code if invoke fails?
+            return juce::var::undefined();
         }
 
         // Collect the return value
         auto result = readVarFromDukStack(ctx, -1);
-
         duk_pop(ctx);
+
         return result;
     }
 
@@ -522,6 +537,7 @@ namespace blueprint
                             // the global stash and invoke it with the provided args.
                             duk_push_global_stash(ctx);
                             duk_get_prop_string(ctx, -1, funId.toRawUTF8());
+
                             duk_require_function(ctx, -1);
 
                             // Push the args to the duktape stack
--- a/blueprint/core/blueprint_EcmascriptEngine.cpp
+++ b/blueprint/core/blueprint_EcmascriptEngine.cpp
@@ -91,12 +91,39 @@ namespace blueprint
                     return false;
                 }
 
-                 duk_throw_raw(ctx);
+                duk_throw_raw(ctx);
             }
 
             // Indicate success
             return true;
         }
+
+        template<typename T>
+        static bool safeCompileFile(duk_context* ctx, const juce::File& file, T&& callback)
+        {
+            // Push the js filename to be compiled/evaluated
+            duk_push_string(ctx, file.getFileName().toRawUTF8());
+
+            if (duk_pcompile_string_filename(ctx, DUK_COMPILE_EVAL, file.loadFileAsString().toRawUTF8()) != 0)
+            {
+                if (callback != nullptr)
+                {
+                    const juce::String trace = duk_safe_to_stacktrace(ctx, -1);
+                    const juce::String msg = duk_safe_to_string(ctx, -1);
+
+                    // Call the user provided error handler.js
+                    std::invoke(callback, msg, trace);
+
+                    // Indicate failure
+                    return false;
+                }
+
+                duk_throw_raw(ctx);
+            }
+
+            // Indicate success
+            return true;
+       }
     }
 
     //==============================================================================
@@ -123,6 +150,37 @@ namespace blueprint
         auto result = readVarFromDukStack(ctx, -1);
 
         duk_pop(ctx);
+        return result;
+    }
+
+    juce::var EcmascriptEngine::evaluate (const juce::File& code)
+    {
+        juce::var result = juce::var::undefined();
+
+        if (detail::safeCompileFile(ctx, code, onUncaughtError))
+        {
+            // Call compiled function
+            if (duk_pcall(ctx, 0) != DUK_EXEC_SUCCESS)
+            {
+                if (onUncaughtError != nullptr)
+                {
+                    const juce::String trace = duk_safe_to_stacktrace(ctx, -1);
+                    const juce::String msg = duk_safe_to_string(ctx, -1);
+
+                    // Call the user provided error handler
+                    std::invoke(onUncaughtError, msg, trace);
+
+                    return juce::var::undefined();
+                }
+
+                duk_throw_raw(ctx);
+            }
+        }
+
+        // Collect the return value
+        result = readVarFromDukStack(ctx, -1);
+        duk_pop(ctx);
+
         return result;
     }
 
@@ -226,6 +284,28 @@ namespace blueprint
 
         duk_pop(ctx);
         return result;
+    }
+
+    //==============================================================================
+    void EcmascriptEngine::debuggerAttach()
+    {
+        duk_trans_socket_init();
+        duk_trans_socket_waitconn();
+
+        duk_debugger_attach(ctx,
+                            duk_trans_socket_read_cb,
+                            duk_trans_socket_write_cb,
+                            duk_trans_socket_peek_cb,
+                            duk_trans_socket_read_flush_cb,
+                            duk_trans_socket_write_flush_cb,
+                            NULL,
+                            [](duk_context*, void*) { duk_trans_socket_finish(); },
+                            NULL);
+    }
+
+    void EcmascriptEngine::debuggerDetach()
+    {
+        duk_debugger_detach(ctx);
     }
 
     //==============================================================================
--- a/blueprint/core/blueprint_EcmascriptEngine.cpp
+++ b/blueprint/core/blueprint_EcmascriptEngine.cpp
@@ -159,6 +159,7 @@ namespace blueprint
             // finalizers in the event of an evaluation error or duk_pcall failure.
             lambdaReleasePool.clear();
         };
+
     }
 
     EcmascriptEngine::~EcmascriptEngine()
@@ -308,19 +309,33 @@ namespace blueprint
         duk_trans_socket_waitconn();
 
         duk_debugger_attach(ctx,
-                            duk_trans_socket_read_cb,
-                            duk_trans_socket_write_cb,
-                            duk_trans_socket_peek_cb,
-                            duk_trans_socket_read_flush_cb,
-                            duk_trans_socket_write_flush_cb,
-                            NULL,
-                            [](duk_context*, void*) { duk_trans_socket_finish(); },
-                            NULL);
+            duk_trans_socket_read_cb,
+            duk_trans_socket_write_cb,
+            duk_trans_socket_peek_cb,
+            duk_trans_socket_read_flush_cb,
+            duk_trans_socket_write_flush_cb,
+            NULL,
+            [](duk_context* ctx, void* data)
+            {
+                duk_trans_socket_finish();
+
+                auto engine = static_cast<EcmascriptEngine*>(data);
+                engine->stopTimer();
+            },
+            this);
+
+        // Start timer for duk_debugger_cooperate calls
+        startTimer(200);
     }
 
     void EcmascriptEngine::debuggerDetach()
     {
         duk_debugger_detach(ctx);
+    }
+
+    void EcmascriptEngine::timerCallback()
+    {
+        duk_debugger_cooperate(ctx);
     }
 
     //==============================================================================
--- a/blueprint/core/blueprint_EcmascriptEngine.cpp
+++ b/blueprint/core/blueprint_EcmascriptEngine.cpp
@@ -552,7 +552,6 @@ namespace blueprint
                             // the global stash and invoke it with the provided args.
                             duk_push_global_stash(ctx);
                             duk_get_prop_string(ctx, -1, funId.toRawUTF8());
-
                             duk_require_function(ctx, -1);
 
                             // Push the args to the duktape stack
--- a/blueprint/core/blueprint_ReactApplicationRoot.h
+++ b/blueprint/core/blueprint_ReactApplicationRoot.h
@@ -180,43 +180,6 @@ namespace blueprint
 
         //==============================================================================
         /**
-         * Evaluates a javascript bundle/code in the Ecmascript engine.
-         *
-         * It is possible to pass a js bundle file to this evaluate() overload using juce::File::loadAsString. However,
-         * if you require hot-reload functionality you should use the juce::File based evaluate() overload.
-         * This overload can also be used to evaluate js code inline/directly.
-         **/
-        juce::var evaluate (const juce::String& bundle)
-        {
-            JUCE_ASSERT_MESSAGE_THREAD
-
-            // Clear error state from previous js evals
-            errorText.reset();
-
-            if (!bundleValid)
-            {
-                // Install React.js backend rendering methods
-                registerNativeRenderingHooks();
-
-                if (beforeBundleEval)
-                    beforeBundleEval(std::nullopt);
-            }
-
-            // Set bundleValid prior to evaluate. This ensures dispatchEvent/dispatchViewEvent
-            // can be called during EcmascriptEngine::evaluate. This is required to allow things
-            // like triggering View "Measure" callbacks when laying out the initial component tree.
-            bundleValid = true;
-            auto result = engine.evaluate(bundle);
-
-            bundleValid = result != EcmascriptEngine::EvaluationError;
-
-            if (bundleValid && afterBundleEval)
-                afterBundleEval(std::nullopt);
-
-            return result;
-        }
-
-        /**
          * Evaluates a javascript bundle file in the Ecmascript engine.
          *
          *  If hot-reload functionality has been enabled via enableHotReload(), ReactApplicationRoot will watch the
@@ -328,7 +291,7 @@ namespace blueprint
         }
 
         //==============================================================================
-        using BundleEvalCallback = std::function<void(std::optional<juce::File> bundle)>;
+        using BundleEvalCallback = std::function<void(const juce::File& bundle)>;
 
         /**
          * Called before a bundle is loaded/evaluated.
@@ -339,8 +302,7 @@ namespace blueprint
          * Use this callback to register any native methods/properties to be used from within the JS bundle. You can
          * also use this callback to register any custom error handlers via ReactApplicationRoot::onUncaughtError.
          *
-         * @param bundle The reloaded bundle file if applicable. If evaluate() was called with a raw/inline javascript
-         * code string rather than a .js file bundle will be a std::nullopt_t.
+         * @param bundle The reloaded bundle file.
          *
          * For most applications the bundle will simply be the main app root bundle file. However, other use cases
          * for evaluating multiple bundle files may exist, for example to facilitate things like polyfills etc.
@@ -356,29 +318,26 @@ namespace blueprint
          *     // Not strictly required if JUCE_DEBUG set.
          *     appRoot.enableHotReload(true);
          *
-         *     appRoot.beforeBundleEval = [=](std::optional<juce::File> bundle)
+         *     appRoot.beforeBundleEval = [=](juce::File bundle)
          *     {
-         *         if (bundle)
+         *         if (bundle.getFullPathName() == myAppBundle.getFullPathName())
          *         {
-         *             if (bundle->getFullPathName() == myAppBundle.getFullPathName())
-         *             {
-         *                  appRoot.registerNativeMethod(
-         *                      "myNativeMethod",
-         *                      [](void* stash, const juce::var::NativeFunctionArgs& args) {
-         *                          auto* self = reinterpret_cast<MyEditor*>(stash);
+         *              appRoot.registerNativeMethod(
+         *                  "myNativeMethod",
+         *                  [](void* stash, const juce::var::NativeFunctionArgs& args) {
+         *                      auto* self = reinterpret_cast<MyEditor*>(stash);
          *
-         *                          const juce::String& someParam = args.arguments[0].toString();
-         *                          self->myNativeMethod(someParam);
+         *                      const juce::String& someParam = args.arguments[0].toString();
+         *                      self->myNativeMethod(someParam);
          *
-         *                          return juce::var::undefined();
-         *                      },
-         *                      (void*) this
-         *                  );
-         *             }
-         *             else
-         *             {
-         *                 // You have loaded some other js bundle. i.e. a polyfill
-         *             }
+         *                      return juce::var::undefined();
+         *                  },
+         *                  (void*) this
+         *              );
+         *         }
+         *         else
+         *         {
+         *             // You have loaded some other js bundle. i.e. a polyfill
          *         }
          *     };
          *
@@ -397,8 +356,7 @@ namespace blueprint
          *
          * Use this callback to dispatch any initial state/events required by the React/JS application on load/reload.
          *
-         * @param bundle The reloaded bundle file if applicable. If evaluate() was called with a raw/inline javascript
-         * code string rather than a .js file bundle will be a std::nullopt_t.
+         * @param bundle The reloaded bundle file.
          *
          * For most applications the bundle will simply be the main app root bundle file. However, other use cases
          * for evaluating multiple bundle files may exist, for example to facilitate things like polyfills etc.
@@ -414,19 +372,16 @@ namespace blueprint
          *     // Not strictly required if JUCE_DEBUG set.
          *     appRoot.enableHotReload(true);
          *
-         *     appRoot.afterBundleEval = [=](std::optional<juce::File> bundle)
+         *     appRoot.afterBundleEval = [=](const juce::File& bundle)
          *     {
-         *         if (bundle)
+         *         if (bundle.getFullPathName() == myAppBundle.getFullPathName())
          *         {
-         *             if (bundle->getFullPathName() == myAppBundle.getFullPathName())
-         *             {
-         *                 juce::String message("This is an important message");
-         *                 appRoot.dispatchEvent("importantMessage", message);
-         *             }
-         *             else
-         *             {
-         *                 // You have loaded some other js bundle. i.e. a polyfill
-         *             }
+         *             juce::String message("This is an important message");
+         *             appRoot.dispatchEvent("importantMessage", message);
+         *         }
+         *         else
+         *         {
+         *             // You have loaded some other js bundle. i.e. a polyfill
          *         }
          *     };
          *
--- a/blueprint/core/blueprint_ReactApplicationRoot.h
+++ b/blueprint/core/blueprint_ReactApplicationRoot.h
@@ -74,7 +74,13 @@ namespace blueprint
                 {
                     const auto lmt = wb.bundle.getLastModificationTime();
 
-                    if (lmt > wb.bundleLastModifiedTime)
+                    // In some instances webpack rebuilds temporarily result in an empty
+                    // bundle file before rebuilding. We do not want to trigger evaluation
+                    // callbacks when the bundle file is empty as this is likely to cause errors
+                    // when users attempt to call javascript functions from C/C++. Calling eval/call
+                    // in duktape with an empty js file/string does not result in an error and simply
+                    // pushes undefined to the top of the duktape stack as the
+                    if (lmt > wb.bundleLastModifiedTime && wb.bundle.loadFileAsString().isNotEmpty())
                     {
                         onBundleChanged(wb.bundle);
                         wb.bundleLastModifiedTime = lmt;
@@ -119,11 +125,12 @@ namespace blueprint
         {
             JUCE_ASSERT_MESSAGE_THREAD
 
-            // Initialise the EcmascriptEngine
-            initEngine();
-
             // Initialise the ViewManager
             initViewManager();
+
+            engine.onUncaughtError = [this](const juce::String& msg, const juce::String& trace) {
+                handleBundleError(trace);
+            };
 
 #if JUCE_DEBUG
             enableHotReloading();
@@ -160,9 +167,9 @@ namespace blueprint
 #if JUCE_DEBUG
            const auto startDebugCommand = juce::KeyPress('d', juce::ModifierKeys::commandModifier, 0);
 
-           if (engine && key == startDebugCommand)
+           if (key == startDebugCommand)
            {
-               engine->debuggerAttach();
+               engine.debuggerAttach();
            }
 #endif
             return true;
@@ -183,13 +190,19 @@ namespace blueprint
             // Clear error state from previous js evals
             errorText.reset();
 
-            if (beforeBundleEval)
-                beforeBundleEval(std::nullopt);
+            if (!bundleValid)
+            {
+                // Install React.js backend rendering methods
+                registerNativeRenderingHooks();
 
-            jassert(engine);
-            auto result = engine->evaluate(bundle);
+                if (beforeBundleEval)
+                    beforeBundleEval(std::nullopt);
+            }
 
-            if (afterBundleEval)
+            auto result = engine.evaluate(bundle);
+            bundleValid = result != EcmascriptEngine::EvaluationError;
+
+            if (bundleValid && afterBundleEval)
                 afterBundleEval(std::nullopt);
 
             return result;
@@ -208,23 +221,28 @@ namespace blueprint
         {
             JUCE_ASSERT_MESSAGE_THREAD
 
-            jassert(bundle.existsAsFile());
-            jassert(engine);
-
             // Clear error state from previous js evals
             errorText.reset();
 
-            if (beforeBundleEval)
-                beforeBundleEval(bundle);
+            if (!bundleValid)
+            {
+                // Install React.js backend rendering methods
+                registerNativeRenderingHooks();
 
-            auto result = engine->evaluate(bundle);
+                if (beforeBundleEval)
+                    beforeBundleEval(bundle);
+            }
 
-            if (afterBundleEval)
+            auto result = engine.evaluate(bundle);
+            bundleValid = result != EcmascriptEngine::EvaluationError;
+
+            if (bundleValid && afterBundleEval)
                 afterBundleEval(bundle);
 
             if (hotReloadEnabled)
             {
                 jassert(bundleWatcher);
+
                 if (!bundleWatcher->watching(bundle))
                     bundleWatcher->watch(bundle);
             }
@@ -241,13 +259,13 @@ namespace blueprint
         void dispatchViewEvent (ViewId viewId, const juce::String& eventType, T... args)
         {
             JUCE_ASSERT_MESSAGE_THREAD
-            jassert(engine);
 
-            // engine may have been reset in the event of a bundle eval error. We don't want to
-            // crash if client code tries to immediately dispatch an event after bundle loading.
-            // ReactApplicationRoot should report a generic error to the user rather than crashing.
-            if (engine)
-                engine->invoke("__BlueprintNative__.dispatchViewEvent", viewId, eventType, std::forward<T>(args)...);
+            //TODO: Provide safety handle here ? Cannot rely on bundleValid as otherwise we
+            //      miss out on initial dispatch of things like onMeasure in call to evaluate.
+            //      Could give EcmascriptEngine a reset() function ?
+            //      Alternatively we could set bundleValid to true right before call to evaluate() ?
+
+            engine.invoke("__BlueprintNative__.dispatchViewEvent", viewId, eventType, std::forward<T>(args)...);
         }
 
         /** Dispatches an event through Blueprint's EventBridge. */
@@ -255,13 +273,13 @@ namespace blueprint
         void dispatchEvent (const juce::String& eventType, T... args)
         {
             JUCE_ASSERT_MESSAGE_THREAD
-            jassert(engine);
 
-            // engine may have been reset in the event of a bundle eval error. We don't want to
-            // crash if client code tries to immediately dispatch an event after bundle loading.
-            // ReactApplicationRoot should report a generic error to the user rather than crashing.
-            if (engine)
-                engine->invoke("__BlueprintNative__.dispatchEvent", eventType, std::forward<T>(args)...);
+            //TODO: Provide safety handle here ? Cannot rely on bundleValid as otherwise we
+            //      miss out on initial dispatch of things like onMeasure in call to evaluate.
+            //      Could give EcmascriptEngine a reset() function ?
+            //      Alternatively we could set bundleValid to true right before call to evaluate() ?
+
+            engine.invoke("__BlueprintNative__.dispatchEvent", eventType, std::forward<T>(args)...);
         }
 
         //==============================================================================
@@ -405,22 +423,25 @@ namespace blueprint
         BundleEvalCallback afterBundleEval;
 
         //==============================================================================
-        /** The ReactApplicationRoot's engine instance. Note it is possible for engine to be a nullptr in the event of a bundle eval callback. */
-        std::unique_ptr<EcmascriptEngine> engine;
+        /** The ReactApplicationRoot's engine instance. */
+        EcmascriptEngine engine;
 
     private:
         //==============================================================================
         void handleBundleChanged(const juce::File& bundle)
         {
-            initViewManager();
-            registerNativeRenderingHooks();
+            JUCE_ASSERT_MESSAGE_THREAD
 
+            initViewManager();
             evaluate(bundle);
         }
 
         //==============================================================================
         void handleBundleError(const juce::String& trace)
         {
+            JUCE_ASSERT_MESSAGE_THREAD
+
+            bundleValid = false;
             viewManager.reset();
 
             errorText = std::make_unique<juce::AttributedString>(trace);
@@ -432,20 +453,6 @@ namespace blueprint
             errorText->setFont(18);
 #endif
             repaint();
-        }
-
-        //==============================================================================
-        /** Initialises the EcmascriptEngine and registers our native render hooks and a generic error handler/reporter */
-        void initEngine()
-        {
-            engine = std::make_unique<EcmascriptEngine>();
-
-            // Install React.js backend rendering methods
-            registerNativeRenderingHooks();
-
-            engine->onUncaughtError = [this](const juce::String& msg, const juce::String& trace) {
-                handleBundleError(trace);
-            };
         }
 
         //==============================================================================
@@ -513,11 +520,9 @@ namespace blueprint
 
         void registerNativeRenderingHooks()
         {
-            jassert(engine);
+            engine.registerNativeProperty("__BlueprintNative__", juce::JSON::parse("{}"));
 
-            engine->registerNativeProperty("__BlueprintNative__", juce::JSON::parse("{}"));
-
-            engine->registerNativeMethod("__BlueprintNative__", "createViewInstance", [](void* stash, const juce::var::NativeFunctionArgs& args) {
+            engine.registerNativeMethod("__BlueprintNative__", "createViewInstance", [](void* stash, const juce::var::NativeFunctionArgs& args) {
                 auto self = reinterpret_cast<ReactApplicationRoot*>(stash);
 
                 jassert (self != nullptr);
@@ -533,7 +538,7 @@ namespace blueprint
                 return juce::var(viewId);
             }, (void *) this);
 
-            engine->registerNativeMethod("__BlueprintNative__", "createTextViewInstance", [](void* stash, const juce::var::NativeFunctionArgs& args) {
+            engine.registerNativeMethod("__BlueprintNative__", "createTextViewInstance", [](void* stash, const juce::var::NativeFunctionArgs& args) {
                 auto self = reinterpret_cast<ReactApplicationRoot*>(stash);
 
                 jassert (self != nullptr);
@@ -548,7 +553,7 @@ namespace blueprint
                 return juce::var(viewId);
             }, (void *) this);
 
-            engine->registerNativeMethod("__BlueprintNative__", "setViewProperty", [](void* stash, const juce::var::NativeFunctionArgs& args) {
+            engine.registerNativeMethod("__BlueprintNative__", "setViewProperty", [](void* stash, const juce::var::NativeFunctionArgs& args) {
                 auto self = reinterpret_cast<ReactApplicationRoot*>(stash);
 
                 jassert (self != nullptr);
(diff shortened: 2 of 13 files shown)


blueprint/core/blueprint_EcmascriptEngine.cpp
```diff
@@ -126,6 +161,20 @@ namespace blueprint
         return result;
     }
 
+    juce::var EcmascriptEngine::evaluate (const juce::File& code)
+    {
+        juce::var result = juce::var::undefined();
+
+        if (detail::safeCompileFile(ctx, code, onUncaughtError))
```
nick-thompson, line 168: Maybe we should return the result in a successful run also? The bundles we're using never really return anything but I could imagine people using `EcmascriptEngine` for all sorts of things.

```
if (detail::safeCompileFile(ctx, code, onUncaughtError))
    duk_call(ctx, 0);

result = readVarFromDukStack(ctx, -1);
duk_pop(ctx);

return result;
```

blueprint/core/blueprint_EcmascriptEngine.cpp
```diff
@@ -228,6 +277,31 @@ namespace blueprint
...
+        duk_trans_socket_waitconn();
+
+        duk_debugger_attach(ctx,
+                            duk_trans_socket_read_cb,
+                            duk_trans_socket_write_cb,
+                            duk_trans_socket_peek_cb,
+                            duk_trans_socket_read_flush_cb,
+                            duk_trans_socket_write_flush_cb,
+                            NULL,
+                            detail::handle_debugger_detached,
```
nick-thompson, line 295: For little functions like this I prefer inline lambdas (which cast to C-style function pointers as long as there's no capture group):

```
duk_debugger_attach(ctx,
    etc,
    etc,
    [](duk_context* ctx, void*) { duk_trans_socket_finish(); },
    NULL
);
```

blueprint/core/blueprint_ReactApplicationRoot.h
```diff
@@ -132,6 +133,21 @@ namespace blueprint
                 View::paint(g);
             }
         }
+        //==============================================================================
+        bool keyPressed(const juce::KeyPress& key) override
+        {
+            if (jsDebugEnabled)
```
nick-thompson, line 139: Maybe we just `#if JUCE_DEBUG` this stuff and omit the `enableJSDebug()` method? I feel like it's fair to just always turn this on when in debug mode?

blueprint/duktape/src-noline/duk_config.h
```diff
@@ -2898,7 +2898,7 @@ typedef struct duk_hthread duk_context;
 #undef DUK_USE_DEBUGGER_DUMPHEAP
 #undef DUK_USE_DEBUGGER_INSPECT
 #undef DUK_USE_DEBUGGER_PAUSE_UNCAUGHT
-#undef DUK_USE_DEBUGGER_SUPPORT
+#define DUK_USE_DEBUGGER_SUPPORT
```
nick-thompson, line 2901: Yea, definitely approaching the time where we should figure out the right way to do this :)

My only thought here– this would enable for Release mode too yea? I feel like it's probably helpful to compile Release without the debugger support

blueprint/core/blueprint_ReactApplicationRoot.h
JoshMarler: Yeah I was going to ask exactly this actually! Nice. Will do so.

blueprint/core/blueprint_EcmascriptEngine.cpp
JoshMarler: Well I just learnt a thing. Agreed.

blueprint/duktape/src-noline/duk_config.h
JoshMarler: Yeah indeed. I'm changing a couple more of these defines so I'll workout how to switch this between debug and release sensibly. We're happy just hacking this file inline for now though then?

nick-thompson: :+1: as long as we can do it sensibly between debug/release, yea I think we can make this change inline.

When this commit goes in, let's add a new issue to discuss how to better integrate duktape and referring to this commit to clarify/track that we edited these files inline. That way we keep track of these changes as we consider better Duktape integrations

JoshMarler: OK Awesome. I'll get this all cleaned up.

JoshMarler: Will move all the new defines to bottom of duk_config.h with a comment and wrap under JUCE_DEBUG.

JoshMarler:
@nick-thompson, could do with a quick opinon on a thing. 

Should we include a default launch.json file for the duktape debugger setup? For example, if I load the jsui folder in GainPlugin/Source/jsui as the workspace root the launch.json for a duktape debug task looks as follows:

```
{
    // Use IntelliSense to learn about possible attributes.
    // Hover to view descriptions of existing attributes.
    // For more information, visit: https://go.microsoft.com/fwlink/?linkid=830387
    "version": "0.2.0",
    "configurations": [
        {
            "name": "Attach (duk)",
            "type": "duk",
            "request": "attach",
            "address": "localhost",
            "port": 9091,
            "localRoot": "${workspaceFolder}",
            "sourceMaps": true,
            "outDir": "${workspaceFolder}/build/js",
            "stopOnEntry": false,
            "debugLog": false
        }
    ]
}
```

Some users may wish to load the workspaceFolder as the root GainPlugin directory, i.e. if they wish to edit C++ in VsCode and stay in one editor session for everything. What do you think we should provide as the example for this?

blueprint/core/blueprint_EcmascriptEngine.cpp
JoshMarler: Oh yeah. I think there will only be a result object on the stack in an error state with the duk_compile calls. Otherwise it's a function object. I'll sanity check when I get back to me desk.

nick-thompson: Ah right but then executing that function should return a result right? So if the compile succeeds, execute whatevers on top and return the result. If the compile fails, return whatevers on top because that shows the error. Although maybe I'm overthinking it... even if your bundle had a return value, it might be the case that when duktape compiles it and puts a function at stack top, _that_ function might not have a return.

/shrug maybe ignore me :)

JoshMarler: Yeah. Getting back to this shortly so I'll.just run it under the debugger and check. Not super clear from the API docs but I've not diced into duktape itself much yet.

nick-thompson:
Good question. I lean towards leaving editor config files out of the repository because it's such a personalized thing, but we should _definitely_ have a clear guide in the docs for this. I think Create React App's "Setting up your editor" is a good example: https://create-react-app.dev/docs/setting-up-your-editor/

blueprint/core/blueprint_EcmascriptEngine.cpp
JoshMarler: @nick-thompson, you are indeed correct. https://duktape.org/api.html#duk_call

JoshMarler: I miss understood the docs

blueprint/core/blueprint_ReactApplicationRoot.h
JoshMarler: @nick-thompson, in which case should ReactApplicationRoot call setWantsKeyboardFocus() in it's constructor? Seems like we probably want to make that the standard?

blueprint/core/blueprint_EcmascriptEngine.cpp
nick-thompson: Cool, thanks for the link

blueprint/core/blueprint_ReactApplicationRoot.h
nick-thompson: Yea, that sounds good to me

JoshMarler:
@nick-thompson. Quick question about how you think hot-reload and debug should behave together. Because we reset the EcmascriptEngine object during a bundle reload the debugging session is terminated. Should we concern ourselves with this? Is there any need to attempt persisting debug sessions whilst making changes with a watch running? I'm not sure how this would work right now.

Perhaps we could get away with only resetting the EcmascriptEngine instance in the event of an evaluation error?

nick-thompson:
Yea, another good question, and that makes me realize– if we require CMD+D/CTRL+D to attach the debugger, then how would you catch a `debugger` statement or an uncaught error that comes up in the initial bundle evaluation? I suppose you'd have to CMD/CTRL+D to attach the debugger, then maybe CMD/CTRL+R to reload/re-evaluate the bundle?

This actually makes me think I was wrong earlier– perhaps carrying a member flag `enableDebugging` on ReactApplicationRoot is a good idea (toggled `true`/`false` by CMD/CTRL+D), and at the time that we reset the EcmascriptEngine (basically right before `beforeBundleEval`) we can also say `if (enableDebugging) { engine->attach(); }` to debug the bundle eval step.

That would also satisfy your question: if you've enabled debugging with CMD/CTRL+D, we can do the hot reload like normal and attach the debugger again before the bundle eval.

What do you think?

JoshMarler:
> Yea, another good question, and that makes me realize– if we require CMD+D/CTRL+D to attach the debugger, then how would you catch a `debugger` statement or an uncaught error that comes up in the initial bundle evaluation? I suppose you'd have to CMD/CTRL+D to attach the debugger, then maybe CMD/CTRL+R to reload/re-evaluate the bundle?
> 
> This actually makes me think I was wrong earlier– perhaps carrying a member flag `enableDebugging` on ReactApplicationRoot is a good idea (toggled `true`/`false` by CMD/CTRL+D), and at the time that we reset the EcmascriptEngine (basically right before `beforeBundleEval`) we can also say `if (enableDebugging) { engine->attach(); }` to debug the bundle eval step.
> 
> That would also satisfy your question: if you've enabled debugging with CMD/CTRL+D, we can do the hot reload like normal and attach the debugger again before the bundle eval.
> 
> What do you think?

Yeah this is basically the sort of scheme I had in mind initially but I thought I'd maybe over-thought/over-engineered it a bit. OK. I'll have an experiment with this. If fixed everything else otherwise and then I can add a simple addition to the README with VSCode editor setup and debug instructions. 

I want to give this a good test on Linux, Mac and Windows to sanity check before merging.

blueprint/core/blueprint_EcmascriptEngine.cpp
JoshMarler: @nick-thompson, Now that I'm taking a second pass, I think I might change this to use duk_pcall for safety in the same way that you have duk_pcal in EcmascriptEngine::invoke.

JoshMarler:
@nick-thompson, the CMD-R reload idea might still be nice regardless actually.

JoshMarler:
@nick-thompson, I'm finishing this off at the moment. 

One question, 

As a user of a debug client, I probably expect detaching (i.e. hitting stop in vscode) to switch off the ```enableDebugging``` flag. i.e. I don't want blueprint to continue calling ```duk_debugger_attach``` (blocking the UI) each time there is a hot-reload once I have deliberately disconnected from the debug client. 

However, there is no simple mechanism in duktape to determine from where a ```duk_debug_detached_function``` callback has originated. i.e. it may have been triggered from the debug client (stop in vscode) or it may have been triggered by some socket connection error, or a hot-reload etc. So there is no straight forward way to reset ```enableDebugging``` when a detach is triggered from a debug-client vs leaving ```enableDebugging``` set when detach is triggered from a hot-reload.

There is probably a long winded way to make this work using the user data pointer supplied by duktape to the ```duk_debug_detached_function``` callback and passing some flag/callback from ```ReactApplicationRoot``` into ```EcmascriptEngine::debuggerDetach()```. However, this feels a bit overkill and potentially error-prone for this edge case? It could be a "debug support 2.0". 

To allow the user to terminate the re-attach behaviour but keep the app running, do we want to offer some sort of CTRL-C/CTRL-E command which forcibly detaches the debugger and resets to ```enableDebugging``` flag? This behaviour would need to be added to the docs as a "thing to know". 

i.e. once you hit CTRL-D/CMD-D blueprint with alway await a debug client connection on a hot-reload until you forcibly reset this behaviour with CTRL-C/CTRL-E. 

What do you think?

JoshMarler:
@nick-thompson, thinking about this more. Is it actually necessary for us to be resetting the EcmascriptEngine object in ```ReactApplicationRoot::handleBundleChanged``` and ```ReactApplicationRoot::handleBundleError``` ? The errors we receive in ```handleBundleError``` look like they should all be recoverable as opposed to the fatal errors handled by ```fatalErrorHandler``` in blueprint_EcmascriptEngine.cpp. 

If we didn't reset engine in these calls that would simplify things greatly. provide we always use the safe variants of the duktape functions like duk_pcall, duk_peval I can't see from the API docs that we are then unable to reuse the context/engine. We always evaluate the bundle as a function and then pop said bundle off of the duk stack. So can we not simply evaluate the changed bundle again in a reload and do away with the engine/context reset?

nick-thompson:
> However, there is no simple mechanism in duktape to determine from where a `duk_debug_detached_function` callback has originated. i.e. it may have been triggered from the debug client (stop in vscode) or it may have been triggered by some socket connection error, or a hot-reload etc. So there is no straight forward way to reset `enableDebugging` when a detach is triggered from a debug-client vs leaving `enableDebugging` set when detach is triggered from a hot-reload.
> 
> There is probably a long winded way to make this work using the user data pointer supplied by duktape to the `duk_debug_detached_function` callback and passing some flag/callback from `ReactApplicationRoot` into `EcmascriptEngine::debuggerDetach()`. However, this feels a bit overkill and potentially error-prone for this edge case? It could be a "debug support 2.0".

Interestinggg, thanks for digging into that. I see what you mean, there might be a way to thread some `void* udata` through there but I agree that it seems error-prone and tricky. I like your suggestion, let's call that "debugging 2.0," come back to it later, and for now implement the "manually detach" hotkey. So we'll have CTRL+D = attach, CTRL+E = detach, CTRL+R = re-evaluate the loaded bundle?

nick-thompson:
> @nick-thompson, thinking about this more. Is it actually necessary for us to be resetting the EcmascriptEngine object in `ReactApplicationRoot::handleBundleChanged` and `ReactApplicationRoot::handleBundleError` ? The errors we receive in `handleBundleError` look like they should all be recoverable as opposed to the fatal errors handled by `fatalErrorHandler` in blueprint_EcmascriptEngine.cpp? https://duktape.org/guide.html#normal-and-fatal-errors
> 
> If we didn't reset engine in these calls that would simplify things greatly. Provided we always use the safe variants of the duktape functions like duk_pcall, duk_peval I can't see from the API docs that we are then unable to reuse the context/engine. Am I right in saying we always evaluate the bundle as a function and then pop said function off of the duk stack after calling it? So can we not simply evaluate the changed bundle again in a reload and do away with the engine/context reset?

Right, this is a great question and I think worth a careful discussion. If we can get away without resetting the engine and that simplifies things like debugging for us, it's obviously a win. I spent a min with the docs again and I think you're right, but I see two "gotchas" here that I think we should figure out: the first is what happens if the user doesn't supply the `handleError` callback? For the engine that ReactApplicationRoot owns, we can easily guarantee that this callback is provided, but what does that look like? In debug mode it's great to pop up that red screen unless the user has overriden the error handler callback, but what is sensible behavior for release mode? Is it fair to pop up that red screen in debug _and_ release and enforce that the user, if they want different behavior, must override? Maybe it is... Then the second one is that the user always has access this global state in the ecmascript engine via the `global` keyword (this is common in js engines, it's `window` in browsers and `global` in node), and there's nothing to prevent the user from writing state to `global`. When we reset the engine, we guarantee that the next bundle eval will start from a clean slate. When we don't, the user kind of has this opportunity to shoot themselves in the foot. I can make arguments for either case, but I think it's something we should at least be aware of and careful of as we make this decision

JoshMarler:
1. "what happens if the user doesn't supply the handleError callback"

Yeah I thought about this. So, we could make this a requirement, i.e. a constructor dependency on EcmascriptEngine. Failing that, from a personal point of view I'd have no objection to the red error screen being displayed in production as my bundles are minified and stack traces/errors are obfuscated. Obviously that doesn't speak for all the other users out there! Clearly stating in the docs that users must override this if they wish different behaviour seems fair to me though?

I also think we might want to provide callbacks for both "normal" and "fatal" errors. This is particularly useful in embedded environments. Then perhaps clearly state the engine/context CANNOT be reused in the event of a fatal error etc. We could link to the duktape docs there?

2. "user always has access to global state in the ecmascript engine via the global keyword"

The idea being that a user could modify the ```global``` object/state and this may have side effects on bundle eval because the expect ```global``` to be untouched at evaluation? Yeah you would know far more about the pitfalls here than I. I think you'd be best to make the call whether we can state "Avoid this as best practice and if you modify ```global``` be aware you do so at your own risk. Generally speaking this should only be an issue in debug builds when hot-reload is enabled (at least for the moment).

JoshMarler:
"Then they think they've made their app robust, never see the red screen anymore and forget to write a release-mode override, suddenly they're shipping an app that red-screens on their users in random edge cases"

Yeah I totally get the concern. The main dev process is probably going to come through ReactApplicationRoot which will have the default error handler switched on in debug builds. It might not be a major issue to throw an unhandled exception if no error handler is supplied to be fair?. After all the user can't expect their component to render after an eval error and if they do want to handle production evaluation error in any specific way they override the handler. This could be clearly documented. 

"Anyway I'm generally in favor of finding the right way to not reset the engine because I do see that it would simplify our code"

I think the good thing about it is it delivers a pretty familiar development experience to people used to working with hot-reload and debug in a browser. Definitely needs careful code/consideration though. I've got the basics running fine but need to stamp out all the little gotchas.

JoshMarler:
@nick-thompson, the other thing I think we might need to do is to copy out our own version of duk_trans_socket_unix.c from the duktape examples to make a couple modifications. In some cases disconnecting the vscode debugger and interacting with the UI soon after can cause a SIGPIPE to be triggered when attempting to write debug infomation to the socket associated with the debug tcp connection. 

The duk_cmdline.c app actually swallows SIGPIPE to stop it from terminating:

```
static void set_sigint_handler(void) {
	(void) signal(SIGINT, my_sighandler);
	(void) signal(SIGPIPE, SIG_IGN);  /* avoid SIGPIPE killing process */
}
```
I suspect we may want a similar handler somewhere or a more refined version of duk_trans_socket_unix.c which handles SIGPIPE signals.

nick-thompson:
Hmm how about this:

* For the ecmascript engine reset, let's ignore the case where the user can shoot themselves in the foot by writing to `global`. It's generally just a bad practice and I don't think we should go out of our way to support a bad practice
* For the error screen, I'm cool with the idea of showing the red screen in both debug and release and making it clear in the docs that they should override that for release builds
* I'm not sure I follow the "normal"/"fatal" error distinction– isn't a fatal error just a normal error that was never caught in javascript land? I think we only need the one "fatal" error handler, and if the user wants to get more fine-grained error handling they can always try/catch in javascript and call their own native metods if they need to handle that back in C++
* Re: sockets, interesting, this is definitely not in my wheelhouse either :) I'm happy with whatever we need to do to get it working, and we can always revisit later to improve

JoshMarler:
Awesome, thanks Nick. Nice clear concise set of behaviour/requirements as always.

Cool. I agree with all of the above. I will sanity check on the distinction between "fatal"/"normal" errors in the docs when I'm back at my desk. I was under the impression that normal errors meant you could carry on using the duktape context (i.e. if we do not dispose engine in "normal" errors) whilst "fatal" errors are basically "something has gone terribly wrong, abort abort!". I probably need another read through the guide though. 

The sockets thing probably makes sense to hack inline as we've done with duk comment and just leave a sensible comment and open issue to do better.

JoshMarler:
@nick-thompson, So I'm pretty sure there is a distinction between normal and fatal errors that we should be aware of but we don't need to do anything about it for the moment other than to clearly document it I think. ```EcmascriptEngine``` provides the ```onUncaughtError``` override already, allowing users to handle "normal" errors, it also registers a fatal error handler EcmascriptEngine.cpp:133.

Normal errors are those returned by the duktape protected calls (i.e. duk_pcall etc.). Those errors can be handled either in C/C++ by examining the protected call's return code or they can be handled prior by js code by catching. 

A fatal error is an error that either is not handled or is unrecoverable. 

From the duktape docs:

"An ordinary error is caused by a throw statement, a duk_throw() API call (or similar), or by an internal, recoverable Duktape error. Ordinary errors can be caught with a try-catch in ECMAScript code or e.g. duk_pcall() (see API calls tagged protected) in C code.

A fatal error is caused by an uncaught error, an assertion failure (if enabled), an explicit call to duk_fatal(), or an unrecoverable error inside Duktape."

So for the normal error case, the default onUncaughtError handler provided by ReactApplicationRoot shouldn't reset the EcmascriptEngine instance as the duk context can still be used to evaluate code etc. This makes the debug/hot-reload workflow perform as expected. 

i.e. I can attach a debugger and start my webpack watcher, if I make an edit that causes an eval/webpack-compile error then the ReactApplicationRoot error handler is called (red screen of death) but my debug session is not terminated. I can then read the error, correct my mistake and webpack/blueprint reloads the bundle with my debug session still running. 

In the case of the fatal handler we currently crash out (throwing from EcmascriptEngine). I think at some point we may want to provide an error handler for both cases of a flag to the ```onUnhandledError``` callback identifying whether the error is normal or fatal. For now maybe it makes sense to just clarify the doc/comment around ```EcmascriptEngine::onUncaughtError``` to clarify and we could open a "Comprehensive error handling" issue?

JoshMarler:
@nick-thompson, related question. I'm seeing the error messages/traces printed to the red error screen as a jumbled mess at the moment. I feel like I've hit this before and you showed me how to correct it. Does this ring any bells? Want to ensure this is working for the new development workflow.

JoshMarler:
@nick-thompson, FYI I started a basic first draft for WIKI instructions to set this all up here: https://github.com/JoshMarler/blueprint-wiki/commit/3648d0cf44c0ef6292cf788252f8dab07fc8e824

blueprint/duktape/src-noline/duk_config.h
```diff
@@ -3739,4 +3739,15 @@ typedef struct duk_hthread duk_context;
 #error unsupported: byte order detection failed
 #endif  /* defined(DUK_USE_BYTEORDER) */
 
+
+#if JUCE_DEBUG
+//TODO: We need a better system for allowing custom/user configured duk_config headers
```
JoshMarler, line 3744: @nick-thompson we should probably create an issue for this.

nick-thompson: 💯 for sure. I'll do that now

blueprint/duktape/examples/debug-trans-socket/duk_trans_socket_unix.c
```diff
@@ -241,7 +241,15 @@ duk_size_t duk_trans_socket_write_cb(void *udata, const char *buffer, duk_size_t
 	 * timeout here to recover from "black hole" disconnects.
 	 */
 
+	//TODO: Hacked the example implementation for now to avoid SIGPIPE errors terminating the
```
nick-thompson, line 244: I'll make an issue for this too

JoshMarler:
> @JoshMarler this looks excellent, thanks so much for your work. I'll wait for your update on how the testing goes but otherwise I'm happy to merge! Also I read through the wiki edits and I love it, happy to merge that ASAP as well.
> 
> Re: the jumbled error screen text, can you send a screenshot? I suspect this is the result of the minifier in webpack but hard to say without seeing it.
> 
> Re: errors, ahh ok, thanks for the clarification. So that makes me think that in almost all cases we'll be dealing with recoverable errors (because we do everything in a `duk_pcall` and pop up the red screen when we catch something). Even if it's an uncaught user exception, we'll still catch it at the pcall boundary. It sounds like we'll hit the fatal error handler largely in 2 cases: explicit calls to duk_fatal and duktape internal errors. So as it concerns the recoverable errors, I think that gives us plenty of reason to go ahead with _not_ resetting the EcmascriptEngine in the error handler, which helps keep the debugging simple and sensible. Then for the fatal errors, I anticipate it being a very low percentage of errors, so I think it would be fair to not let this case impact the design decisions around how ReactAppRoot consumes EcmascriptEngine, so long as we provide the user a way to deal with fatal crashes in production (mostly I mean that we should not concern this case too much in our decisions around hot reloading and debugging)

Yeah awesome. Maybe we can provide a separate handler override for fatal errors or give the existing one some sort of flag. Probably worth raising an issue. I'm just giving mac a bit of a test. I have to confess I rarely test Windows so I will need to get my Windows machine setup for blueprint now to check everything.

blueprint/duktape/src-noline/duk_config.h
nick-thompson: https://github.com/nick-thompson/blueprint/issues/62

nick-thompson:
> Yeah awesome. Maybe we can provide a separate handler override for fatal errors or give the existing one some sort of flag. Probably worth raising an issue. I'm just giving mac a bit of a test. I have to confess I rarely test Windows so I will need to get my Windows machine setup for blueprint now to check everything.

Sounds great. Yea, separate handler or flag for fatals sounds good, I'll open an issue there to

And yea, I'll go ahead and fetch/merge the wiki when I can!

blueprint/duktape/examples/debug-trans-socket/duk_trans_socket_unix.c
nick-thompson: https://github.com/nick-thompson/blueprint/issues/63

JoshMarler:
image

@nick-thompson, here is a screenshot of the mangled error. I'll dig into it but thought I would post here incase you've seen similar. Using ```webpace --mode=development```

nick-thompson:
Oh yup, that's webpack trying to add terminal colors to the output. I forget exactly what you need to do but basically need to configure webpack to not use colors

JoshMarler:
> Oh yup, that's webpack trying to add terminal colors to the output. I forget exactly what you need to do but basically need to configure webpack to not use colors

Awesome. Cheers, I'll do a bit of Googling.

JoshMarler:
@nick-thompson, just testing on mac. I hadn't tested mac since the duk_config change. Looks like JUCE_DEBUG isn't defined at the point of processing duk_config.h on mac ... Need to check why. Just in case you do to experiment.

JoshMarler:
@nick-thompson, mac all good. Just testing Windows but build is failing at the moment so will drop you a separate PR shortly fixing some compiler errors. My fault, I've not been testing Windows enough. Will be great when Blueprint has some sort of CI pipeline so save us from fools like myself :-)

nick-thompson:
Agreed! There's an issue open for that... I'll try to tackle it one of these days :D

JoshMarler:
@nick-thompson, FYI I have found why debugging isn't working on Windows using the extension. I've raised an issue here: https://github.com/harold-b/vscode-duktape-debug/issues/63

I'm going to test hacking this out locally. We could potentially offer a modified version of the extension bundle/dist folder Blueprint users can copy until this is resolved. I'm not sure how actively  supported the extension is. I could fork it and we could reference the fork in the docs etc.

JoshMarler:
@nick-thompson, I've found one additional issue whilst I've been back and forth between hot-reloading and debugging on windows.

When running the GainPlugin example we hit exceptions in `EcmascriptEngine::invoke` when calling `duk_require_function`. This is because the `BlueprintGenericPluginEditor` triggers `parameterValueChanged` in it's `afterBundleEval` callback which dispatches an event via the engine. 

The mistake here is that `afterBundleEval` is called and `BlueprintGenericPluginEditor` has no knowledge of the fact that there was an evaluation error. We got away with this before because `ReactApplicationRoot` checks for `if (engine)` before calling `engine->invoke`. The `engine` instance would have been `nullptr` in the old approach where we reset the engine in the event of an error. 

So, we need to give afterBundleEval callbacks someway to determine whether bundle evaluation passed or failed. I suspect the thing that makes the most sense here is to return a result from `EcmascriptEngine::evaluate` which clearly signals that an eval error has occured:

```
 if (onUncaughtError != nullptr)
                {
                    const juce::String trace = duk_safe_to_stacktrace(ctx, -1);
                    const juce::String msg = duk_safe_to_string(ctx, -1);

                    // Call the user provided error handler
                    std::invoke(onUncaughtError, msg, trace);

                }
```

Right now we simply return `juce::var::undefined`. Perhaps we could return/define some constant wrapped in a  `juce::var` which signals that an error has occured and pass this to the `afterBundleEval` callback. i.e. `juce::var result = juce::String("Evaluation_Error")` etc. 

The `afterBundleEval` callback type would become something like:

`using BundleEvalCallback = std::function<void(std::optional<juce::File> bundle, juce::var result)>;`

What do you think?

JoshMarler:
Cannot work this out. There should be no race condition with the flag stored in `afterBundleEval` on `BlueprintGenericEditor` but I'm still seeing occasional throws from `duk_require_function` when dispatching the `parameterValueChange` event.

```
        appRoot.afterBundleEval = [=] (std::optional<juce::File> bundle, juce::var result)
        {
            if (bundle && bundle->getFullPathName() == bundleFile.getFullPathName())
            {
                if (result != EcmascriptEngine::EvaluationError)
                {
                    bundleValid = true;
                    afterBundleEvaluated();
                }
                else
                {
                    bundleValid = false;
                }
            }
        };
    }
```

```
      throttleMap.throttle(parameterIndex, 1000.0 / 30.0, [=]() mutable {
            juce::MessageManager::callAsync([=]() mutable {
                if (bundleValid)
                {
                    appRoot.dispatchEvent(
                        "parameterValueChange",
                        parameterIndex,
                        id,
                        defaultValue,
                        newValue,
                        stringValue
                    );
                }
            });
        });
```

JoshMarler:
@nick-thompson, aaahhhh OK. I think I have fixed the issue. I was filling up the duk value stack by repeatedly calling `EcmascriptEngine::registerNativeRenderingHooks` in response to a bundle change. Effectively registering duplicates. I've removed this and now only `registerNativeRenderingHooks` in the EcmascriptEngine constructor and it's all working as expected. This works as we no longer dispose the duktape ctx in response to an error. 

Would still be good to get your thoughts on where the responsibility for avoiding calls to `dispatchEvent` should live when we are in the middle of a bundle eval error. i.e. do we pass it to `afterBundleEval` or do we set some state on `EcmascriptEngine` to signal that we have a bundle eval error until the next successful call to `evaluate` (could get more complex with multiple bundles but I think that whole story is a thing for later)

JoshMarler:
@nick-thompson, Now that I'm looking at the lambda release pool stuff along with this I'm starting to wonder if changing the reset of EcmascriptEngine might be a rather larger task than I had initially realised right now ... It may be that we keep the reset and users just have to be aware they will need to reconnect the debugger manually if they are hot-reloading at the same time etc. 

I'm reading through the code now to inform the decision ....

nick-thompson:
Hey @JoshMarler this is awesome, sorry I didn't get back to you sooner!

Ok, got you, great catch on the duplicte native hooks, that makes sense. Sounds like a triiiicky bug lol

Re: the callbacks, yea, I haven't thought about this case but you're totally right. My gut is that it would feel nicer, from the user's perspective, for `afterBundleEval` to only be invoked in the case of a successful evaluation. I hear you though, it could get tricky to implement that. After a brief minute with it, I think my proposal would be this: change the return types of our `invoke`/`evaluate` methods from `juce::var` to `std::optional<juce::var>`. The optional will be empty if we hit an error along the way (and if you've defined the `onUncaughtError` callback you will have received your call to it), and otherwise will contain the result of the evaluation (which might be `undefined`, but that would be different from an empty optional). Then in ReactAppRoot we would do something like:

```
if (auto result = engine.evaluate(bundle) && afterBundleEval) {
    afterBundleEval();
}
```

How does that sound?

Also, love this trick!
```
duk_push_context_dump(ctx);
DBG(duk_to_string(ctx, -1));
duk_pop(ctx);
```
I didn't know about it but I can tell immediately how useful that will be

nick-thompson:
Oh, just saw the comment on the release pool– yea, I would say don't worry about tackling that yet. I think it's fair to try to limit the scope of this PR and come back in a second PR to try to address that, whether that involvese keeping the Engine reset in place for now or allowing the release pool to grow. Either way we can come back and fix those, but I definitely understand the feeling when a PR feels like it's growing!

JoshMarler:
@nick-thompson, Yeah OK  interesting. I basically just had the following at the moment. 

EcmascriptEngine held a static member type to denote an error:

```
 static constexpr auto EvaluationError = "EcmascriptEngineEvalError";
```

evaluate then returned said code in the event of an error:

```
    juce::var EcmascriptEngine::evaluate (const juce::File& code)
    {
        juce::var result = juce::var::undefined();

        if (!detail::safeCompileFile(ctx, code, onUncaughtError))
            return juce::var(EvaluationError);

        // Call compiled function
        if (duk_pcall(ctx, 0) != DUK_EXEC_SUCCESS)
        {
            if (onUncaughtError != nullptr)
            {
                const juce::String trace = duk_safe_to_stacktrace(ctx, -1);
                const juce::String msg = duk_safe_to_string(ctx, -1);

                // Call the user provided error handler
                std::invoke(onUncaughtError, msg, trace);

                duk_pop(ctx);
                return juce::var(EvaluationError);
            }

            duk_throw_raw(ctx);
        }


        // Collect the return value
        result = readVarFromDukStack(ctx, -1);
        duk_pop(ctx);

        return result;
    }
```

However if we do this or return std::optional it's arguable that things like `EcmascriptEngine::invoke` might want to follow suit rather than returning `juce::var::undefined`. 

The `beforeBundleEval` callback disappears if we do away with the engine reset and I had a single `onBundleEval`/`afterBundleEval` callback which was of the following type to include error code:

```
        using BundleEvalCallback  = std::function<void(std::optional<juce::File> bundle, juce::var result)>;
```

User's have the `onUncaughtError` callback so I couldn't decide whether returning the error code in the bundle eval callback is of worth or not. `onUncaughtError` may be raised from multiple locations whilst passing the error result to `onBundleEval` makes it clear where the error came from and allows for special case handling in hot reload etc. etc.

JoshMarler:
> Oh, just saw the comment on the release pool– yea, I would say don't worry about tackling that yet. I think it's fair to try to limit the scope of this PR and come back in a second PR to try to address that, whether that involvese keeping the Engine reset in place for now or allowing the release pool to grow. Either way we can come back and fix those, but I definitely understand the feeling when a PR feels like it's growing!

Yeah ... It would be really nice to keep the debug/reload behaviour we get by not resetting the EcmascriptEngine by it's starting to look like that might require some significant changes. I'm gonna spend the next few hours going over it all and workout exactly what needs to be done to make it work safely and then decide whether to bail out.

JoshMarler:
@nick-thompson, hmmm. You know what. It might not need to be as complicated as I thought to sort this out. 

My worry was that I would go and clear out the LambdaReleasePool which would break the native methods and props that are no longer registered via a `beforeBundleEval` callback, because those registrations are now effectively a "one time" initialisation. Actually, we just need an equivalent callback clients use to register those same things. That way clearing the LambdaReleasePool should be safe. This could even still be the `beforeBundleEval` callback. 

I'll have an experiment and push up a change for you to peruse. Can always be rolled back if this ends up unravelling.

JoshMarler:
@nick-thompson, yeah I think keeping the EcmascriptEngine reset is the only sensible way to go at the moment with this. I've tried numerous approaches but there are just way too many subtle problems. For starters, for every bundle eval/re-eval we end up duplicating the value stack if we don't reset the duktape heap. I can see after 2 bundle reloads I have 3 copies of core.js's code in the value stack. 

There's also a tonne of subtle issues around clearing out the LambdaRelease pool etc. Given this experiment was only to support a particular development workflow I don't think it's worth going through all the pain, this code isn't required in release outside of a development workflow. At least we can attach a debugger and get stuff done. 

I'm going to have one more last run through to see if anything nice can be done in regards to debugger reconnects should people edit and reload at the same time. Might be that the the original scheme where we check a flag after bundle evaluation to see if we should re-attach the debugger works best. A better route might even be for a debug client to repeatedly attempt reconnects when it has been forcefully killed or something. 

Eesh, that was a rabbit hole to climb down .... Oh well, I actually understand how the guts of EcmascriptEngine/Blueprint work now so it was a valuable experiment. Thanks for all the back and forth on this one, I went too deep!

JoshMarler:
@nick-thompson, I can't quite believe it but even with the reset of EcmascriptEngine back in I'm still seeing the intermittent crash from `duk_require_function` in `EcmascriptEngine` invoke when reloading. 

I'm dumping the context at various points. 

In most casse after the call to `duk_pcall` in `EcmascriptEngine::evaluate` the context shows all content, i.e. it has the corejs stuff in there etc. 

In the case where we crash I can clearly see that after `duk_pcal` (without pcall returning an error) the context/stack is missing the corejs stuff and looks exactly as it did prior to `duk_pcall`, that is the stack contains no content from the bundle, it only contains native C functions/props registered in beforeBundleEval and then an `undefined` object. 

You know what ... I wonder if the timer checking the bundle state briefly see's the main.js file as empty or something when webpack rebuilds it ... maybe the rebuild of the bundle file isn't atomic and in some cases the BundleWatcher see's an empty/partially written file, thus `duk_pcall` doesn't fail and instead parses the bundle as `undefined`. 

Hmmmm .... Bit of a theory but plausible. I can't see any other cause for this intermittent issue at the moment other than something is corrupting the value stack. Possible it's my implementation of the stuff around `duk_pcompile` etc.

blueprint/core/blueprint_ReactApplicationRoot.h
```diff
@@ -74,7 +74,13 @@ namespace blueprint
...
                     const auto lmt = wb.bundle.getLastModificationTime();
 
-                    if (lmt > wb.bundleLastModifiedTime)
+                    // In some instances webpack rebuilds temporarily result in an empty
+                    // bundle file before rebuilding. We do not want to trigger evaluation
+                    // callbacks when the bundle file is empty as this is likely to cause errors
+                    // when users attempt to call javascript functions from C/C++. Calling eval/call
+                    // in duktape with an empty js file/string does not result in an error and simply
+                    // pushes undefined to the top of the duktape stack as the
+                    if (lmt > wb.bundleLastModifiedTime && wb.bundle.loadFileAsString().isNotEmpty())
```
JoshMarler, line 83: This should probably have been in a separate commit but I'll squash this all down into one when we merge.

JoshMarler:
@nick-thompson, I managed to get things working without the reset of EcmascriptEngine with some changes. The management of the duk stack and lambda release pool all works, I've done a bunch of dumping of the duktape context to check things are behaving. The below commit has a basic shape for this. 

It's a somewhat chunky/nuanced change so I'd totally understand if you felt this was too risky just to allow hot-reload support under a debugger. I'm happy to back out back to resetting the EcmascriptEngine if that is the way we decide to go. It does allow users to attach a debugger, set breakpoints and edit/hot-reload, all without having to CTRL-D and manually reconnect the debug client every-time an edit is made. 

https://github.com/JoshMarler/blueprint/commit/686179aa6619e0f72adefd8e83412947fb146ae6

blueprint/core/blueprint_EcmascriptEngine.cpp
```diff
@@ -74,29 +74,57 @@ namespace blueprint
...
-                 duk_throw_raw(ctx);
+                duk_throw_raw(ctx);
             }
 
             // Indicate success
             return true;
-        }
+       }
+
+       static void dumpContext(duk_context* ctx)
```
JoshMarler, line 122: I've been using this a lot whilst experimenting with this stuff so left it here in case you think it is worth using.

examples/GainPlugin/Source/jsui/webpack.config.js
```diff
@@ -2,8 +2,12 @@ module.exports = {
   entry: './src/index.js',
   output: {
     path: __dirname + '/build/js',
-    filename: 'main.js'
+    filename: 'main.js',
+    sourceMapFilename: "[file].map",
+     devtoolModuleFilenameTemplate: info =>
+    `webpack:///${info.absoluteResourcePath.replace(/\\/g, '/')}`
```
JoshMarler, line 8: Required to make vscode debug extension on Windows behave. See: https://github.com/harold-b/vscode-duktape-debug/issues/63

blueprint/core/blueprint_ReactApplicationRoot.h
```diff
@@ -148,13 +193,24 @@ namespace blueprint
             // Clear error state from previous js evals
             errorText.reset();
 
-            if (beforeBundleEval)
```
JoshMarler, line 151: @nick-thompson, one question around these changes. There decent chunk of duplication in the two `ReactApplicationRoot::evaluate` calls now. I've just been musing over ways to get the body of the two overrides down to a single function. I'm wondering whether `ReactApplicationRoot` could just always work with a `juce::File`. It seems this wouldn't be a big issue for callers, if for some reason direct/inline JS evaluation was required users could just call EcmascriptEngine directly. It feels like the vast majority of `ReactApplicationRoot` client code will be working with .js files. What do you think?

blueprint/core/blueprint_EcmascriptEngine.cpp
```diff
@@ -207,27 +289,55 @@ namespace blueprint
...
+            errorHandler();
 
-                // Call the user provided error handler
-                std::invoke(onUncaughtError, msg, trace);
-
-                return juce::var::undefined();
-            }
-
-            duk_throw_raw(ctx);
+            // TODO: Return specific error code if invoke fails?
```
JoshMarler, line 294: @nick-thompson, would be great if you could just weigh in on the best error return approach for things like `EcmascriptEngine::invoke` and `EcmascriptEngine::evaluate`. I hear you on potentially using std::optional but personally I'm never a huge fan of std::optional as a return type if it can be avoided. It always feels like it bloats client code. We could just switch to a generic `EcmascriptEngineError`  type for all calls or alternatively provide an enum. I can see some possible benefits on using an enum and passing this to `onUncaughtError` callbacks so client code can handle specific error cases differently in future. 

Failing that though I'm not totally against std::optional, would just be good to have a consistent pattern.

blueprint/core/blueprint_ReactApplicationRoot.h
```diff
@@ -9,7 +9,7 @@
 
 #pragma once
 
-#include <map>
+#include <optional>
```
JoshMarler, line 12: include here required to fix Windows builds. My fault. I broke this in an earlier PR as I hadn't checked Windows for a while. Used to CI telling me when I've been stupid. Apologies, I shall be more careful in future and ensure I test thoroughly on all 3 platforms!

blueprint/core/blueprint_ReactApplicationRoot.h
nick-thompson: Totally happy to have this here! I'll reply here also to your comment above about this saga since it seems a good place for it :D First, WOW. Thank you x1000 for going through that, what a pain.

>     1. Do we add checks for empty strings/bundle files in the two `EcmascriptEngine::evaluate` calls and assert and/or bail out?

Maybe I'm misunderstanding this slightly, but evaluating an empty string should always be a valid operation– an empty string is a valid expression. So for the `evaluate(const juce::String&)` overload, I think we should allow an empty string argument and make sure we don't call `duk_require_function`. If Duktape then has trouble with it, that seems like a Duktape bug (if that's the case, I have a workaround, but hope to not use it). Then for the `evaluate(const juce::File&)` overload, I'm wondering if this is a bit of a gotcha with `duk_compile_file`– maybe `duk_pcompile_string_filename` on an empty string doesn't leave a function on top? Maybe in that case, for the `evaluate(const juce::File&)` overload we should check if `file.isEmpty()` and just return early if it is?

>     2. With `EcmascriptEngine::invoke` should we only call `duk_require_function` when JUCE_DEBUG is not enabled? There is every chance we might remove some JS function whilst hot reloading that C++ code attempts to call. Alternatively, rather than removing in release, it might make sense to stick a debug print or comment next to the `duk_require_function` call, informing users that they have attempted to call a JS function which doesn't exist if it throws.

Ahh ok, so during the bundle reload, something like `CanvasView` might try to call `props["onDraw"]` which will have been wiped out already in the js environment? Originally I liked the idea of `duk_require_function` in the `invoke` implementation because if you invoke something that doesn't exist, it's a problem. But, now that I'm thinking about it again, attempting to invoke a non-existant function _is_ valid javascript, it should just throw a js error. With that in mind, I'm in favor of removing the `duk_require_function` line altogether, and then this discussion merges with the "how to handle errors the right way" discussion from below

blueprint/core/blueprint_ReactApplicationRoot.h
nick-thompson: Yea, I think that makes plenty of sense. If the user wants more direct interaction with the js engine, they can go straight to the ecmascript engine. Sounds good!

blueprint/core/blueprint_ReactApplicationRoot.h
nick-thompson: Heh, all good, sounds like it's really about time for me to handle #22 :)

blueprint/core/blueprint_ReactApplicationRoot.h
JoshMarler: Yeah duktape handles calling evaluate with empty strings just fine. I'll sanity check when the undefined object was pushed to the stack after evaluating an empty string, I thiiiink it was at the point of `duk_pcall` inside `EcmascriptEngine::evaluate`, i.e. after the compile. I'll sanity check that.  

Cool with removing `duk_require_function` the alternative I was going to suggest was only call it if `JUCE_DEBUG` is not set. 

An early return on an empty file sounds fine to me. Do we put this at the `EcmascriptEngine::evaluate` level? In theory we could just remove the change to the bundle watcher than but maybe it's worth leaving that too to avoid the needless work of the `handleBundleChanged` callback?

blueprint/core/blueprint_ReactApplicationRoot.h
JoshMarler: Cool. File based interface only on ReactApplicationRoot it is.

blueprint/core/blueprint_ReactApplicationRoot.h
JoshMarler: Ha yeah what a pain. I've done a fair amount of Jenkins in my time so shout if I can help with anything. No idea what you're thinking of using

blueprint/core/blueprint_EcmascriptEngine.cpp
nick-thompson: Yea, we should definitely figure this out. I like the std::optional idea but the hard part is that it's such an odd flow: "this might return something, and if it doesn't you should have already received a call to your uncaught error handler." Writing code like that sounds.. meh.

What do you think about this: `invoke` and `evaluate` can throw? If you try to evaluate something and it causes an error, we catch it in our own `uncaughtErrorHandler` and throw one of our own errors: either an `EcmascriptEngine::FatalError` or `EcmascriptEngine::Error`?

```cpp
try {
    auto result = engine.evaluate("2 + @");
    doSomethingWithResult(result); // Will never reach here
} catch (const EcmascriptEngine::Error& e) {
    // handle it
} catch (const EcmascriptEngine::FatalError& e) {
    // handle it
}
```

Writing try/catch code always feels a little verbose, but this feels like a good unifying story for how to interact with the engine, how to handle errors, how to differentiaite between regular and fatal errors, etc. What do you think?

nick-thompson:
Sorry I didn't get back to you sooner @JoshMarler! Busy weekend. Just left you some comments and I'm looking over the code again now

JoshMarler:
> Sorry I didn't get back to you sooner @JoshMarler! Busy weekend. Just left you some comments and I'm looking over the code again now

No worries at all. Please take as much time as you need when I'm spamming you and feel free to ignore me or ask if I can slow down for a bit! I can keep some PR's on ice and drop them over as and when you have the time.

blueprint/core/blueprint_EcmascriptEngine.cpp
```diff
@@ -107,6 +135,31 @@ namespace blueprint
...
+            // Clear out the stack so we can re-register native functions
+            // after we clear out the lambda release pool etc.
+            while (duk_get_top(ctx))
+            {
+                duk_remove(ctx, duk_get_top_index(ctx));
+            }
+
+            // Clear the LambdaHelper release pool as duktape does not call object
+            // finalizers in the event of an evaluation error or duk_pcall failure.
+            lambdaReleasePool.clear();
```
nick-thompson, line 160: Hm, I must have missed this in the earlier discussion, but if an error is successfully caught and handled, why do we have to clear the lambda pool and the duk stack?

blueprint/core/blueprint_EcmascriptEngine.cpp
```diff
@@ -117,12 +170,38 @@ namespace blueprint
...
     {
-        if (!detail::safeEvalString(ctx, code, onUncaughtError))
-            return juce::var::undefined();
+        jassert(code.isNotEmpty());
+
+        if (!detail::safeEvalString(ctx, code, errorHandler))
+            return juce::var(EvaluationError);
 
         auto result = readVarFromDukStack(ctx, -1);
+        duk_pop(ctx);
```
nick-thompson, line 179: ooh good one!

blueprint/core/blueprint_EcmascriptEngine.cpp
```diff
@@ -207,27 +289,55 @@ namespace blueprint
...
+            duk_trans_socket_peek_cb,
+            duk_trans_socket_read_flush_cb,
+            duk_trans_socket_write_flush_cb,
+            NULL,
+            [](duk_context* ctx, void* data)
+            {
+                duk_trans_socket_finish();
+
+                auto engine = static_cast<EcmascriptEngine*>(data);
+                engine->stopTimer();
```
nick-thompson, line 323: Ah nice

blueprint/core/blueprint_ReactApplicationRoot.h
nick-thompson: I like keeping the check in there on the bundle watcher, that feels like good expected behavior from the user perspective. I just want to make sure that the `EcmascriptEngine` API allows evaluating an emptry string since it is technically a valid operation... but that doesn't mean ReactAppRoot needs to let an empty bundle file go through during hot reload

blueprint/core/blueprint_ReactApplicationRoot.h
```diff
@@ -202,27 +272,29 @@ namespace blueprint
...
-            jassert(engine);
-            
-            // engine may have been reset in the event of a bundle eval error. We don't want to
-            // crash if client code tries to immediately dispatch an event after bundle loading.
-            // ReactApplicationRoot should report a generic error to the user rather than crashing.
-            if (engine)
-                engine->invoke("__BlueprintNative__.dispatchViewEvent", viewId, eventType, std::forward<T>(args)...);
+            jassert(bundleValid);
+
+            // We wish to safe guard against client code which attempts to dispatch an event immediately
```
nick-thompson, line 277: Yea this seems like a consequence of not having a consistent error handling story. Glad we're thinking about this now

blueprint/core/blueprint_EcmascriptEngine.cpp
JoshMarler: Yeah, it is all getting a bit "wordy". 

It still feels like everything should come through a single point. I guess try/catch is familiar to people. 

I wonder how bad it is to just have `onUncaughtError` take and error code for any special cases required and then have everything just return `juce::var::undefined` in the event of an error. 

For example, in `ReactApplicationRoot` we could call `evaluate` and have just `ReactApplicationRoot::handleError` responsible for setting `bundleIsValid`. Then we just return the `evaluate` result to the caller and don't check against any kinds of error codes. 

The only thing that "smells" to with the try/catch approach (as a client of the API) is "Wait .. what? I've written an `onUncaughtError` handler and I still need to stick try/catch all over the place?"

JoshMarler: Part of me still wonders if a single error code for returns and then all handling primarily through `onUncaughtError` makes sense as I'm talking this through. 

i.e. `static constexpr auto EcmascriptEngineError = "EcmascriptEngineError"`

The only other issue with the try/catch approach is callers can override `onUncaughtError` and take responsibility away from `ReactApplicationRoot`. Then it feels like it all gets a bit weird .... 

This could be solved by passing `onUncaughtError` as a constructor param to both `ReactApplicationRoot` and `EcmascriptEngine`

nick-thompson: Oh right, didn't explain that part of my thought process– with `try/catch` I think we could remove the whole idea of setting error callbacks as a user. We'll add our own error handling callbacks internally, but it will only ever surface to the user via try/catch. The reason I think that works nicely is that if `evaluate` always returns `juce::var::undefined()` in the case of an error, it's really hard to tell if you should proceed after the evaluate– you have to manage your own state to set a flag in the error handler callback and check it after every evaluate call. Although I don't love try/catch, at least it localizes the whole flow and relieves the end user from having to manage any state/flags in the error callback

JoshMarler: Cool. I'm all for this then in that case. I'll get to work on this now.

blueprint/core/blueprint_ReactApplicationRoot.h
JoshMarler: Yep totally. Should be able to remove this with the exception handling approach.

blueprint/core/blueprint_EcmascriptEngine.cpp
JoshMarler: So, as far as I can see duktape didn't call any finalizers when we reload the bundle and an evaluation errors occurs, meaning the size of the release pool may steadily increases. This may have been antagonised by what appears to be a separate problem with the `CanvasView`, with `animate` enabled the UI eventually locks up when just sat idle as there seems to be some delay garbage collection all the intermediate function objects associated with the `CanvasContext`.  That seems to result in the release poll getting huge and then when it is finally clears it blocks the UI for an age. With animate enabled under a hot-reload workflow this got even worse when there were evaluation errors as things weren't getting cleaned up. 

Also, in the event of a bundle error we may reload our bundle in a second run (after fixing an error), causing `beforeBundleEval` to be called. This results in duplicating all native props and functions on the duktape stack because we/duktape does not clear the stack of these items in the event of an error. 

This may require more scrutiny but this approach provides consistent results. i.e. when there is an error we "reset" the `EcmascriptEngine`. This requires callers fixup the bundle and reload (or if the error is in C++ they'll have to rebuild). 

Arguably we should only clear out the EcmascriptEngine when the error is an eval error on a bundle. It might make more sense to give `EcmascriptEngine` a `reset` function or something which `ReactApplicationRoot` can call under specific error cases. 

Not totally sure what the best approach is here.

JoshMarler:
> uld we try to sort that and tackle it here? Or

It suits me to do this hear if you're OK sweating through this PR Nick. Gotten to be a bit of a beast but I think it makes sense to get this all sorted here now.

blueprint/core/blueprint_ReactApplicationRoot.h
JoshMarler: @nick-thompson, resolved here: https://github.com/nick-thompson/blueprint/pull/61/commits/399e49da36a06cb5185775dd35690f1dee55b82a

Feel free to resolve this conversation if you're happy with this change.

blueprint/core/blueprint_EcmascriptEngine.cpp
JoshMarler: @nick-thompson, just to throw one last curve ball. 

The other way I could see handling errors would be for all `EcmascriptEngine` functions to return a `juce::var`. We should only ever be calling the "safe" variants of the duktape API. In which case we can return an error code from all functions when we know errors are recoverable. For things like `EcmascriptEngine::registerNativeMethod` which don't have a real return result we can just return `juce::var::undefined`. The convention being that client code using the `EcmascriptEngine` API always checks as follows:

````
if (!EcmascriptEngine::thing() == EcmascriptEngine::Error)
```
Callers could just override a "fatal" error handler if needed. I.e. for the things we don't catch. 
It would then be reasonable enough for `ReactApplicationRoot` to keep it's `handleBundleError` function and maintain it's current behaviour. 

This is slightly less verbose than try/catch potentially? At the moment every public EcmascriptEngine API call could potentially throw. This could make for pretty cumbersome client code. What do you think?

JoshMarler: Hmmm. You know what. If `ReactApplicationRoot` is probably going to be used more often than anything not the exception use case might not be quite as bad... I'll wait to hear from you just in case. I can do a decent chunk of the refactor without needing to know anyway.

JoshMarler:
@nick-thompson, You know what. I've started working on the error handling refactor and it's a reasonable chunk so if you're happy with merging the debugger changes in as is now I'll open up a separate PR for the error handling stuff. I've got a decent chunk of it implemented now but it might be nice to get this stuff in.

blueprint/core/blueprint_EcmascriptEngine.cpp
nick-thompson: Ahh interesting, I see what you mean. I agree, there's some nuance here. For example:

```js
function a() {
  try {
    return 2 + doesntExist();
  } catch (e) {
    console.log('caught it');
  }
}

a();
```

if I evaluate the above in a javascript environment, `a` is still defined, and can be invoked again. We could imagine removing that try/catch block from within `a` here and instead catching that at the cpp level with our new approach to error handling:

```cpp
try {
    engine.invoke("a");
} catch (EcmascriptEngine::Error const& e) {
    DBG("Caught it");
}
```

Here I think we might stilll expect `a` to be defined in the engine and able to invoke it again, as in the case above. Now what if this function `a` is actually a native method with an attached lambda in the pool? This case would be a problem.

That said, I do think there's a lot of nuance here and I also think this PR has grown already to take on many problems :) Let's go ahead with this part as is so that we can lock in debugging, and we can continue the conversation on error handling + implementation in a separate issue. Sound good?

blueprint/core/blueprint_EcmascriptEngine.cpp
nick-thompson: I could imagine a convention like `if (!EcmascriptEngine::thing() == EcmascriptEngine::Error)` being similarly verbose to try/catch given that all the `evaluate`/`invoke` methods could yield a case that needs to be checked, so I think I still lean to try/catch. And you're right– most users won't interact with the EcmascriptEngine API directly, it will be through ReactAppRoot, where yea I like your idea that we only do the handling for them in Debug mode where we pop up the red UI screen (and can definitely add a flag to disable that). When the user moves to production they should probably try/catch their own call to `ReactAppRoot.evaluate` there

nick-thompson:
Thanks @JoshMarler, you're a champion. Sorry for my slow replies, crazy week over here.

Yea, I'm with you, let's merge here and revisit the error handling refactor in a separate PR. I'll take one last pass here and pull it in if all good!

blueprint/core/blueprint_EcmascriptEngine.cpp
JoshMarler: Agreed, work for separate PR following this up I think. I have most of it in place.

blueprint/core/blueprint_EcmascriptEngine.cpp
JoshMarler: Sounds good to me @nick-thompson. Maybe we should raise a "Refine error handling" issue or similar to cover this and I'll drop over an initial PR in the next couple of days with some ideas.

JoshMarler: FYI, I've tested current state of this branch on all 3 platforms and everything is good. We'll just need to add Debug setup documentation to the WIKI at some point. Need to add a section detailing that Windows currently requires a hack to the vscode extension. We could just link to a branch on my own fork or something for now.

nick-thompson: Yea, perfect!

JoshMarler:
> Thanks @JoshMarler, you're a champion. Sorry for my slow replies, crazy week over here.
> 
> Yea, I'm with you, let's merge here and revisit the error handling refactor in a separate PR. I'll take one last pass here and pull it in if all good!

Not a problem! I've been hammering you with notifications I expect. Just happy to be on full steam with Blueprint at the moment.

blueprint/core/blueprint_EcmascriptEngine.cpp
JoshMarler: Cool. I'll get that WIKI page finished off an drop you a message or something when it's done and you can grab from my fork. Annoying you can't PR a wiki page!

nick-thompson: Sounds great, thanks. And yea, annoying indeed... there must be a way to make that easier

JoshMarler:
> Ok, yea, this is great! There's a couple conversations inline about the error handling, we'll just loop back on those in the followup. Happy to go ahead and merge here, mind if I rebase + squash on the way in?

Nah go ahead and rebase as you like. Wasn't too sure whether squash this all up into one or not.

nick-thompson:
Done 🎉 🙌 🍻

JoshMarler:
@nick-thompson, you the man.

Issue #51, opened 2020-05-24
Add Typescript definitions/support


JoshMarler:
It would be great to add TypeScript definitions/support to Blueprint. Strong types will be a familiar warm blanket for less JS savvy C++ devs.

Being able to define interfaces and the like for component properties will help with avoiding errors at compile time (View would certainly benefit from this addition). It's inherently a little more complex to debug type mismatches and referencing undefined  vars in Blueprint.

nick-thompson:
💯 love the idea! I think `View` is a great place to start, and after #44 and #43 go in, the `View` API should be pretty stable, so sounds like great timing

nick-thompson:
Marking this as closed after #99. We're now generally moving ahead with typescript support, and as the APIs are stabilizing so too will the type defs. Huge thanks to both @JoshMarler and @rbaxter08– I really appreciate it guys!

PR #67, opened 2020-06-26
Add Typescript support to juce-blueprint

rbaxter08:
#### PR 

Saw that you'd welcome typescript support here https://github.com/nick-thompson/blueprint/issues/51

I'm happy to keep it going and work on types for the `View` component and convert more files, just wanted to start with the config and make sure that everything still works..  Would also be open to adding `eslint` and `prettier` support per https://github.com/nick-thompson/blueprint/issues/24

Also happy to update the template and examples to TS eventually, but I'm a fan of small and specific PRs

I removed `babel`, I don't think it should be necessary anymore unless the project is using any experimental features, but  everything seems to work on my end. 

#### Unrelated

I really love the idea of this project.  I've had the idea of looking into VST development for a bit now, but as a react developer it's hard for me to get excited about building UIs in anything but react. :P

I'm curious if you have plans or thought about also building a bridge for the audio processing parts of JUCE?  That way you could make an entire VST with just react / JS and no C++ code.  Not sure how feasible that is, probably a lot of work too.


nick-thompson:
Thank you @rbaxter08 and welcome!

Yes this is definitely something I think we should get integrated and spread to the different View types, and I'd love your help with that. Also huge +1 for eslint and prettier, I haven't had time to get to those yet.

I have to admit though, I'm a total typescript newb, so as I'm trying to review this I'll probably be asking you to teach me many of these things :) My general view on adding typescript defs here is that (1) obviously this is a plus for typescript users and (2) because I suspect this project might attract people from the C++ world, having types extend into the js side of this project would be a big plus. But, as I'm not really a typescript user myself I don't know the right way to do it, and I think ideally we do it in a way that is trivially opt-in, which allows people who aren't typscript users to continue using whatever toolchain they prefer. So, my first few questions– does adding typescript defs to the js file itself make the file not compilable by something like babel (thinking about the template and the examples here)? If we use typescript to build the juce-blueprint module, but still aim to ship ES5 (i.e. when publishing to `npm` in the future), how can we enable typescript users to refer to those types?

> I really love the idea of this project. I've had the idea of looking into VST development for a bit now, but as a react developer it's hard for me to get excited about building UIs in anything but react. :P

I'm glad to hear that! And yes, I totally understand :) I came from the same place.

> I'm curious if you have plans or thought about also building a bridge for the audio processing parts of JUCE? That way you could make an entire VST with just react / JS and no C++ code. Not sure how feasible that is, probably a lot of work too.

Not quite, but this is something I've thought about a lot. I don't think JS is the right tool for that job, unless we're talking about precompiled dsp blocks that JS can assemble (like the Web Audio API, although I'm still not totally convinced by the web audio API). But I do look to tools like Faust and SOUL to facilitate a better environment for writing the DSP side of a VST, with hooks for loading a blueprint UI. For example, both Faust and SOUL have an option to "emit" a juce project. You could write DSP in one of those tools, emit a juce project, add the few lines to mount a `blueprint::ReactApplicationRoot` to the top of the GUI window, and carry the project the rest of the way in JS. To me that direction has more promise, and surely there are ways to make that whole pipeline smooth and easy to work with.

rbaxter08:
@nick-thompson 

>So, my first few questions– does adding typescript defs to the js file itself make the file not compilable by something like babel (thinking about the template and the examples here)? If we use typescript to build the juce-blueprint module, but still aim to ship ES5 (i.e. when publishing to npm in the future), how can we enable typescript users to refer to those types?

Yeah, great questions and actually you reminded me that I forgot to generate the type definition files 😅  I've added the missing config options to the `tsconfig.json`.

So, even when using typescript to build juce-blueprint, it's still outputting es5 js files, but it also outputs a corresponding type definition file (`*.d.ts`).  You can see the build artifacts here:

Screen Shot 2020-06-30 at 6 32 51 PM

So, if I'm a `js` user consuming `juce-blueprint` then the `d.ts` files are simply not used and I don't need to change anything to my config, which is exactly how the examples are working now. I haven't made any changes to those directories, but they still build and run:

Screen Shot 2020-06-30 at 5 56 07 PM


If I am a `ts` user, then whenever you import a file typescript will automatically look for the `.d.ts` file with the matching name, so you can actually just import that types like so:

Screen Shot 2020-06-30 at 5 02 08 PM
Screen Shot 2020-06-30 at 5 02 13 PM

Note that even though `CanvasProps` is defined in `Blueprint.d.ts` it is found when importing from `juce-blueprint` even though that lives in `Blueprint.js`

There is technically a slight draw back for `js` users in that they still have to download the type files even though they won't use them.   If that's a concern, you can break out the types and submit them to `https://github.com/DefinitelyTyped/DefinitelyTyped` where they can be installed separately via `npm add --save-dev @types/juce-blueprint`.  This is pretty common and actually are where the `react` types are stored (you can see I've added `@types/react` to the `package.json`.   Personally, if the bundle isn't _that_ much larger, I just like having the types included in the package I'm using, but 

One last aside, you can have tyepscript concat all the `.d.ts` files into a single `global.d.ts`, so you don't have so many new files in the build artifacts (this makes it easier for consumers to import the types), but I can't do that until I remove all the `.js` file from `juce-blueprint`.

packages/juce-blueprint/package.json
```diff
@@ -3,27 +3,22 @@
...
+    "@types/invariant": "^2.2.33",
+    "@types/react": "^16.9.41",
     "chalk": "^2.4.2",
-    "fs-extra": "^8.1.0"
+    "fs-extra": "^8.1.0",
+    "typescript": "^3.9.5"
   },
   "dependencies": {
     "core-js": "2.6.0",
+    "invariant": "^2.2.4",
```
rbaxter08, line 21: @nick-thompson 
this lib is used in the `BlueprintRenderer.js` https://github.com/nick-thompson/blueprint/blob/af0e1390c6cb0c0d05ba96abc8ba4a2031b60f95/packages/juce-blueprint/esm/lib/BlueprintRenderer.js#L5 and I'm actually sure where it was coming from originally.

nick-thompson:
@rbaxter08 awesome, thanks for such a detailed description. This sounds great. So basically, we write `juce-blueprint` in typescript and everybody can use it trivially and typescript users can use it trivially– awesome.

> There is technically a slight draw back for `js` users in that they still have to download the type files even though they won't use them. If that's a concern, you can break out the types and submit them to `https://github.com/DefinitelyTyped/DefinitelyTyped` where they can be installed separately via `npm add --save-dev @types/juce-blueprint`. This is pretty common and actually are where the `react` types are stored (you can see I've added `@types/react` to the `package.json`. Personally, if the bundle isn't _that_ much larger, I just like having the types included in the package I'm using, but

Gotchya. I'm with you– let's just have them included in the package.

Looking over the diff again real quick, this looks great. I want to just pull it down and get it running locally to confirm before I merge but I anticipate that'll be trivial, then I'll be back to merge it! And then yea, let's keep going on the other components if you're interested.

packages/juce-blueprint/package.json
nick-thompson: Ah right, I think that comes in in the react package, but I think it's good to have it here explicitly

nick-thompson:
Hey @rbaxter08 so I'm playing around with typescript right now finally, and I'm confused about one more piece here– TypeScript's export to the ES5 language feature set seems odd. For example, if I write `Array.from` and try to target ES5, I get an error about no such language feature rather than a polyfill or a codegen replacement. Maybe that's just not typescript's job, which is fine, but that makes me think that in the `juce-blueprint` package we'll need a `tsc` stage and then a `babel` stage in our `npm run build` so that we guarantee that the `juce-blueprint` library code is ES5 compatible (otherwise we're at risk of shipping code that just breaks Duktape). What do you think? Maybe I'm just missing something?

rbaxter08:
@nick-thompson Hmm, I hadn't looked into the ES restrictions of Duktape.  I think we could configure Typescript to throw errors on anything not supported by Duktape (basically adding or removing things from the `lib` array in the `tsconfig.json`), but that might be difficult considering Duktape only partially supports some versions.  

I think we can just have babel compile our typescript, so we can have all the latest ES stuff without having to worry about it.

Converting this to a draft PR for now, should be able to get the babel version up soon.

nick-thompson:
:+1: Yea I think running babel + Typescript gets us everything we might want while guaranteeing duktape can run it

rbaxter08:
@nick-thompson Got it building with babel.

I did just wanna note that it looks like babel isn't transpiling `Array.from`, I'm poking around a bit to see what's up, but this behavior does already exist in master.


Screen Shot 2020-07-09 at 3 35 58 PM

nick-thompson:
Thanks @rbaxter08 this looks great.

You're totally right– I don't think babel will transform `Array.from`, but I've found that with the `transform-runtime` plugin, babel will search for calls to `Array.from` and inject a `core-js` polyfill into the bundle when it finds one.

I used this to fix a bug in the example projects recently: https://github.com/nick-thompson/blueprint/commit/d68987d5bd7fc36f339e826de7b6c9efd53aeee5, but your comment made me realize we should add the same thing to the juce-blueprint `package.json` as well.

Sound good? If you're on board with that, I'll go ahead and merge and then follow up with a commit to add the runtime-transform. (Or feel free to add it here and I'll merge it all in at the same time!)

nick-thompson:
💥 thank you @rbaxter08 !

JoshMarler:
Awesome stuff @rbaxter08 ! I'm just converting a reasonably sized blueprint js project to Typescript. 

I wonder if you'd be kind enough to help me out. I'm hitting an issue using the `Canvas` component at the moment:

>TS2604: JSX element type 'Canvas' does not have any construct or call signatures.   

My usage of the component is as follows:

```
      <View onMouseDown={this._onMouseDown} onMeasure={this._onMeasure} {...other}>
        <Canvas autoclear={true} {...styles.canvas} onDraw={this._drawButton} />
        {this.props.children}
      </View>
```

Importing Canvas as:

```
import Blueprint, { View, Canvas } from 'juce-blueprint'
```

I'm googling around but the solutions I'm seeing for this issue don't seem to fit. I've used Typescript in the past but my knowledge is fairly minimal. I'd like to start strongly typing component props and the like to better document common components for the other devs on my team. 

Any help you can provide would be MOST appreciated. 

For reference my tsconfig currently looks as follows:

```
{
  "compilerOptions": {
    "sourceMap": true,
    "module": "commonjs",
    "moduleResolution": "classic",
    "allowJs": true,
    "noImplicitAny": false,
    "allowSyntheticDefaultImports": true,
    "strict": true,
    "strictNullChecks": true,
    "jsx": "react",
    "esModuleInterop": true,
    "removeComments": true
  },
  "include": [
    "./src/"
  ]

```

JoshMarler:
@rbaxter08, I maaay have fixed it. Using babel-loader now for typescript transpile stage and it appears to me working. Will reply back when I know all is sorted.

rbaxter08:
@JoshMarler I'll poke around and see if I can reproduce a similar error to the one you're getting, quick random thought, have you installed the `@types/react` package?

JoshMarler:
@rbaxter08. Yeah I've got all the type packages. I can get things to work if I just transpile js with the babel typescript preset. I'm having issues actually running the Typescript compiler though. Running some tests to see if this is even possible. 

I'm still hitting Blueprint.d.ts is not a module at the moment amongst other errors. Cool that I can run it with babel but obviously there's no actual type checking!

rbaxter08:
@JoshMarler I believe this PR should fix your issue

https://github.com/nick-thompson/blueprint/pull/86

JoshMarler:
@rbaxter08, you the man. I need to read more Typescript docs! 

With this and the @ts-ignore tags it might not look too horrible :-)

JoshMarler:
For reference. I got things working fine with `@ts-ignore`, only needed in a couple places as by duktape registered global `__MyAppNative__` is only referenced in a single class inside method implementations, so not other ts files use it. 

Looking like the optimum approach is to use a `tsc --noEmit` phase before running webpack build to go through `babel-loader` using `@babel/preset-typescript`. That avoids issues like the `Array.From` problem you identified @rbaxter08 . 

I just setup a couple scripts in my package.json:

`"type-check": "tsc --noEmit"`

`"build": "yarn run type-check && webpack --mode=production"`

All working just fine and dandy now. 

@rbaxter08 , I agree a TS project in the examples would be nice. I think it would be worth have both a pure js example and a ts example personally. @nick-thompson has been saying it'd be nice for us to have some more examples. Be a worthwhile addition.

JoshMarler:
@rbaxter08 , I'm still hitting issues with this. Currently testing combinations of awesome-typescript-loader and/or babel-loader with the typescript preset. I'm finding various issues at runtime like imports of enum types being undefined, createElement failures etc. I'm debating abandoning TS soon and just going with jsDoc as this is eating a lot of time, but if you happen to have a TS setup for a blueprint project which you've found works I'd be most grateful to see it!

rbaxter08:
Sorry it's giving you headaches @JoshMarler 

Here is the GainPlugin example with TS added that I've been using to test my work. 
https://github.com/rbaxter08/blueprint/pull/1

If this project is open source, I'd be happy to take a look for ya.

JoshMarler:
@rbaxter08 , I blame you trendy javascript kids! Thanks for the example! Sadly it's not open source but I'm going to knock together a smaller scale version of the project layout I have and go from there. I may request a quick bit of assistance if I can replicate the error in a smaller project! 

Your project setup looks entirely sane and pretty much identical to what I have, give or take a few extra plugins etc. There appears to be some issue importing my "components" package in my blueprint app. I'll keep digging! Will let you know if I find anything useful for future reference. 

Cheers

JoshMarler:
@rbaxter08 , thanks for the help. Got through it in the end. Couple issues. I wanted type-checking and transpile with babel to work as part of a webpack watch process. `awesome-typescript-loader` in babel mode did the trick. 

```
module.exports = {
  entry: {
    mixer_page:   './src/pages/mixer-page/index.tsx',
    kitedit_page: './src/pages/kit-edit-page/index.tsx'
  },
  output: {
    path: __dirname + './dist',
    filename: '[name].js',
    sourceMapFilename: "[file].map",
    devtoolModuleFilenameTemplate: 'webpack:///[absolute-resource-path]'
  },
  devtool: "source-map",
  resolve: {
    extensions: [".ts", ".tsx", ".js", ".jsx"]
  },
  module: {
    rules: [
      {
        test: /\.(ts|js)x?$/,
        exclude: /node_modules/,
        use: [
          { 
            loader: "awesome-typescript-loader",
            options: {
              useBabel: true,
              useTranspileModule: true,
              babelCore: '@babel/core'
            }
          }
        ]
      },
      {
        test: /\.svg$/,
        exclude: /node_modules/,
        use: ['svg-inline-loader']
      },
    ]
  },
  watchOptions: {
    ignored: /node_modules/
  },
}
```

For the moment I just import my "generic-components" directly from the components packages src/.tsx files in my top level app. To actually build my "generic-components" lib as it's own package and just use definition files is proving pretty hard. Various little issues. I'll make that a 2.0. 

I hit this for starters which was not pleasant: https://stackoverflow.com/questions/58053626/how-to-use-typescript-modules-with-duktape,

Might be that I need to load/evaluate both the transpiled genric-components bundle and my app bundle in Blueprint/duktape to get this all working but I've not tried loading multiple bundles yet. 

Will reply back if I get this TS component library structure working.

nick-thompson:
Oh man I missed some conversation here! Sorry for the absence.

> @rbaxter08 , I agree a TS project in the examples would be nice. I think it would be worth have both a pure js example and a ts example personally. @nick-thompson has been saying it'd be nice for us to have some more examples. Be a worthwhile addition.

Heh, I should have read here before writing my reply on the pull request. Pretty much wrote the same exact thing!

Also want to say that I really appreciate the conversation and work going on here. Getting through these issues as we make the transition is how we continue to make the project nice for first-time users, which I think is super important. Thanks for the help, and let me know how we can integrate these findings into the startup flow!

P.S. Blueprint just hit 200 github stars 🎉 that's pretty cool, I'm excited to see it continue growing. Cheers guys

Issue #66, opened 2020-06-22
Provide data/event flow examples and facilities


JoshMarler:
Currently `ReactApplicationRoot` and `EcmascriptEngine` support pushing events from native/C++ code into JS/React via the `dispatchEvent` mechanism. 

`dispatchEvent` is a function registered/created against the global `__BlueprintNative__` object by EventBridge.js. 

It seems that a lot of apps and client code would benefit from a generic "event based" framework or pattern to communicate between C++ and JS. 

For example, it would be nice to both send an `event` (consisting of an ID and payload/args) from C++ to Javascript and from Javascript to C++. 

I might wish subscribe to a "parameterChanged" event in JS and also trigger/signal that same event back to C++.

Is this something that blueprint could/should offer via `EcmascriptEngine`/`ReactApplicationRoot`?

I could imagine a complimentary function to `ReactApplicationRoot::dispatchEvent` along the lines of `ReactApplicationRoot::handleEvent`.

i.e.

```
virtual void ReactApplicationRoot::handleEvent(const juce::String& eventType, const juce::NativeFunctionArgs& args) = 0;
```

This might require some change or "fleshing out" of an `EventBridge`  to support a `publish` and `subscribe` like interface. 

@nick-thompson I'd love to hear your ideas around this area. I'm going to imnplement some sort of event based communication in my app over the coming days so will let you know if I come up with any interesting stuff.

JoshMarler:
@nick-thompson, in the middle of implementing two-way event dispatch at the moment as a test so I can open up a PR if you like the sound of this idea?

JoshMarler:
@nick-thompson, FYI I've got this working locally and so far seems pretty  good.

nick-thompson:
Hmmm a lot to think about here, thanks for starting the conversation @JoshMarler.

> The big initial question feels like "Should Blueprint provide this or should examples be given with data-flow patterns/implementations left to the users of Blueprint?"

I think that's exactly the right place to start. I suppose my initial stance is that Blueprint itself shouldn't take a stance on data flow patterns, because every app is different and everbody has their own opinions on data flow patterns. In React you see a ton of Flux/Redux, but there are plenty of people who prefer an RxJS style approach, for example. However, I think Blueprint _should_ provide _some_ set of facilities for easily implementing your data flow pattern of choice.

[...]

PR #68, opened 2020-06-29
Add support for two-way event dispatch

JoshMarler:
Allows pushing events from JS to C++ and vice versa

### DIFF
--- a/blueprint/core/blueprint_EcmascriptEngine.cpp
+++ b/blueprint/core/blueprint_EcmascriptEngine.cpp
@@ -119,12 +119,14 @@ namespace blueprint
             return true;
        }
 
+/*
        static void dumpContext(duk_context* ctx)
        {
            duk_push_context_dump(ctx);
            DBG(duk_to_string(ctx, -1));
            duk_pop(ctx);
        }
+*/
     }
 
     //==============================================================================
--- a/blueprint/core/blueprint_ReactApplicationRoot.h
+++ b/blueprint/core/blueprint_ReactApplicationRoot.h
@@ -393,6 +393,33 @@ namespace blueprint
         BundleEvalCallback afterBundleEval;
 
         //==============================================================================
+        using EventSubscriptionToken = int;
+
+        /**
+         * Subscribe to a given event from js. This should be called in your beforeBundleEval
+         * callback as the set of listeners/subscriptions will be cleared on bundle reloads.
+         * Returns an EventSubscriptionToken instance which callers may store and later use to unsubscribe()
+         * a previously registered event handler.
+         * Note, callers should take care to avoid callback loops between subscribe and dispatchEvent etc.
+         **/
+        EventSubscriptionToken subscribe(const juce::String &eventType, juce::var::NativeFunction handler)
+        {
+            JUCE_ASSERT_MESSAGE_THREAD
+            jassert(handler);
+
+            return engine.invoke("__BlueprintNative__.subscribe", eventType, std::move(handler));
+        }
+
+        /**
+         * Removes/Unsubscribes a previously registered event handler/callback.
+         */
+        void unsubscribe(const EventSubscriptionToken &token)
+        {
+            JUCE_ASSERT_MESSAGE_THREAD
+            engine.invoke("__BlueprintNative__.unsubscribe", token);
+        }
+
+        //==============================================================================
         /** The ReactApplicationRoot's engine instance. */
         EcmascriptEngine engine;
 
@@ -609,6 +636,7 @@ namespace blueprint
         std::unique_ptr<ViewManager>            viewManager;
         std::unique_ptr<BundleWatcher>          bundleWatcher;
         std::unique_ptr<juce::AttributedString> errorText;
+
         //==============================================================================
         JUCE_DECLARE_NON_COPYABLE_WITH_LEAK_DETECTOR (ReactApplicationRoot)
     };
--- a/packages/juce-blueprint/esm/lib/EventBridge.js
+++ b/packages/juce-blueprint/esm/lib/EventBridge.js
@@ -1,14 +1,34 @@
 import EventEmitter from 'events';
 
-
 const EventBridge = new EventEmitter();
 
 EventBridge.setMaxListeners(30);
+
+EventBridge.__listeners = {};
+EventBridge.__nextToken = 0;
 
 // An internal hook for the native side, from which we propagate events through
 // the EventEmitter interface.
 __BlueprintNative__.dispatchEvent = function dispatchEvent(eventType, ...args) {
   EventBridge.emit(eventType, ...args);
 }
+
+__BlueprintNative__.subscribe = function(eventType, handler) {
+    EventBridge.addListener(eventType, handler);
+
+    const token = EventBridge.__nextToken++;
+    EventBridge.__listeners[token] = {eventType, handler};
+    return token;
+};
+
+__BlueprintNative__.unsubscribe = function subscribe(token) {
+    if (!(token in EventBridge.__listeners)) {
+      console.log("Error: Attempt to remove non-existent EventBridge listener. Token invalid.");
+      return;
+    }
+
+    const { eventType, handler} = EventBridge.__listeners[token];
+    EventBridge.removeListener(eventType, handler);
+};
 
 export default EventBridge;


JoshMarler:
@nick-thompson, Opened this up to start a discussion around API for 2 way event dispatch. 

I'm using this fairly minimal implementation at the moment which is working just fine. I'd be interested to hear your thoughts about the API for this. 

Relates to https://github.com/nick-thompson/blueprint/issues/66

packages/juce-blueprint/esm/lib/EventBridge.js
```diff
@@ -1,14 +1,36 @@
...
+    }
+  }
+
+  // Override EventEmitter.emit()
+  emit(eventType, ...args) {
+    if (this.emittingFromNative) {
+      //TODO: For some reason duktape does not like us using
+      //      super.emit() here. This may be something to do with 
+      //      how Babel transpiles such a call. Investigate. 
+      EventEmitter.prototype.emit.call(this, eventType, ...args);
```
JoshMarler, line 27: This was rather annoying and might be a useful thing to note for yourself @nick-thompson.

blueprint/core/blueprint_EcmascriptEngine.cpp
```diff
@@ -119,12 +119,14 @@ namespace blueprint
             return true;
        }
 
+/*
        static void dumpContext(duk_context* ctx)
```
JoshMarler, line 123: One of our build variants shouted about an unused method. Could [maybe_unused] but I believe blueprint is currently using C++11 as minimum standard?

JoshMarler: Could just as easily remove the function etc.

JoshMarler:
@nick-thompson, This may or may not be of interest to you. Whilst I've been playing with this 2-way event dispatch pattern I've implemented some simple components (Slider, ToggleButton etc.). I've basically created simple/basic components which can then be wrapped by a higher level component which deals with binding to the EventBridge:

``` _emitEvent(event, ...args) {
        EventBridge.emit(event,  ...args);
    }

// In render
<ToggleButton 
     onToggled={(toggled) => { this._emitEvent(this.Events.INSTRUMENT_MUTE_CHANGED, toggled); }} 
     toggled={this.state.mute}
     text='M' 
     drawToggleButton={ LdmdLookAndFeel.drawToggleButton } 
     {...styles.mute_solo_button}>
</ToggleButton>
                  
```

JoshMarler:
@nick-thompson, Yep. Solved my performance issue. The state for the control was too high. Below approach works:

```
import React, { Component } from 'react';

import { View, EventBridge } from 'juce-blueprint';

import Slider from '../Slider';

class EventSlider extends Component {
  constructor(props) {
    super(props);

    this._onMeasure = this._onMeasure.bind(this);

    this.state = {
      width: 0,
      height: 0,
      sliderValue: 0.0
    };

    EventBridge.addListener(this.props.eventID, (value) => {
        this.setState({
            sliderValue: value
        });
    });
  }

  _onMeasure(width, height) {
    this.setState({
      width: width,
      height: height,
    });
  }

  _emitEvent(event, ...args) {
      EventBridge.emit(event,  ...args);
  }

  render() {
    const {eventID, sliderStyle, drawSlider, ...other} = this.props;

    return (
        <View {...other} onMeasure={this._onMeasure}>
              <Slider 
                onValueChanged={(value) => { this._emitEvent(eventID, value); }} 
                value={this.state.sliderValue}
                sliderStyle={sliderStyle} 
                drawSlider={drawSlider}>
              </Slider>
        </View>
    );
  }
}

export default EventSlider;
```

A worthwhile thing for Blueprint docs. State colocation is IMPORTANT. Suspect even more so with all the duktape binding.

nick-thompson:
Indeed, it looks like you just found your way to something of a React best practice :)

There are two things to note there in terms of performance: the first is that when a react component changes, its entire subtree undergoes the process of rendering a new tree, diffing with the old tree, and applying state updates. When this subtree is massive, that can be expensive. State colocation is one way to address this, and another way is to add `shouldComponentUpdate` hooks at the component level so that a component can mark itself (and its subtree) as not needing to be considered for this update: https://reactjs.org/docs/react-component.html#shouldcomponentupdate. This way you can still carry state relatively high up in the tree but only re-render the branches of the subtree that need it. The other thing you may be hitting is that during a slider drag, those value change callbacks fire _so_ fast. If you're using parameter listeners in particular, they also fire on the RT thread. My philosophy around this detail has been (1) always put this on the message thread and (2) throttle the events. The GUI, more often that not, doesn't actually need to react to _each_ of those slider change events, especially if they come in faster than ~60fps or ~30fps (depending on how reactive your GUI is). I try to keep my GUI out of the path of business logic... for example when a slider value changes, different parts of my app should be able to update without relying on event propagation through the GUI, and I'll shuffle events over to the GUI at ~30fps to make sure that what is drawn on screen reflects the up-to-date state values without flooding it. This is what the `ThrottleMap` class in the blueprint directory is for, although obviously I haven't yet included it in any example code!

I'll take a pass through the code here in just a sec

blueprint/core/blueprint_EcmascriptEngine.cpp
nick-thompson: Ohh man, that's a bummer. I think making blueprint available back to C++14/11 would maybe help user adoption but I _do_ often reach for C++17 features. What do you think is better here?

JoshMarler:
> Indeed, it looks like you just found your way to something of a React best practice :)
> 
> There are two things to note there in terms of performance: the first is that when a react component changes, its entire subtree undergoes the process of rendering a new tree, diffing with the old tree, and applying state updates. When this subtree is massive, that can be expensive. State colocation is one way to address this, and another way is to add `shouldComponentUpdate` hooks at the component level so that a component can mark itself (and its subtree) as not needing to be considered for this update: https://reactjs.org/docs/react-component.html#shouldcomponentupdate. This way you can still carry state relatively high up in the tree but only re-render the branches of the subtree that need it. The other thing you may be hitting is that during a slider drag, those value change callbacks fire _so_ fast. If you're using parameter listeners in particular, they also fire on the RT thread. My philosophy around this detail has been (1) always put this on the message thread and (2) throttle the events. The GUI, more often that not, doesn't actually need to react to _each_ of those slider change events, especially if they come in faster than ~60fps or ~30fps (depending on how reactive your GUI is). I try to keep my GUI out of the path of business logic... for example when a slider value changes, different parts of my app should be able to update without relying on event propagation through the GUI, and I'll shuffle events over to the GUI at ~30fps to make sure that what is drawn on screen reflects the up-to-date state values without flooding it. This is what the `ThrottleMap` class in the blueprint directory is for, although obviously I haven't yet included it in any example code!
> 
> I'll take a pass through the code here in just a sec

Indeed! (On all counts). I have relatively basic React knowledge and I think this little experiment has proved I should probably get reading up on a few more "first-principles" whilst I'm working on this stuff. 

I'm experimenting with some `ThrottleMap` approaches in my current app but the "colocation" approach has served me well here and works well in my current design I think.

JoshMarler:
>> Looking through the code I have one high level though– does it make sense to distinguish between who's emitting the event? That is, dispatchEventFromNative vs dispatchEventToNative?

My gut feeling is that EventBridge can, and maybe should be, totally dumb: just a plain old pubsub and you can subscribe from either side (native or js) and emit from either side. If you emit an event from native that you're also listening for on native, I think you should still get that callback. Then if the user wants this kind of distinguishing between native and js, they can enforce that by their own event naming conventions. For example, emitting from native, they might prefix all event names with "native:", and from js prefix all event names with "js:". Then as a subscriber you can choose if you want to respond to native or js or some of each. This feels a little more flexible to me for the end user. What do you think?

So to clarify, are we talking about removing the distinction between the two event sources in just EventBridge but leaving the C++ code in ReactApplicationRoot as is? I totally understand what you're getting at, I had similar thoughts myself. 

So are you suggesting something more along the lines of the below?

```
import EventEmitter from 'events';

class EventBridge extends EventEmitter {
  constructor() {
    super();

    this.emit = this.emit.bind(this);

    this.setMaxListeners(30);

    __BlueprintNative__.dispatchEventFromNative = (eventType, ...args) => {
      this.emit(eventType, ...args);
    }
  }

  // Override EventEmitter.emit()
  emit(eventType, ...args) {
      //TODO: For some reason duktape does not like us using
      //      super.emit() here. This may be something to do with 
      //      how Babel transpiles such a call. Investigate. 
      EventEmitter.prototype.emit.call(this, eventType, ...args);
      __BlueprintNative__.dispatchEventToNative(eventType, ...args);
  }
}

const _singletonInstance = new EventBridge();
export default _singletonInstance
```

In my specific use-case this would actually result in a circular callback loop, obviously I could fix that by tagging/prefixing events as you have suggested ("js: {event_name}", "native: {event_name}"). Something about this feels a little dirty but totally get your point. I'd be able to make the tag/prefix work for my use-case if we decide that is the best way to go

Would you mind giving a rough example of how you think the EventBridge should look for pub/sub to and from C++ ? I'm not totally sure I'm interpreting your ideas correctly!

nick-thompson:
Exactly, yea. That seems to be about all we would need from the javascript side. So to wire this up on the C++ side we would only really need the corresponding public ReactApplicationRoot API:

Dispatching an event from native:
```js
appRoot.dispatchEvent("native:ParameterValueChange", paramIndex, paramValue);
```

Receiving an event from native:
```js
appRoot.subscribe("js:ParameterValueChange", [this](juce::var::NativeFunctionArgs const& args) {
  // Do stuff
});
```

Then of course the implementation details involve marshaling to those `__BlueprintNative__` hooks that you showed above.

And for sure, if I omitted those prefixes, then when dispatching the event from native, that subscribe hook would get called. And if I were to update a value tree in that subscribe callback, then the value tree listener dispatches the event again, you'd get a loop.

To me the event name prefix doesn't seem so bad for addressing that? At least, the design of EventBridge this way seems to put the explicit control in the users hands (do some event name prefixing or manage your own flag to prevent a loop). I've been frustrated in the past when a library implements some implicit behavior that's hard to get around. That said, I'm totally open to discussion on this– I make no claim that my proposal is the best way to do this :)

Let me know what you think. Also, I'll be on vacation this week so apologies if my replies are slower than usual!

JoshMarler:
> Let me know what you think. Also, I'll be on vacation this week so apologies if my replies are slower than usual!

Unacceptable! I expect all our future recreational activities to be completely synchronised. 

Nope fair enough. I think I agree with you in regards to `EventBridge`. Question, would you prefer an API more like `appRoot.subscribe()` as you have demonstrated above?

At the moment I have the approach with an assignable `std::function` for a "HandleEvent" callback. The `subscribe` approach would involve storing/tracking listener functors in some sort of map I guess.
i.e.
 `std::map<juce::String, std::function<void(const juce::var::NativeFunctionArgs &args)>`

That's might be a nicer interface to be fair, less limiting. `dispatchEvent`, `subscribeEvent` ?

nick-thompson:
> Unacceptable! I expect all our future recreational activities to be completely synchronised.

😆 

> Nope fair enough. I think I agree with you in regards to `EventBridge`. Question, would you prefer an API more like `appRoot.subscribe()` as you have demonstrated above?
> 
> At the moment I have the approach with an assignable `std::function` for a "HandleEvent" callback. The `subscribe` approach would involve storing/tracking listener functors in some sort of map I guess.
> i.e.
> `std::map<juce::String, std::function<void(const juce::var::NativeFunctionArgs &args)>`
> 
> That might be a nicer interface to be fair, less limiting. `dispatchEvent`, `subscribeEvent` ?

Yea, that all sounds good to me

JoshMarler:
@nick-thompson, Apologies I've been distracted with other things over the last couple of weeks. I'll see if I can get this finished off tomorrow. Quick question, If we take your subscribe like API: 

```
appRoot.subscribe("js:ParameterValueChange", [this](juce::var::NativeFunctionArgs const& args) {
  // Do stuff
});
```

Do we want to provide an `unsubscribe` call ? We could have subscribe return some token/identifier to pass into `unsubscribe(subscriptionID` or similar?

nick-thompson:
> @nick-thompson, Apologies I've been distracted with other things over the last couple of weeks.

Oh no worries ever on that! I've been quite busy with other things as well anyway

> Do we want to provide an `unsubscribe` call ? We could have subscribe return some token/identifier to pass into `unsubscribe(subscriptionID)` or similar?

Yes, please! Great idea, and definitely something I overlooked above :)

nick-thompson:
Oh, also, I came up with something in one of my plugins last week that I think is significantly safer and more effective than the `ThrottleMap` idea above. Let's ignore the whole `ThrottleMap` thing as part of this PR; I'll push a commit shortly to demonstrate the better (in my opinion) way for event throttling, and I'll start a separate issue for discussion on that

nick-thompson:
Ok, see #78 for the discussion on `ThrottleMap`

JoshMarler:
@nick-thompson, hopefully things are more in line with your desired approach here: https://github.com/nick-thompson/blueprint/commit/b4c28db2a4b0bb9478607bb2e44f57724ee82cd2

In regards to the the throttle map, yeah nice one. I was actually going to mention the whole "Async message allocates" thing the other day. Using the atomic dirty flags seems like a good shout to me. Nice to have that in place.

JoshMarler:
Realised I wasn't clearing subscriptions on reload. Fixed here: https://github.com/nick-thompson/blueprint/pull/68/commits/23689303bfacd2a0ddba18010ea95b6e2ef789aa

nick-thompson:
> Ahh yeah. Looks like I removed this to get stuff compiling. Shall I just pull it out entirely or any desire to rethink the whole C++17 thing?

I'm torn on that 😞. Personally I would say let's just go C++17 and then it's a solved problem, but I wonder if that precludes people from using the library. What do you think? Honestly I think I'm leaning towards "let's just go C++17," and honestly I think we've already kind of stepped in that direction using `std::optional` in EcmascriptEngine, right?

Reviewing now!

blueprint/core/blueprint_ReactApplicationRoot.h
```diff
@@ -392,6 +393,59 @@ namespace blueprint
...
+
+            const size_t count = std::count_if(eventHandlerSubscriptions.begin(),
+                                               eventHandlerSubscriptions.end(),
+                                               [=] (const EventSubscription & sub)
+                                               {
+                                                   return sub.eventType == eventType;
+                                               });
+
+            EventSubscription sub = {eventType, count, std::move(handler)};
+            eventHandlerSubscriptions.push_back(sub);
```
nick-thompson, line 426: I'm wondering if a different data structure here would make the bookkeeping easier here. E.g. a `std::map<juce::Uuid, EventHandler>`?

```
EventToken subscribe(const juce::String& type, EventHandler && handler) {
    JUCE_ASSERT_MESSAGE_THREAD
    juce::Uuid token; // default construct
    eventSubscriptions[token] = std::move(handler):
    return token;
}
```

blueprint/core/blueprint_ReactApplicationRoot.h
```diff
@@ -402,6 +456,10 @@ namespace blueprint
         {
             JUCE_ASSERT_MESSAGE_THREAD
 
+            // Clear out previously registered subscriptions. Clients should
+            // re-register their event callbacks in beforeBundleEval.
+            eventHandlerSubscriptions.clear();
```
nick-thompson, line 461: :+1:

JoshMarler:
>  Honestly I think I'm leaning towards "let's just go C++17," and honestly I think we've already kind of stepped in that direction using std::optional in EcmascriptEngine, right?

Do we have std::optional!? I hadn't realised that

JoshMarler:
Thinking about it, I wonder if a move to C++17 makes sense if blueprint we're to adopt juce 6 or something? Brave new world?

blueprint/core/blueprint_ReactApplicationRoot.h
```diff
@@ -602,13 +660,54 @@ namespace blueprint
...
+
+                std::for_each(self->eventHandlerSubscriptions.begin(),
+                              self->eventHandlerSubscriptions.end(),
+                              [=] (const EventSubscription &sub)
+                              {
+                                  if (sub.eventType == eventType)
+                                  {
+                                      auto args = juce::var::NativeFunctionArgs(
+                                                      juce::var(),
+                                                      eventArgs.data(),
```
nick-thompson, line 690: I think we can skip the copy-to-vector step and just pull directly from the `args` at the top, right?

```
auto args = juce::var::NativeFunctionArgs(juce::var(), args.arguments + 1, args.numArguments - 1);
```

packages/juce-blueprint/esm/lib/EventBridge.js
nick-thompson: Hm... indeed, that's annoying!

nick-thompson:
> Thinking about it, I wonder if a move to C++17 makes sense if blueprint was to adopt juce 6 or something? Brave new world?

:D And I do think we want to adopt juce 6 because proper CMake support would make things like integrating custom Duktape builds way more straightforward.

Yea, shoot, let's go for it. C++17!

blueprint/core/blueprint_ReactApplicationRoot.h
JoshMarler: Yeah I debated whether we should have a map here. Happy to switch things around that way.

blueprint/core/blueprint_ReactApplicationRoot.h
JoshMarler: Yup. Good point. Ha, that was daft.

JoshMarler:
Yeah totally, very good point on the duktape configuration. I'll get that sorted then and update the example projects etc. I should really look at the juce 6 build system soon ... I've done quite a lot of CMake so I'd be happy to contribute a branch bringing that stuff in.

nick-thompson:
> Yeah totally, very good point on the duktape configuration. I'll get that sorted then and update the example projects etc. I should really look at the juce 6 build system soon ... I've done quite a lot of CMake so I'd be happy to contribute a branch bringing that stuff in.

That'd be so much appreciated. My CMake skills are weak :)

blueprint/core/blueprint_ReactApplicationRoot.h
JoshMarler: If we're storing a `juce::Uuid` token/key we probably still want a struct to store the `eventType` string and `EventHandler` together so we can call the correct handlers in our native `dispatchEventToNative` implementation?

JoshMarler: i.e. 

```
 using EventHandler = std::function<void(const juce::var::NativeFunctionArgs& args)>;
        using EventSubscriptionToken = juce::Uuid;

        struct EventSubscription
        {
            juce::String eventType;
            EventHandler handler;
        };

        /**
         * Subscribe to a given event from js. This should be called in your beforeBundleEval
         * callback as the set of listeners/subscriptions will be cleared on bundle reloads.
         * Returns an EventSubscriptionToken instance which callers may store and later use to unsubscribe()
         * a previously registered EventHandler.
         * Note, callers should take care to avoid callback loops between subscribe and dispatchEvent etc.
         **/
        EventSubscriptionToken subscribe(const juce::String &eventType, EventHandler &&handler)
        {
            JUCE_ASSERT_MESSAGE_THREAD
            jassert(handler);

            EventSubscriptionToken token;
            EventSubscription sub = {eventType, std::move(handler)};
            eventHandlerSubscriptions[token] = std::move(sub);

            return token;
        }
```

What dya think?

JoshMarler:
```
 void registerNativeEventHandlers()
        {
            engine.registerNativeMethod("__BlueprintNative__", "dispatchEventToNative", [](void* stash, const juce::var::NativeFunctionArgs& args) {
                auto self = reinterpret_cast<ReactApplicationRoot*>(stash);
                jassert (self != nullptr);
                jassert (args.numArguments >= 1);

                const juce::String eventType = args.arguments[0].toString();

                std::for_each(self->eventHandlerSubscriptions.begin(),
                              self->eventHandlerSubscriptions.end(),
                              [&] (const EventSubscription &sub)
                              {
                                  if (sub.eventType == eventType)
                                  {
                                      sub.handler(juce::var::NativeFunctionArgs(
                                                      juce::var(),
                                                      args.arguments + 1,
                                                      args.numArguments - 1));
                                  }
                              });

            }, (void *) this);
        }
```

For reference

JoshMarler:
> > Yeah totally, very good point on the duktape configuration. I'll get that sorted then and update the example projects etc. I should really look at the juce 6 build system soon ... I've done quite a lot of CMake so I'd be happy to contribute a branch bringing that stuff in.
> 
> That'd be so much appreciated. My CMake skills are weak :)

Yeah no worries. I'll drop a separate PR over after this with the migration to C++17 and then juce 6 can be a separate bit of work after that. Cool for me to just leave that function commented out for the moment and I'll [[maybe_unused]] in the C++17 PR ?

blueprint/core/blueprint_ReactApplicationRoot.h
nick-thompson: Oh, geez, of course. Good catch, sorry about that!

Actually, somehow reading that just made me realize there's an even easier way:

```js
const EventBridge = new EventEmitter();

EventBridge.setMaxListeners(30);

// An internal hook for the native side, from which we propagate events through
// the EventEmitter interface.
__BlueprintNative__.dispatchEvent = function dispatchEvent(eventType, ...args) {
  EventBridge.emit(eventType, ...args);
}

__BlueprintNative__.subscribe = function subscribe(eventType, callback) {
    EventBridge.addListener(eventType, callback);
});

__BlueprintNative__.unsubscribe = function subscribe(eventType, callback) {
    EventBridge.removeListener(eventType, callback);
});

export default EventBridge;
```

Then `ReactApplicationRoot` only needs this:

```cpp
void subscribe (const juce::String& eventType, EventHandler && handler) {
    engine.invoke("__BlueprintNative__.subscribe", handler);
});
```

Because of the lambdaReleasePool stuff, this should all be able to live just fine in the javascript side of things with way less code. I'm hand-waving a little bit over the unsubscribe part– I think we'll need return tokens on that side. I can elaborate there if it's helpful, but what do you think?

JoshMarler: Ugh yeah of course. I didn't even think about the fact that we could stash the lambda like that. That would be neat. 

So in `EventBridge.unsubscribe` we could just return a counter/token and store a ref to the callback function in JS ?

nick-thompson: Exactly!

nick-thompson:
> Yeah no worries. I'll drop a separate PR over after this with the migration to C++17 and then juce 6 can be a separate bit of work after that. Cool for me to just leave that function commented out for the moment and I'll [[maybe_unused]] in the C++17 PR ?

Yep!

blueprint/core/blueprint_ReactApplicationRoot.h
JoshMarler: Ha flipping eck. That is so much more elegant. I expect as I continue working with Blueprint I'll spot these patterns more often but this gets rid of all the `ToNative` `FromNative` as well. Jolly good catch sir.

JoshMarler: I'm trying to decide whether unsubscribe is actually essential ... it's the only bit of ugliness here. I suppose there could be valid reasons for dynamic subscribe/unsubscribe to the events, it's not something I need at the moment personally. What do you think? We probably should have it ...

JoshMarler: Listeners will be cleared on a bundle reload anyway as the release pool and ctx get cleared out

JoshMarler: It would be nice to be able to keep the expected signature:

```
__BlueprintNative__.unsubscribe = function unsubscribe(eventType, callback) {
    EventBridge.removeListener(eventType, callback);
});
```

I understand easily enough you could return an int counter or whatever from `subscribe` and change this signature to:

```
__BlueprintNative__.unsubscribe = function unsubscribe(token) {
    let listener = listeners[token];
    EventBridge.removeListener(listener.eventType, listener.callback);
});
```

Is there a better way to do this?

nick-thompson: Yea I think it's worth having an unsubscribe call. And I think we'll have to go the `token` approach because, at least now, passing the same lambda into `EcmascriptEngine::invoke` will just wrap it in a new function in the duktape context, so you won't be able to do `listener === cb` and get a `true` ever. I think what you wrote above there is plenty good.

```
EventBridge.__listeners = {};
EventBridge.__nextToken = 0;

__BlueprintNative__.subscribe = function(eventType, handler) {
    EventBridge.addListener(eventType, handler);

    const token = EventBridge.__nextToken++;
    EventBridge.__listeners[token] = handler;
    return token;
};
```
Then use exactly your `unsubscribe` up there.

JoshMarler: Yeah cool. I was musing over whether there was any smart way to pull the relevant lambda out of the duk stack etc but that sounds like either voodoo or an antipattern. Cool. Will get this tidied up when I'm back at my desk later. Cheers

nick-thompson: Awesome, thanks. Yea, I'm sure there's a way to pull it out of the duk stack but I feel like it would be way more complicated than the token pattern here

JoshMarler: @nick-thompson, OK hopefully the PR now fits our new expectations. Let me know if that's all good and I can squash it all into a single commit.

packages/juce-blueprint/esm/lib/EventBridge.js
```diff
@@ -1,14 +1,34 @@
...
+    return token;
+};
+
+__BlueprintNative__.unsubscribe = function subscribe(token) {
+    if (!(token in EventBridge.__listeners)) {
+      console.log("Error: Attempt to remove non-existent EventBridge listener. Token invalid.");
+      return;
+    }
+
+    const { eventType, handler} = EventBridge.__listeners;
```
nick-thompson, line 30: Little typo here I think?

```
    const { eventType, handler} = EventBridge.__listeners[token];
```

nick-thompson:
Perfect, @JoshMarler, thanks for seeing this through. One little typo in there while you're squashing, then happy to merge ASAP

packages/juce-blueprint/esm/lib/EventBridge.js
JoshMarler: That's really quite odd as I tested it and it was sane .... I'm hoping that happened by mistake after I built! Will fix and squash.

JoshMarler:
@nick-thompson, OK all squashed and corrected!

blueprint/core/blueprint_ReactApplicationRoot.h
```diff
@@ -392,6 +392,34 @@ namespace blueprint
...
+         * Returns an EventSubscriptionToken instance which callers may store and later use to unsubscribe()
+         * a previously registered event handler.
+         * Note, callers should take care to avoid callback loops between subscribe and dispatchEvent etc.
+         **/
+        EventSubscriptionToken subscribe(const juce::String &eventType, juce::var::NativeFunction handler)
+        {
+            JUCE_ASSERT_MESSAGE_THREAD
+            jassert(handler);
+
+            engine.invoke("__BlueprintNative__.subscribe", eventType, std::move(handler));
```
nick-thompson, line 410: Oh, oops, I missed this: shouldn't this be `return engine.invoke("__BlueprintNative__.subscribe", eventType, std::move(handler));` Right now this will just drop javascript's understanding of the token and give out a 0 every time

JoshMarler: Arrrg sorry! I altered that earlier whilst testing something. Good job you're on the ball Nick. I'm clearly a bit sleepy today. Will fix.

JoshMarler: OK resolved and tested it actually works again.

nick-thompson:
🎉  Thanks Josh!

PR #78, opened 2020-07-19
Update the GainPlugin example parameter throttling

nick-thompson:
This commit introduces a safer way of broadcasting parameter value updates to the javascript engine.

The way it was previously written, we had the `juce::MessageManager::callAsync` call in the parameter change callback. Although in practice that's not a huge deal, it is technically non-deterministic, and is running _way_ frequently on the realtime thread when a parameter is actually changing.

The way I had been planning to rewrite this involved the [ThrottleMap](https://github.com/nick-thompson/blueprint/blob/master/blueprint/core/blueprint_ThrottleMap.h) abstraction, but actually this also isn't a great solution for the realtime thread because of the syscall to collect the current time and the allocation upon inserting into the `std::map`.

This new way that I'm showing here I think is the safest and most efficient, and width this I'll propose that we remove the ThrottleMap altogether.

### DIFF
--- a/examples/GainPlugin/Source/PluginEditor.cpp
+++ b/examples/GainPlugin/Source/PluginEditor.cpp
@@ -71,15 +71,25 @@ GainPluginAudioProcessorEditor::GainPluginAudioProcessorEditor (GainPluginAudioP
     addAndMakeVisible(appRoot);
     appRoot.evaluate(bundle);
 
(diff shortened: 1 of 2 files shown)


JoshMarler:
Looks totally sane to me Nick. Good to avoid any potential locking here

nick-thompson:
:+1: awesome, thanks for the look over!

Issue #36, opened 2020-03-27
Reference to ‘AudioBuffer’ is Ambiguous


zigen:
failed to build examples/BlueprintPlguin AU Plugin with XCode 11.4(11E146) on macOS 10.15.14 (Catalina)

the error shows 'Reference to ‘AudioBuffer’ is Ambiguous'.
JUCE causes this error on XCode11.4 and [already solved](https://forum.juce.com/t/juce-is-busted-in-xcode-11-4/38249/5)  in JUCE's `develop` branch. [the commit is here](https://github.com/WeAreROLI/JUCE/commit/dddeb1ad68f8539f5ea2fbac36c07532f12bf9cf)

## work around
just checkout `ext/juce` to `develop` from `5.4.1` (because the commit dddeb1 has not released yet)

## solution
when JUCE released new version which includes the commit, we need to update submodule `ext/juce` to latest

nicochatzi:
The fix is in `6.0.0`. @nick-thompson are you ok with bumping a major version to close this issue?

nick-thompson:
Yep, definitely. In conversation with @JoshMarler over the course of a few different PRs we're thinking that moving to JUCE6, C++17, and CMake will generally clean things up quite a bit.

If it's a seamless update from where we are now to 6.0.0 then we can go ahead and do it now. If it's messy, I wonder if it might be worth it to bite off the whole move to CMake at the same time.

@nicochatzi have you tried building the GainPlugin example after updating `ext/juce` to 6.0.0? Was it easy?

emurray2:
I got the same error when I tried to build the GainPlugin example with Xcode 11.6(11E708) on macOS Catalina 10.15.6 . I replaced the existing folders inside `ext/juce` with the folders from my current JUCE directory. I have JUCE 6.0.1, and I was able to compile and run GainPlugin flawlessly!

Thank you all for the suggestions! I can't wait to learn more about this project and possibly contribute a bit once I figure things out.

nick-thompson:
:+1: thanks! Definitely, we'll be making the update to juce6 asap.

@emurray2 I'd be happy to get you involved any time!

emurray2:
Sounds good! Awesome, I'll keep you updated and take a look at the project tracker to see what I can work on.

nick-thompson:
Done in 30fb81bc91cb5d9d2e0c2cc0319043042b57d67b!

PR #81, opened 2020-07-31
Fix performance issues around the LambdaReleasePool in EcmascriptEngine.

### LINKED ISSUE #79: Hot Reload Hang on MacOS
I'm finding that during a hot reload on MacOS, Xcode just hangs in a permanent loop and the project freezes.

Dropping the debugger in shows that we're stuck in the Duktape finalizer callstack searching through the `lambdaReleasePool`, but it seems like there are a million things to be finalized. It looks like there indeed might be a bug with the lambdaReleasePool, because it's huge, but I'm not sure that's exactly what's hanging up Xcode here.

@JoshMarler have you seen this one?

JoshMarler:
This fix resolves #79. The freeze was caused by an O(Nsquared) lookup
 on the LambdaReleasePool vector when GC kicked in and triggered our
 LambdaReleasePool finalizers.

 Note this also bumps juce from 5.4.1 to 5.4.2 as std::hash
 specialisation for juce::Uuid was added to the Uuid header in 5.4.2.

### DIFF
--- a/blueprint/core/blueprint_EcmascriptEngine.h
+++ b/blueprint/core/blueprint_EcmascriptEngine.h
@@ -9,6 +9,7 @@
 
 #pragma once
 
+#include <unordered_map>
 
 namespace blueprint
 {
@@ -153,11 +154,12 @@ namespace blueprint
             static duk_ret_t callbackFinalizer (duk_context* ctx);
 
             juce::var::NativeFunction callback;
+            juce::Uuid                id;
         };
 
         //==============================================================================
         duk_context* ctx;
-        std::vector<std::unique_ptr<LambdaHelper>> lambdaReleasePool;
+        std::unordered_map<juce::Uuid, std::unique_ptr<LambdaHelper>> lambdaReleasePool;
 
         //==============================================================================
         void removeLambdaHelper (LambdaHelper* helper);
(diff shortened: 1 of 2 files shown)


blueprint/core/blueprint_EcmascriptEngine.h
```diff
@@ -153,11 +154,12 @@ namespace blueprint
             static duk_ret_t callbackFinalizer (duk_context* ctx);
 
             juce::var::NativeFunction callback;
+            juce::Uuid                id;
         };
 
         //==============================================================================
         duk_context* ctx;
-        std::vector<std::unique_ptr<LambdaHelper>> lambdaReleasePool;
+        std::unordered_map<juce::Uuid, std::unique_ptr<LambdaHelper>> lambdaReleasePool;
```
JoshMarler, line 162: @nick-thompson, went for `unordered_map` in the end as it's easier to use with `unique_ptr`. I decided against a static counter in `LambdaHelper` as the key for the moment, could be inflexible in the future with multiple `EcmascriptEngine` instances? i.e. if you have multiple ReactApplicationRoot instances in your app etc.  

Had to bump juce to 5.4.2 to include the `std::hash` specialisation for `juce::Uuid`. Could use something else if necessary but I think this is OK ?

nick-thompson: Nice, I like it!

Yep, happy for the bump to 5.4.2. I want to bump to 6.x soon (#36) with CMake soon anyway. That should also make #22 quite easy. Much to do! :)

Issue #69, opened 2020-06-30
Arrow function class props (Can we use babel-plugin-transform-class-properties). 


JoshMarler:
There is a lot of manual binding of `this` in Blueprint component constructor examples at the moment. Is there some reason for not using `babel-plugin-transform-class-properties` ? i.e. does duktape object?

Would this be a worthwhile addition to allow arrow function properties on components and avoid some boiler plate if desired?

nick-thompson:
Yea, definitely can add this! Honestly I just haven't tried it yet, but I bet via babel we'd get JS that duktape is fine with

JoshMarler:
Cool Cool!

I'll give it a go today and let you know. Can drop a PR adding it to the templates etc.

JoshMarler:
@nick-thompson, just switched this on in the gain example and all looks good. Nice to get away from all that manual binding. What would you prefer in the example code though? 

I can change all the components in the `GainPlugin` example to use the arrow function approach if desired. We're on babel 7 so it's the "plugin-proposal-class-properties" package. I can add it to the template etc.

JoshMarler:
I'm still learning JS ins and outs to some extent so I'm not sure what is preferred here. 

https://medium.com/@charpeni/arrow-functions-in-class-properties-might-not-be-as-great-as-we-think-3b3551c440b1

Are arrow function members worth it!?

JoshMarler:
The plugin is still probably useful for other purposes I guess.

nick-thompson:
Thanks @JoshMarler. I just merged the PR, I think it's a useful add. I like the arrow function class members because the manual binding is obviously a pain.

I don't put much stock in that article for two reasons: (1) mockability and inheritance are both big players only in the classic OO world. Personally I try to avoid both of them as much as I can in my code and lean towards a more functional approach, which either mitigates or solves both of these problems inherently. (2) Performance is a big one, but Duktape doesn't even have a JIT, so there is no notion of the hit counter + fast path to begin with. If we integrate JavascriptCore or V8 it might be worth measuring this and avoiding it in `juce-blueprint`, but for client code we'll just leave it up to the user to tune their own performance.

I think we can go ahead and close this issue now– I don't think we need to make a point of going back to update the example code, we can just make incremental updates as we're working in there on other tasks. Sound good?

JoshMarler:
Sounds good to me @nick-thompson ! Cheers (thanks for the useful know how)

PR #82, opened 2020-07-31
Add babel plugin-proposal-class-properties

JoshMarler:
I've not added this to the juce-blueprint packages dependencies yet @nick-thompson. I've added it for the purposes of client code. Let me know if it's worth having a the top level for future development. 

Relates to https://github.com/nick-thompson/blueprint/issues/69


nick-thompson:
:+1: I think client code is the right spot for now. This is great, I'll go ahead and merge!

Issue #13, opened 2019-06-26
Camelcase to hyphen?


nick-thompson:
I'm using hyphenated property names (mostly because Yoga uses hyphenated property names and I didn't want to deal with any mapping), which can be annoying when writing a javascript object literal because you have to quote the key names.

I'm not totally sure I want to add this feature, but it might be a nice to have: some camelCase to hyphenated library so that you can write js object literals without quotes and the renderer will map to hyphenated before pushing to the native code.

nick-thompson:
Adding this to beta, will close momentarily with #190. Thanks @olidacombe !

PR #83, opened 2020-08-02
Improve ShadowView layout performance.

JoshMarler:
This commit removes unnecessary root-layout computations in response
  to View/Component property changes and statically initialises
  ShadowView's property identifiers to improve comparision
  performance.

@nick-thompson, did a quick bit of grunt work to improve `juce::Identifier` performance and remove unnecessary calls to `performRootShadowTreeLayout`. This is an initial attempt to address #4 and #5. We can discuss the TODO comments around your `juce::Async` updater idea here and also my point in regards to possibly making `ShadowView::setProperty` return a bool if we wished to further reduce identifier comparisions. 

If you're happy with the overall approach I can add similar static initialisation of `juce::Identifier` properties in the other core component types in a separate PR.

### DIFF
--- a/blueprint/core/blueprint_ShadowView.cpp
+++ b/blueprint/core/blueprint_ShadowView.cpp
@@ -83,6 +83,111 @@ namespace blueprint
 
     namespace
     {
+        //==============================================================================
+        struct FlexProperties
+        {
+            // Flex enums
+            juce::Identifier direction       = "direction";
+            juce::Identifier flexDirection   = "flex-direction";
+            juce::Identifier justifyContent  = "justify-content";
+            juce::Identifier alignItems      = "align-items";
+            juce::Identifier alignContent    = "align-content";
+            juce::Identifier alignSelf       = "align-self";
+            juce::Identifier position        = "position";
+            juce::Identifier flexWrap        = "flex-wrap";
+            juce::Identifier overflow        = "overflow";
+
+            // Flex dimensions
+            juce::Identifier flex            = "flex";
+            juce::Identifier flexGrow        = "flex-grow";
+            juce::Identifier flexShrink      = "flex-shrink";
+            juce::Identifier flexBasis       = "flex-basis";
+            juce::Identifier width           = "width";
+            juce::Identifier height          = "height";
+            juce::Identifier minWidth        = "min-width";
+            juce::Identifier minHeight       = "min-height";
+            juce::Identifier maxWidth        = "max-width";
+            juce::Identifier maxHeight       = "max-height";
+            juce::Identifier aspectRatio     = "aspect-ratio";
+
+            // Margin
+            juce::Identifier marginMetaProp   = "margin";
+            juce::Identifier marginLeft       = juce::String("margin-") + YGEdgeToString(YGEdgeLeft);
+            juce::Identifier marginRight      = juce::String("margin-") + YGEdgeToString(YGEdgeRight);
+            juce::Identifier marginTop        = juce::String("margin-") + YGEdgeToString(YGEdgeTop);
+            juce::Identifier marginBottom     = juce::String("margin-") + YGEdgeToString(YGEdgeBottom);
+            juce::Identifier marginStart      = juce::String("margin-") + YGEdgeToString(YGEdgeStart);
+            juce::Identifier marginEnd        = juce::String("margin-") + YGEdgeToString(YGEdgeEnd);
+            juce::Identifier marginHorizontal = juce::String("margin-") + YGEdgeToString(YGEdgeHorizontal);
+            juce::Identifier marginVertical   = juce::String("margin-") + YGEdgeToString(YGEdgeVertical);
+
+            // Padding
+            juce::Identifier paddingMetaProp   = "padding";
+            juce::Identifier paddingLeft       = juce::String("padding-") + YGEdgeToString(YGEdgeLeft);
+            juce::Identifier paddingRight      = juce::String("padding-") + YGEdgeToString(YGEdgeRight);
+            juce::Identifier paddingTop        = juce::String("padding-") + YGEdgeToString(YGEdgeTop);
+            juce::Identifier paddingBottom     = juce::String("padding-") + YGEdgeToString(YGEdgeBottom);
+            juce::Identifier paddingStart      = juce::String("padding-") + YGEdgeToString(YGEdgeStart);
+            juce::Identifier paddingEnd        = juce::String("padding-") + YGEdgeToString(YGEdgeEnd);
+            juce::Identifier paddingHorizontal = juce::String("padding-") + YGEdgeToString(YGEdgeHorizontal);
+            juce::Identifier paddingVertical   = juce::String("padding-") + YGEdgeToString(YGEdgeVertical);
+
+            //Position
+            juce::Identifier positionLeft      = YGEdgeToString(YGEdgeLeft);
+            juce::Identifier positionRight     = YGEdgeToString(YGEdgeRight);
+            juce::Identifier positionTop       = YGEdgeToString(YGEdgeTop);
+            juce::Identifier positionBottom    = YGEdgeToString(YGEdgeBottom);
+            //TODO: Any need for other edge values? i.e. start, end ?
+        } flexProperties;
+
+        bool isMarginProp(const juce::Identifier& prop)
+        {
+            static const std::array<juce::Identifier, 9> marginProps =
+            {
+                flexProperties.marginMetaProp,
+                flexProperties.marginLeft,
+                flexProperties.marginRight,
+                flexProperties.marginTop,
+                flexProperties.marginBottom,
+                flexProperties.marginStart,
+                flexProperties.marginEnd,
+                flexProperties.marginHorizontal,
+                flexProperties.marginVertical
+            };
+
+            return std::find(marginProps.cbegin(), marginProps.cend(), prop) != marginProps.cend();
+        }
+
+        bool isPaddingProp(const juce::Identifier& prop)
+        {
+            static std::array<juce::Identifier, 9> paddingProps =
+            {
+                flexProperties.paddingMetaProp,
+                flexProperties.paddingLeft,
+                flexProperties.paddingRight,
+                flexProperties.paddingTop,
+                flexProperties.paddingBottom,
+                flexProperties.paddingStart,
+                flexProperties.paddingEnd,
+                flexProperties.paddingHorizontal,
+                flexProperties.paddingVertical
+            };
+
+            return std::find(paddingProps.cbegin(), paddingProps.cend(), prop) != paddingProps.cend();
+        }
+
+        bool isPositionProp(const juce::Identifier& prop)
+        {
+            static std::array<juce::Identifier, 4> positionProps =
+            {
+                flexProperties.positionLeft,
+                flexProperties.positionRight,
+                flexProperties.positionTop,
+                flexProperties.positionBottom,
+            };
+
+            return std::find(positionProps.cbegin(), positionProps.cend(), prop) != positionProps.cend();
+        }
 
         //==============================================================================
         std::map<juce::String, YGDirection> ValidDirectionValues {
@@ -150,17 +255,51 @@ namespace blueprint
         template<typename T>
         bool validateFlexProperty (juce::String value, std::map<juce::String, T> validValues)
         {
-            for (const auto& [flexValue, enumValue] : validValues)
+            return std::any_of(validValues.cbegin(), validValues.cend(), [=] (const auto &pair)
             {
-                if (value.equalsIgnoreCase(flexValue))
-                {
-                    return true;
-                }
-            }
-
-            return false;
+                return value.equalsIgnoreCase(pair.first);
+            });
         }
+    }
 
+    //==============================================================================
+    bool ShadowView::isLayoutProperty(const juce::Identifier& property)
+    {
+        static const std::array<juce::Identifier, 21> properties =
+        {
+            flexProperties.direction,
+            flexProperties.flexDirection,
+            flexProperties.justifyContent,
+            flexProperties.alignItems,
+            flexProperties.alignContent,
+            flexProperties.alignSelf,
+            flexProperties.position,
+            flexProperties.flexWrap,
+            flexProperties.overflow,
+            flexProperties.flex,
+            flexProperties.flexGrow,
+            flexProperties.flexShrink,
+            flexProperties.flexBasis,
+            flexProperties.width,
+            flexProperties.height,
+            flexProperties.minWidth,
+            flexProperties.minHeight,
+            flexProperties.minWidth,
+            flexProperties.maxWidth,
+            flexProperties.maxHeight,
+            flexProperties.aspectRatio
+        };
+
+        if (std::find(properties.cbegin(), properties.cend(), property) != properties.cend())
+            return true;
+        else if (isMarginProp(property))
+            return true;
+        else if (isPaddingProp(property))
+            return true;
+        else if (isPositionProp(property))
+            return true;
+        else
+            return false;
     }
 
     //==============================================================================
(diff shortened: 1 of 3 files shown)


blueprint/core/blueprint_ShadowView.cpp
```diff
@@ -83,6 +83,111 @@ namespace blueprint
...
+            juce::Identifier paddingEnd        = juce::String("padding-") + YGEdgeToString(YGEdgeEnd);
+            juce::Identifier paddingHorizontal = juce::String("padding-") + YGEdgeToString(YGEdgeHorizontal);
+            juce::Identifier paddingVertical   = juce::String("padding-") + YGEdgeToString(YGEdgeVertical);
+
+            //Position
+            juce::Identifier positionLeft      = YGEdgeToString(YGEdgeLeft);
+            juce::Identifier positionRight     = YGEdgeToString(YGEdgeRight);
+            juce::Identifier positionTop       = YGEdgeToString(YGEdgeTop);
+            juce::Identifier positionBottom    = YGEdgeToString(YGEdgeBottom);
+            //TODO: Any need for other edge values? i.e. start, end ?
```
JoshMarler, line 140: I don't think the other values from `ValidEdges` are actually needed here?

blueprint/core/blueprint_ShadowView.cpp
```diff
@@ -150,17 +255,51 @@ namespace blueprint
         template<typename T>
         bool validateFlexProperty (juce::String value, std::map<juce::String, T> validValues)
         {
-            for (const auto& [flexValue, enumValue] : validValues)
+            return std::any_of(validValues.cbegin(), validValues.cend(), [=] (const auto &pair)
```
JoshMarler, line 258: clang-tidy was yelling at me about this. Avoids potential of an unused var in the structured binding and also allows compiler stl loopy unroll shenanigans I guess.

blueprint/core/blueprint_ShadowView.cpp
```diff
@@ -170,122 +309,157 @@ namespace blueprint
...
-        juce::Identifier marginMetaProp ("margin");
-
-        for (const auto& [edgeName, enumValue] : ValidEdgeValues)
+        if (isMarginProp(name))
         {
-            juce::Identifier propId (juce::String("margin-") + edgeName);
+            if (name == flexProperties.marginMetaProp)
+                BP_SET_FLEX_DIMENSION_PROPERTY_AUTO(newValue, YGNodeStyleSetMargin, yogaNode, YGEdgeAll)
+            else
+                BP_SET_FLEX_DIMENSION_PROPERTY_AUTO(newValue, YGNodeStyleSetMargin, yogaNode, ValidEdgeValues[name.toString().replace("margin-", "")])
```
JoshMarler, line 440: OK with this removal of the  "margin-" prefix here?

blueprint/core/blueprint_ViewManager.cpp
```diff
@@ -52,14 +52,22 @@ namespace blueprint
...
+        if (ShadowView::isLayoutProperty(name))
+        {
+            //TODO: Here is where we could use juce::AsyncUpdater to coalesce layout property
+            //      updates into a single root layout operation.
+            shadow->setProperty(name, value);
+            performRootShadowTreeLayout();
+        }
+        else
+        {
+            view->setProperty(name, value);
```
JoshMarler, line 68: Is there a need for `View` to hold the `ShadowView` properties also?

blueprint/core/blueprint_ShadowView.cpp
nick-thompson: :+1: looks good. Nitpicky, but while we're here let's capture `value` by ref and avoid the copy?

blueprint/core/blueprint_ShadowView.cpp
nick-thompson: Agreed, and I think we can remove the `start` and `end` props from padding and margin as well tbh. They're properties that aren't supported in standard CSS so it seems sensible to ignore them here

blueprint/core/blueprint_ShadowView.cpp
nick-thompson: Pretty sure we don't need that `toString().replace` at all? `name == flexProperties.marginMetaProp` which doesn't include a `"margin-"`. Or I'm missing something :)

blueprint/core/blueprint_ShadowView.cpp
```diff
@@ -170,122 +309,157 @@ namespace blueprint
...
-        juce::Identifier paddingMetaProp ("padding");
-
-        for (const auto& [edgeName, enumValue] : ValidEdgeValues)
+        if (isPaddingProp(name))
         {
-            juce::Identifier propId (juce::String("padding-") + edgeName);
+            if (name == flexProperties.paddingMetaProp)
+                BP_SET_FLEX_DIMENSION_PROPERTY(newValue, YGNodeStyleSetPadding, yogaNode, YGEdgeAll)
+            else
+                BP_SET_FLEX_DIMENSION_PROPERTY(newValue, YGNodeStyleSetMargin, yogaNode, ValidEdgeValues[name.toString().replace("padding-", "")])
```
nick-thompson, line 452: Same here, can omit the `toString().replace()` I think

blueprint/core/blueprint_ViewManager.cpp
```diff
@@ -52,14 +52,22 @@ namespace blueprint
...
 
-        view->setProperty(name, value);
-        shadow->setProperty(name, value);
-
-        // For now, we just assume that any new property update means we
-        // need to redraw or lay out our tree again. This is an easy future
-        // optimization.
-        performRootShadowTreeLayout();
-        view->repaint();
+        //TODO: There is an element of duplicate work checking ShadowView::isLayoutProperty and
```
nick-thompson, line 55: Good point... I wonder if there's a better way?

It's been a minute since I've been in this part of the codebase, but I wonder if we could have a `std::map<juce::Identifier, YogaFlexEnumValue>` so that we could have a constant lookup time to map the property name to a Yoga enum. If the map doesn't contain the identifier, we know it's not a layout prop. That, I think, reduces 2N comparisons down to the cost of the map lookup.

As for maybe returning a bool, I think we can actually do a couple more things here. So a `ShadowView` is in a lot of ways just a C++ wrapper around a YogaNode. That YogaNode has its own flag for indicating whether it's dirty or not (i.e. has a property changed since the last time we computed a layout). So (1) we could add a method on ShadowView to query the dirty state of that internal yoga node. Then (2) the way I've been doing this so far is if I set _any_ property, I redo a layout from the root of the application. Obviously we're adding the check here for `isLayoutProperty`, but it's still the root tree node. Yoga's fast, and caches well, so that's not really a problem, but in truth we often only need to do a layout from some node deeper in the tree, and only considering a much smaller subtree.

I'm thinking that after React is done applying all property updates for a given batch, we run through the whole list of shadowViews with a "if you're dirty, redo your layout for your subtree." This way the only penalty we pay is if we layout a particular subtree and then layout a parent of that subtree, but here we still get yoga's caching.

If we can do that, we can avoid this whole idea about AsyncUpdater and coalescing commits... the trick will be getting React to tell us it's done with the given batch.

What do you think? Want to try for it?

blueprint/core/blueprint_ViewManager.cpp
nick-thompson: Hmm... off the top of my head I can't think of any need. I have this feeling that it might come in handy for a `View` to know about all of its ShadowView's properties and vice versa, but the only things I can think of off hand are contrived examples like "what if I want a View that paints itself red when it's flex-direction: column," which is so obviously contrived and in reality that would just be something done at the js level in user code anyway lol

I'm cool to go ahead with this like this

blueprint/core/blueprint_ShadowView.cpp
JoshMarler: Yep will update.

JoshMarler:
@nick-thompson. Gosh. Really need to get back to this and actually finish it at some point. Many other things have taken precedence. Apologies for letting this dangle for so long! Would be be nice to get these little performance hacks in.

nick-thompson:
:D No worries at all

blueprint/core/blueprint_ViewManager.cpp
olidacombe: Hi,

Thinking about what @nick-thompson suggested, using `std::map<juce::Identifier, YogaFlexEnumValue>`, and looking at the property validations / setters, does it not make sense to create a map to individual setter functions?  Then you get faster than current linear-time property setter lookups, with validation (which looks essentially like property name to setter routing) for free via std::map bounds checking.

Something like https://github.com/JoshMarler/blueprint/compare/improve_shadowview_performance...olidacombe:improve_shadowview_performance__pre_master_merge_compare 

Note the working branch is https://github.com/olidacombe/blueprint/tree/improve_shadowview_performance but that has master merged into it so the diff was pretty mental, so I made the branch above to be a readable set of changes.

I encapsulated the map into the class `PropertySetterMap` because I have designs on solving https://github.com/nick-thompson/blueprint/issues/13 as trivial side effect (in the `PropertySetterMap` constructur, copy all keys in the map with hyphens to camelCase keys with the same value).

And I've worked with the idea of returning a bool from `setProperty`, but ultimately think the search for dirty roots on end of updates batch is a great idea.

JoshMarler: @olidacombe, 

Been away from this for a while as other priorities took over but yeah nice! Map to setters looks great. I think you'd still want `isLayoutProp` or equivalent to `ShadowView::setProperty` (at whatever node/subtree level this occurs in), so that exceptions out of the map really are exceptional and not fired for every user defined prop etc. 

Also, I didn't think `juce::Identifier` has a `std::hash` specialisation for that map to work? @nick-thompson and I have actually had a few conversations about whether `juce::Identifier` is the right thing to use for `View` props. Every time a prop update comes in from JS/React we'll be constructing `juce::Identifier` instances on the fly (not an intended use pattern). So if you do have to switch that map to `juce::String` keys it might not be the end of the world. 

I've got this on my TODO list after a couple other things but if you want to carry on from where you've left off don't let me stop you! Looks like you're heading down a nice track.

olidacombe: Cool, thanks for having a look.  When I get some time I'll crack on then with all of the above in mind.

nick-thompson: Interesting, nice one! I love the map of functions idea, that seems very clean. Looking over it again, I think we could likely reduce all of the preprocessor macro code there to template functions, yea? If we can do that, I think we'd clean the code up even further (wish I had thought of this ages ago hah)

Re: #13– I think we should do that on the javascript side. In general, my philosophy is to write as little C++ as possible, but write it in a way such that it is maximally flexible and maximally stable. The reason for that is that iterating on the javascript side is much faster and much easier, and would require no build changes from consumers (like SOUL, where we want to be able to ship new features and updates to SOUL end users without requiring SOUL to pull + rebuild a new binary, then requiring all SOUL users to update their install)

olidacombe: Yeah I don't love the macros, especially not being able to condition on 0 variadic args, so will make a pass with variadic templates at some point.

As for https://github.com/nick-thompson/blueprint/issues/13 I wrote [babel-plugin-transform-property-cases](https://github.com/olidacombe/babel-plugin-transform-property-cases) to do this on javascript side at compile time (e.g. usage https://github.com/nick-thompson/blueprint/compare/master...olidacombe:kebabinate, annoyingly needs `transformRotate` manually mapped because `transform-rotate` is not css), but don't think it's worthy for inclusion in examples because there may be hidden pitfalls (say if someone iterates though some `{top: 5, botton: 6}` object to dynamically build `marginTop, marginBottom`, the transform can't help you).  It's fine if you know the limitations though.  Anyway, that's why I had designs on doing it C++ side at initialization because that would be fast.  Perfectly happy to abandon that idea.

Side note for @JoshMarler looking back at the map usage you get a single O(logN) lookup without throwing via `.find` and checking the iterator, so will move over to that instead of a validation pass followed by the lookup.  I was just being a rookie there.

nick-thompson: Awesome, thanks!

Right, yea, I think we will probably want to update `transform-rotate` anyways to be more at parity with CSS. That is, instead of supporting a `transform-rotate` property, support a `transport` property with syntax like `transform: rotate(90deg);`. I think trying to do this at compile time is likely harder than doing it at runtime (and same thing for the camel case transform). For example, we can still write our native property names with hyphens (which seems to be helpful with Yoga's enum to string helper), but catch and transform property names at the javascript layer before it reaches native right here: https://github.com/nick-thompson/blueprint/blob/master/packages/juce-blueprint/src/lib/BlueprintBackend.ts#L94. Similarly, we can implement what I'll call "macro" properties right there.

For example:
```js
  setProperty(propKey: string, value: any): any {
    this._props = Object.assign({}, this._props, {
      [propKey]: value,
    });

    if (propKey === 'transform') {
      return handleTransformMacro(value);
    }

    .... etc

    // Note the prop key transformation here
    return __BlueprintNative__.setViewProperty(this._id, camelToHyphen(propKey), value);
  },

  handleTransformMacro(value) {
    let transformList = value.split(' ');
    transformList.forEach((t) => {
      // Regex to match over e.g. rotate(90deg) such that
      // transformFunction = "rotate";
      // transformValue = 90;
      this.setProperty(`transform-${transformFunction}`, transformValue);
    });
  }
```

So you can see here that as long as our native properties support the necessary transforms, we can always compose their behavior into macro functions that support a more familiar API for the end user.

Generally I think this is the place and the strategy we'll want to use to providing a good user experience in terms of view props.

olidacombe: Cool, I'm happy to do that, and will open a PR soon.  Runtime in javascript was bottom of my list for performance reasons but I see your reasoning.  It's also just easier 👍

nick-thompson: I get that, but I favor the simplicity :) And I suspect the runtime cost here will be insignificant, but if it's not, `shouldComponentUpdate` is more likely the correct optimization

olidacombe: I've made a pass at ditching those macros.  I still used a couple of short ones to append `Percent` and `Auto` to symbols, but it's a bit cleaner now.  Also cut out the step of constructing a `YGValue`, since it didn't seem necessary.

The diff looks like this https://github.com/olidacombe/blueprint/compare/improve_shadowview_performance...olidacombe:improve_shadowview_performance-macro_strip

The `PropertySetterMap::K` typedef is ready to flip over to `juce::String` as @JoshMarler suggests if we want to do that.

olidacombe: I also had a look at how to trim calls to `performRootShadowTreeLayout()`.  Do you think `resetAfterCommit` is the right signal to trigger a traverse for dirty yoga nodes (e.g. https://github.com/olidacombe/blueprint/compare/improve_shadowview_performance-macro_strip...olidacombe:resetAfterCommit-hook)?

olidacombe:
@JoshMarler @nick-thompson I've been working on this a bit, and merging master into my branch to stay up to date so the diff with this PR is quite unreadable.  If you want to see where I'm up to keep an eye on https://github.com/nick-thompson/blueprint/compare/master...olidacombe:improve_shadowview_performance .  Alternatively I could open a PR just like this one, but that seemed a bit cheeky.

JoshMarler:
> @JoshMarler @nick-thompson I've been working on this a bit, and merging master into my branch to stay up to date so the diff with this PR is quite unreadable. If you want to see where I'm up to keep an eye on [master...olidacombe:improve_shadowview_performance](https://github.com/nick-thompson/blueprint/compare/master...olidacombe:improve_shadowview_performance) . Alternatively I could open a PR just like this one, but that seemed a bit cheeky.

@olidacombe , No need to spare any cheek for my sake! If you've tackled it great! I'll run over your branch tomorrow. @nick-thompson , if @olidacombe has gone further with this maybe we close this branch and open up PR for the new version?

nick-thompson:
@JoshMarler @olidacombe my apologies! I clearly lost track of this one.

> Alternatively I could open a PR just like this one, but that seemed a bit cheeky.

> if @olidacombe has gone further with this maybe we close this branch and open up PR for the new version? I've not had time to get to this and it's sat stale for far too long.

Yea, that sounds great. @olidacombe please do feel free to open that PR. I've just looked over your branch briefly and it looks fantastic. I will spend some more time with it on the PR but my early impression is great. And +1 for @JoshMarler's idea for an A/B profile if you have the time, would love to see the data

JoshMarler:
Closing as @olidacombe has taken this much further than I did.

Issue #31, opened 2019-12-17
AnimatedView


nick-thompson:
The `blueprint::View` has a `setFloatBounds` method that the ShadowView calls into during a layout step. That method then delegates on to the `juce::Component`. I want to continue keeping this piece decoupled from the `juce::Component` interface, but there's a neat opportunity here because of the decoupling:

I want to create an `AnimatedView`, similiar to juce's `ComponentAnimator`, where when the `View` receives a call to `setFloatBounds`, instead of immediately calling the `juce::Component::setBounds`, start a timer at 60Hz or something to animate the bounds of the `juce::Component`. Basically on every tick of the timer nudge the current bounds closer to the target bounds until they're close enough to equal, at which point we stop the timer. How much to nudge on every frame can be configured via a settable property with easeIn, easeOut, linear, etc, options.

JoshMarler:
Hey @nick-thompson ! 

I'm just starting to look into this as I need a horizontal Expander/Accordion component which expands and shows some content in response to a button click event etc. 

So, it seems like just wrapping `juce::ComponentAnimator` in a `blueprint::AnimateView` or similar would be reasonably straight forward. However I'm curious about your ideas and obviously would like to contribute something directly if I can rather than rolling my own interpretation of this. 

So you mention starting a timer and gradually incrementing/decrementing bounds, basically exactly what `juce::ComponentAnimator` does. What's the rational behind doing blueprint's own thing here as opposed to just wrapping `juce::ComponentAnimator` ? 

I've also been looking at the API for ReactNative's `Animator` API: https://reactnative.dev/docs/animations

Obviously this is much more comprehensive than the simple bounds/opacity animations offered by `juce::ComponentAnimator`. I wonder whether mimicking it's API is worth consideration though, it could provide a powerful framework to follow. I appreciate this all has a bit of overlap with the <Canvas> API's animation offerings. 

Be cool to get a discussion going on this one!

Cheers

JoshMarler:
This also looks like a relevant bit of API inspiration: https://reactnative.dev/docs/layoutanimation

nick-thompson:
Awesome, yea, happy to kick this convo off! I think this would be a really fun component to have.

I've studied the RN APIs quite a bit already as I've thought about this task. I have a couple high level thoughts and I actually also have a branch I started months ago that has this mostly working. I'll push the branch up just for discussion's sake in a minute.

[...]

PR #85, opened 2020-08-05
View bounds animation

nick-thompson:
Wrapping up what we started on #31, @JoshMarler mind taking a quick look?

### DIFF
--- a/blueprint/core/blueprint_ShadowView.h
+++ b/blueprint/core/blueprint_ShadowView.h
@@ -15,6 +15,80 @@
 namespace blueprint
 {
 
+    struct BoundsAnimator : public juce::Timer {
+        using StepCallback = std::function<void(juce::Rectangle<float>)>;
+
+        enum class EasingType {
+            Linear,
+            QuadraticIn,
+            QuadraticOut,
+            QuadraticInOut,
+        };
+
+        juce::Rectangle<float> start;
+        juce::Rectangle<float> dest;
+        StepCallback callback;
+        double duration;
+        double startTime;
+        int frameRate = 45;
+        EasingType easingType = EasingType::Linear;
+
+        BoundsAnimator(double durationMs, int frameRate, EasingType et, juce::Rectangle<float> start, juce::Rectangle<float> dest, StepCallback cb)
+            : start(start)
+            , dest(dest)
+            , callback(cb)
+            , duration(durationMs)
+            , frameRate(frameRate)
+            , easingType(et)
+        {
+            startTime = juce::Time::getMillisecondCounterHiRes();
+            startTimerHz(45);
+        }
+
+        ~BoundsAnimator() {
+            stopTimer();
+        }
+
+        float lerp(float a, float b, float t) {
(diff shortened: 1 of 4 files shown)


JoshMarler:
@nick-thompson, dope. One is very excited. Will look over now.

JoshMarler:
@nick-thompson, Awesome, awesome, awesome. That is all. 

Ha no seriously looks great. I pulled it down and ran the example. I like the interface now. Very neat. I wonder if doing a similar thing for opacity handled in `View` would be worth while. 

i.e. 

`<View opacityAnimated />`

That would match the capabilities of `juce::ComponentAnimator`. It should be able to take the same set of properties as the `layoutAnimated` prop/object. 

Just food for thought, that could be argued as a task an `AnimatedValue` implementation but I feel like it could be a valuable half-way house? I'd be happy to implement and just follow the ground work you've already laid down ?

Things like the switch on `EasingType` for the `lerp` call could be pulled to some generic location to be shared by `BoundsAnimator` and `OpacityAnimator`. 

Looks great though. Looking forward to using this, I've a carousel like component to implement!

nick-thompson:
Great, happy to hear that!

Yea I kind of just skipped right over opacity, maybe just due to personal bias (I'm realizing I don't often use opacity animation, I guess!). I'd be curious to hear the use cases you have in mind?

Also wondering if it'd make sense to just add opacity animation to the same prop?

```js
<View {...styles.container} />
```

JoshMarler:
> Great, happy to hear that!
> 
> Yea I kind of just skipped right over opacity, maybe just due to personal bias (I'm realizing I don't often use opacity animation, I guess!). I'd be curious to hear the use cases you have in mind?
> 
> Also wondering if it'd make sense to just add opacity animation to the same prop?
> 
> ```js
> <View {...styles.container} />
> 
> const styles = {
>     container: {
>         layoutAnimated: {
>             duration: 200.0,
>             easing: View.EasingFunctions.quadraticInOut,
>             opacityStart: 0.1,
>             opacityEnd: 0.9
>         }
>     }
> };
> ```
> 
> I realize opacity isn't quite "layout," but in another sense it seems to make sense to group them.

Yeah interesting. I'm no sure whether they should be kept distinct or not at the moment. To be honest, I haven't required opacity animation just yet, I'll take a read through my current project's design specs again and see if I need it. 

I don't have strong feelings about opacity animation being a distinct property.... That might be one to muse over.

nick-thompson:
:+1: sounds good, let me know! Maybe it's better to wait until we have a concrete case to help inform the API design. If you have a related feature in your current project, let's discuss here. Otherwise, I think we go ahead with layout animation for now and return to opacity when that first concrete case comes up.

JoshMarler:
I will have an example for opacity transitions (So a clever UX man tells me) but for now I agree, just go with layout and we can deal with opacity later. Merge it, merge it, merge it! 

https://media.giphy.com/media/69iB0zVgoKkatcVDrA/giphy.gif

nick-thompson:
Done! :D

PR #86, opened 2020-08-07
Rename main file to match generated typescript module

rbaxter08:
This should resolve the issue @JoshMarler is describing here:

https://github.com/nick-thompson/blueprint/pull/67#issuecomment-669844428

The issue is that when you import `... from 'juce-blueprint';` typescript is looking for a module with the name `juce-blueprint`, but can't find it because the actual module name is created from the source file, so `Blueprint` in this case. 

So, two possible fixes:

1. We can rename the file from `Blueprint` -> `juce-blueprint` (This PR)
2. @JoshMarler You can change your import from `... from 'juce-blueprint'` to `... from 'Blueprint'` and adjust the dependency name in your `package.json`

I think `#1` is the nicer fix for users, so they won't have to adjust imports. 

#### Screenshot showing that the import is properly resolved with this change:

Screen Shot 2020-08-07 at 12 12 13 PM


rbaxter08:
@nick-thompson Would you be okay with converting the GainPlugin example to Typescript?

JoshMarler:
+1 on the first fix. Prefer the package name matching the directory if poss.

nick-thompson:
Great, yea happy to merge this. Thanks!

I'm definitely interested in starting to convert the GainPlugin example, and I have two kind of preliminary thoughts:

1) I want to briefly prioritize getting the ball rolling on #22 first; I've had some reports of people struggling to get past some typescript issues in the new project setup flow, so I want to take some safeguards to make sure that as we make the transition we're still making the project accessible to new users. I'll add some details below
2) I wonder if it will be important to have a "plain javascript" example alongside the typescript example. For people coming from C++ land, I anticipate there will be basically two camps: the first is people who are super happy to have typescript support because they prefer type safety and Typescript will feel more familiar. The second camp is people who are generally intimidated by all of the tooling in javascript land (babel, npm, webpack, etc), and don't want the extra layer of tooling from Typescript. So I think by our examples it might be helpful to make clear that users can write with typescript or write in plain ol' js. There's no immediate action item here, just a general thought that we try to make sure we're not leaving any users behind as we migrate to typescript. I have one more example myself that I want to work on when I next find some time. Maybe we take GainPlugin to typescript and the next example project we introduce as plain js? I'll also solicit more example project ideas if y'all have them! 

More details on #22: I think we could add a very simple "first pass" here which just runs through the `npm` related steps of the new project flow. Install + build juce-blueprint, `npm run init` a new project in a tmp dir, cd to the tmp dir, npm install/npm run build, then finally cd to the GainPlugin example and npm run build. This should be pretty easy to get up and running and I think would very sufficiently cover our bases as we migrate.

Let me know what y'all think!

PR #87, opened 2020-08-14
Fix Visual Studio compile errors

JoshMarler:
@nick-thompson, I thought this had been fixed but looks like there are a couple VS errors with -Werror etc.


nick-thompson:
Ah, thanks!

PR #88, opened 2020-08-14
Fix onMeasure being a required prop on CanvasView.

JoshMarler:
@nick-thompson, this minor thing is causing me errors. I don't think `onMeasure` needs to be a required property in `CanvasView` ?


nick-thompson:
Yup, can totally be optional. For example:

```
<Canvas {...styles.canvas} onDraw={this._onDraw} />

where styles.canvas = {
    width: 500,
    height: 500,
    flex: 0,
};
```

User never needs the measurement, they know it will always be 500x500.

Thanks!

Issue #64, opened 2020-06-16
Distinguish Errors and Fatal Errors


nick-thompson:
Sparked from conversation in PR #61: https://github.com/nick-thompson/blueprint/pull/61#issuecomment-644797255

In Duktape we have sort of two classes of errors: normal errors from which we can recover the js engine, and fatal errors from which we cannot. A normal error that goes uncaught becomes a fatal error, and so do explicit fatal cases from within Duktape itself. In ReactAppRoot we can catch almost all errors by making sure we do everything inside `duk_pcall` wrappers, and when we catch those errors we pop up the red stacktrace screen and wait for the user to update the bundle. This works great in dev, but we need a way to let the user define a custom error handler override so that, for example, they don't ship this red stacktrace screen in production. 

Then we need a second override (or perhaps just a flag to the former override) to distinguish for cases where we're dealing with a fatal error, which I anticipate being rare cases, but cases we need to cover. I think it's fair for our default implementation to be just throwing a C++ error and crashing out, but obviously users should be able to customize that behavior.

JoshMarler:
Yeah this covers this nicely. Personally I'm for the flag argument to the existing onUnhandledError callback/override.

nick-thompson:
:+1: I think I lean the same way!

nick-thompson:
Resolved in e3d5057

Issue #44, opened 2020-05-20
Clean up `dispatchViewEvent`


nick-thompson:
With closing #32, we have a much nicer way to support invoking function props from a native view implementation. For example: https://github.com/nick-thompson/blueprint/commit/97ebe8cbe440bc384b10b08e8ab46273645ff2b7#diff-0fb21fdaed83bdf217e9ec1889cc57afR85

That means that all of the `dispatchViewEvent` stuff (e.g. https://github.com/nick-thompson/blueprint/blob/326867e9907cb20edcd7ea34b578d76d1a773364/packages/juce-blueprint/esm/lib/BlueprintBackend.js#L86) can be removed and "view events" can call callback props directly (e.g. https://github.com/nick-thompson/blueprint/blob/master/blueprint/core/blueprint_View.cpp#L156)

nick-thompson:
Closing this as fixed in 7defca5

Issue #43, opened 2020-05-20
Clean up the `nativeMethodWrapper`


nick-thompson:
Commit 3811b0687d63059635254d8b7548abeec3a60877 introduces a better way to handle native method registration, used in the `pushVarToDukStack` and `readVarFromDukStack` methods to support passing e.g. a `DynamicObject` that has a `juce::var::NativeFunction` as a property.

The original `nativeMethodWrapper` is still lingering in there as used by the `registerNativeMethod` API, but it should be replaced with the newer stuff. See the TODOs inline: https://github.com/nick-thompson/blueprint/commit/3811b0687d63059635254d8b7548abeec3a60877#diff-9dd78235906a309d67f6d946b57e928fR17

JoshMarler:
@nick-thompson, having spent a bunch of time around here over the last few days I'd be more than happy to give this a go. Would be great to be able to use capture blocks with the `registerNativeMethod` lambdas and get away from the stash dance. 

Let me know if you particularly wanted this one yourself and if not I'd be interested in taking it.

JoshMarler:
Presumably we can kill of this when done: https://github.com/nick-thompson/blueprint/issues/29

nick-thompson:
Resolved in e3d5057

Issue #29, opened 2019-12-11
Allow references inside lamda


pauliusuza:
I'm blocked by `appRoot->engine.registerNativeMethod` not allowing to pass variables inside lambda by reference: `[&](void* stash, const juce::var::NativeFunctionArgs& args)` - note the `[&](...`

I'd like to reference the cast of my processor inside the lambda function. Would it be possible to add support for this use case?

Great Library. Thanks.

nick-thompson:
Hey @pauliusuza, sorry for the delay here, I'm not sure why but I'm not getting notifications for new issues... need to figure that out.

That's right, yea, the native methods need to conform to a c-style function pointer interface because of how duktape works with function calls. Lambdas without a capture group can be cast as such, but as soon as you capture, it breaks the cast. The workaround is to use that stash argument:

```cpp
struct MyData {
    RefToMyThing& thing1;
    RefToMyOtherThing& thing2;
};

MyData myStash { x, y };

appRoot->engine.registerNativeMethod("test", [](void* stash, const juce::var::NativeFunctionArgs& args) {
    auto* myStash_ = reinterpret_cast<MyData*>(stash);
    RefToMyThing& thing1 = myStash->thing1;
    RefToMyOtherThing& thing2 = myStash->thing2;

    ...
}, &myStash);
```

Of course if you follow this example you have to be very careful with your object lifetimes. Let me know if that works for you!

nick-thompson:
Resolved in e3d5057

PR #89, opened 2020-08-15
Refactor and clean up EcmascriptEngine (#29, #43, #64)

nick-thompson:
Okayyy this is a little bit of work to stabilize the EcmascriptEngine API and clean up the use of our lambdaReleasePool.

Essentially this introduces three things:
1) Refactoring our error handling to `throw`, which simplifies the error handling code in the consumer (ReactAppRoot, for one). In ReactAppRoot, I'm catching and handling `EcmascriptEngine::Error`, because we know we can recover, and ignoring any case where we throw `EcmascriptEngine::FatalError`. That one should be handled above ReactAppRoot by the user, because something's really wrong in that case, and the application will have to decide the best course of action
2) Cleaning up our native method API to now align around the use of lambdas with captures and the lambdaReleasePool, which also helps clean up the consumer code (i.e. ReactAppRoot)
3) Improve the performance of the lambdaReleasePool. In particular, I was aiming at the fact that Duktape doesn't always run it's garbage collector when you might want it to, and at the same time, objects with finalizers (any of our lambda helpers) need 2 garbage collector passes (first one runs the finalizer, second one cleans the object if the finalizer didn't save it). With something like animated canvas paints, our lambdaReleasePool grows like crazy when the garbage collector isn't being explicitly managed. So, to address that, I've added two explicit calls to the garbage collector after instances where we know we may have just allocated a bunch of temporaries. Specifically this is after function calls (`EcmascriptEngine::invoke`, or invoking a `juce::var::NativeFunction` associated with the engine), and this keeps our release pool at near constant size. I also, kind of on a whim, changed the lambdaReleasePool id to an integer rather than juce::Uuid, just on a hunch that we'll be constructing, destructing, and hashing these things _all_ the time. Int felt a little more lightweight, though I have not explicitly measured

After this goes in, I'll tackle #44, which should be fairly quick after these engine improvements.

@JoshMarler mind taking a look over my code here?

### DIFF
--- a/blueprint/core/blueprint_EcmascriptEngine.cpp
+++ b/blueprint/core/blueprint_EcmascriptEngine.cpp
@@ -14,119 +14,51 @@ namespace blueprint
     namespace detail
     {
 
-        /** TODO: This nativeMethodWrapper is on its way out and needs to be replaced
-         *  with the LambdaHelper as used in pushVarToDukStack/readVarFromDukStack.
-         *  The helper allows lambdas with capture groups, so we can skip the manual
-         *  passing of stash in the `registerNativeMethod` hooks as well.
-         */
-        static duk_ret_t nativeMethodWrapper (duk_context* ctx)
-        {
-            // First we have to retrieve the actual function pointer and our engine pointer
-            // See: https://duktape.org/guide.html#hidden-symbol-properties
-            duk_push_current_function(ctx);
-            duk_get_prop_string(ctx, -1, DUK_HIDDEN_SYMBOL("NativeFunctionPtr"));
-
-            auto function = reinterpret_cast<EcmascriptEngine::NativeFunction>(duk_get_pointer(ctx, -1));
-            duk_pop(ctx);
-
-            // Then the engine...
-            duk_get_prop_string(ctx, -1, DUK_HIDDEN_SYMBOL("EnginePtr"));
-            auto* engine = static_cast<EcmascriptEngine*>(duk_get_pointer(ctx, -1));
-            duk_pop(ctx);
-
-            // Then the stash...
-            duk_get_prop_string(ctx, -1, DUK_HIDDEN_SYMBOL("StashPtr"));
-            auto* stash = duk_get_pointer(ctx, -1);
-
-            // Pop back both the pointer and the "current function"
-            duk_pop_2(ctx);
-
-            // Now we can collect our args
-            std::vector<juce::var> args;
-            int nargs = duk_get_top(ctx);
-
-            for (int i = 0; i < nargs; ++i)
-                args.push_back(engine->readVarFromDukStack(ctx, i));
-
-            // Now we can invoke the user method with its arguments
-            auto result = function(stash, juce::var::NativeFunctionArgs(
-                juce::var(),
-                args.data(),
-                static_cast<int>(args.size())
-            ));
-
-            // For an undefined result, return 0 to notify the duktape interpreter
-            if (result.isUndefined())
-                return 0;
-
-            // Otherwise, push the result to the stack and tell duktape
-            engine->pushVarToDukStack(ctx, result);
-            return 1;
-        }
-
         static void fatalErrorHandler (void* udata, const char* msg)
         {
-            (void) udata;  // Ignored in this case, silence warning
-
-            DBG("**Blueprint Fatal Error:** " << msg);
-            DBG("It is now unsafe to execute further code in the EcmascriptEngine.");
-
-            throw std::runtime_error(msg);
+            (void) udata; // Ignored in this case, silence warning
+            throw EcmascriptEngine::FatalError(msg);
         }
 
-        //TODO: Keep safeEvalString and safeCompileFile or make EcmascripEngine member funcs
-        //      now that we have out errorHandle wrapper in EcmascriptEngine? Passing
-        //      callback here feels a bit ambiguous in regards to popping the error
-        //      off of the stack after invoking the error callback. Whose responsibility is it etc.
-        template <typename T>
-        static bool safeEvalString(duk_context* ctx, const juce::String& s, T&& callback)
+        static void safeCall(duk_context* ctx, const int numArgs)
         {
-            if (duk_peval_string(ctx, s.toRawUTF8()) != 0)
+            if (duk_pcall(ctx, numArgs) != DUK_EXEC_SUCCESS)
             {
-                if (callback != nullptr)
-                {
-                    // Call the user provided error handler and indicate failure
-                    std::invoke(callback);
-                    return false;
-                }
+                const juce::String stack = duk_safe_to_stacktrace(ctx, -1);
+                const juce::String msg = duk_safe_to_string(ctx, -1);
 
-                duk_throw_raw(ctx);
+                throw EcmascriptEngine::Error(msg, stack);
             }
-
-            // Indicate success
-            return true;
         }
 
-        template<typename T>
-        static bool safeCompileFile(duk_context* ctx, const juce::File& file, T&& callback)
+        static void safeEvalString(duk_context* ctx, const juce::String& s)
         {
+            if (duk_peval_string(ctx, s.toRawUTF8()) != DUK_EXEC_SUCCESS)
+            {
+                const juce::String stack = duk_safe_to_stacktrace(ctx, -1);
+                const juce::String msg = duk_safe_to_string(ctx, -1);
+
+                throw EcmascriptEngine::Error(msg, stack);
+            }
+        }
+
+        static void safeCompileFile(duk_context* ctx, const juce::File& file)
+        {
+            auto name = file.getFileName();
+            auto body = file.loadFileAsString();
+
             // Push the js filename to be compiled/evaluated
-            duk_push_string(ctx, file.getFileName().toRawUTF8());
+            duk_push_string(ctx, name.toRawUTF8());
 
-            if (duk_pcompile_string_filename(ctx, DUK_COMPILE_EVAL, file.loadFileAsString().toRawUTF8()) != 0)
+            if (duk_pcompile_string_filename(ctx, DUK_COMPILE_EVAL, body.toRawUTF8()) != DUK_EXEC_SUCCESS)
             {
-                if (callback != nullptr)
-                {
-                    // Call the user provided error handler.js and indicate failure
-                    std::invoke(callback);
-                    return false;
-                }
+                const juce::String stack = duk_safe_to_stacktrace(ctx, -1);
+                const juce::String msg = duk_safe_to_string(ctx, -1);
 
-                duk_throw_raw(ctx);
+                throw EcmascriptEngine::Error(msg, stack);
             }
-
-            // Indicate success
-            return true;
        }
 
-/*
-       static void dumpContext(duk_context* ctx)
-       {
-           duk_push_context_dump(ctx);
-           DBG(duk_to_string(ctx, -1));
-           duk_pop(ctx);
-       }
-*/
     }
 
     //==============================================================================
@@ -137,31 +69,6 @@ namespace blueprint
 
         // Add console.log support
         duk_console_init(dukContext, DUK_CONSOLE_FLUSH);
-
-        // Create our error handler wrapping a user supplied error callback.
-        errorHandler = [=]
-        {
-            if (onUncaughtError == nullptr)
-                duk_throw_raw(dukContext);
-
-            const juce::String trace = duk_safe_to_stacktrace(dukContext, -1);
-            const juce::String msg = duk_safe_to_string(dukContext, -1);
-
-            // Call the user provided error handler
-            std::invoke(onUncaughtError, msg, trace);
-
-            // Clear out the stack so we can re-register native functions
-            // after we clear out the lambda release pool etc.
-            while (duk_get_top(dukContext))
-            {
-                duk_remove(dukContext, duk_get_top_index(dukContext));
-            }
-
-            // Clear the LambdaHelper release pool as duktape does not call object
-            // finalizers in the event of an evaluation error or duk_pcall failure.
-            lambdaReleasePool.clear();
-        };
-
     }
 
     EcmascriptEngine::~EcmascriptEngine()
@@ -174,8 +81,12 @@ namespace blueprint
     {
         jassert(code.isNotEmpty());
 
-        if (!detail::safeEvalString(dukContext, code, errorHandler))
-            return juce::var(EvaluationError);
+        try {
+            detail::safeEvalString(dukContext, code);
+        } catch (Error const& err) {
+            reset();
+            throw err;
+        }
 
         auto result = readVarFromDukStack(dukContext, -1);
         duk_pop(dukContext);
@@ -188,66 +99,30 @@ namespace blueprint
         jassert(code.existsAsFile());
         jassert(code.loadFileAsString().isNotEmpty());
 
-        juce::var result = juce::var::undefined();
-
-        if (!detail::safeCompileFile(dukContext, code, errorHandler))
-            return juce::var(EvaluationError);
-
-        // Call compiled function
-        if (duk_pcall(dukContext, 0) != DUK_EXEC_SUCCESS)
-        {
-            errorHandler();
-            return juce::var(EvaluationError);
+        try {
+            detail::safeCompileFile(dukContext, code);
+            detail::safeCall(dukContext, 0);
+        } catch (Error const& err) {
+            reset();
+            throw err;
         }
 
         // Collect the return value
-        result = readVarFromDukStack(dukContext, -1);
+        auto result = readVarFromDukStack(dukContext, -1);
         duk_pop(dukContext);
 
         return result;
     }
 
     //==============================================================================
-    void EcmascriptEngine::registerNativeMethod (const juce::String& name, NativeFunction fn, void* stash)
+    void EcmascriptEngine::registerNativeMethod (const juce::String& name, juce::var::NativeFunction fn)
     {
-        duk_push_global_object(dukContext);
-
-        // We wrap the native function to provide a helper layer storing and retrieving the
-        // stash, and marshalling between the Duktape C interface and the NativeFunction interface
-        duk_push_c_function(dukContext, detail::nativeMethodWrapper, DUK_VARARGS);
-
-        // Now we assign the pointers as properties of the wrapper function
-        duk_push_pointer(dukContext, (void *) fn);
-        duk_put_prop_string(dukContext, -2, DUK_HIDDEN_SYMBOL("NativeFunctionPtr"));
-        duk_push_pointer(dukContext, (void *) this);
-        duk_put_prop_string(dukContext, -2, DUK_HIDDEN_SYMBOL("EnginePtr"));
-        duk_push_pointer(dukContext, stash);
-        duk_put_prop_string(dukContext, -2, DUK_HIDDEN_SYMBOL("StashPtr"));
-
-        // And finally we assign the function to its name in the global namespace
-        duk_put_prop_string(dukContext, -2, name.toRawUTF8());
+        registerNativeProperty(name, juce::var(fn));
     }
 
-    void EcmascriptEngine::registerNativeMethod (const juce::String& target, const juce::String& name, NativeFunction fn, void* stash)
+    void EcmascriptEngine::registerNativeMethod (const juce::String& target, const juce::String& name, juce::var::NativeFunction fn)
     {
-        // Evaluate the target string on the context, leaving the result on the stack
-        if (!detail::safeEvalString(dukContext, target, errorHandler))
-            return;
-
-        // We wrap the native function to provide a helper layer storing and retrieving the
-        // stash, and marshalling between the Duktape C interface and the NativeFunction interface
-        duk_push_c_function(dukContext, detail::nativeMethodWrapper, DUK_VARARGS);
-
-        // Now we assign the pointers as properties of the wrapper function
-        duk_push_pointer(dukContext, (void *) fn);
-        duk_put_prop_string(dukContext, -2, DUK_HIDDEN_SYMBOL("NativeFunctionPtr"));
-        duk_push_pointer(dukContext, (void *) this);
-        duk_put_prop_string(dukContext, -2, DUK_HIDDEN_SYMBOL("EnginePtr"));
-        duk_push_pointer(dukContext, stash);
-        duk_put_prop_string(dukContext, -2, DUK_HIDDEN_SYMBOL("StashPtr"));
-
-        // And finally we assign the function to its name in the global namespace
-        duk_put_prop_string(dukContext, -2, name.toRawUTF8());
+        registerNativeProperty(target, name, juce::var(fn));
     }
 
     //==============================================================================
@@ -260,10 +135,12 @@ namespace blueprint
 
     void EcmascriptEngine::registerNativeProperty (const juce::String& target, const juce::String& name, const juce::var& value)
     {
-        // Evaluate the target string on the context, leaving the result on the stack
-        // TODO: Return specific error code if target not in stack?
-        if (!detail::safeEvalString(dukContext, target, errorHandler))
-            return;
+        try {
+            detail::safeEvalString(dukContext, target);
+        } catch (Error const& err) {
+            reset();
+            throw err;
+        }
 
         // Then assign the property
         pushVarToDukStack(dukContext, value);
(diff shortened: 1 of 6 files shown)


JoshMarler:
This is really exciting Nick! I'll be at my desk later this evening so will give this all a proper look over. I'd be happy to A/B profile and throw you some metrics as well if you haven't already done so. Think I still have some previous profiling data to compare with etc. Cheers.

blueprint/core/blueprint_EcmascriptEngine.cpp
```diff
@@ -14,119 +14,51 @@ namespace blueprint
...
 
-        //TODO: Keep safeEvalString and safeCompileFile or make EcmascripEngine member funcs
-        //      now that we have out errorHandle wrapper in EcmascriptEngine? Passing
-        //      callback here feels a bit ambiguous in regards to popping the error
-        //      off of the stack after invoking the error callback. Whose responsibility is it etc.
-        template <typename T>
-        static bool safeEvalString(duk_context* ctx, const juce::String& s, T&& callback)
+        static void safeEvalString(duk_context* ctx, const juce::String& s)
         {
             if (duk_peval_string(ctx, s.toRawUTF8()) != 0)
```
JoshMarler, line 36: Probably my bad. Better as `DUK_EXEC_SUCCESS` here though I reckon.

blueprint/core/blueprint_EcmascriptEngine.cpp
```diff
@@ -14,119 +14,51 @@ namespace blueprint
...
         {
+            auto const& name = file.getFileName().toRawUTF8();
+            auto const& body = file.loadFileAsString().toRawUTF8();
+
             // Push the js filename to be compiled/evaluated
-            duk_push_string(ctx, file.getFileName().toRawUTF8());
+            duk_push_string(ctx, name);
 
-            if (duk_pcompile_string_filename(ctx, DUK_COMPILE_EVAL, file.loadFileAsString().toRawUTF8()) != 0)
+            if (duk_pcompile_string_filename(ctx, DUK_COMPILE_EVAL, body) != 0)
```
JoshMarler, line 53: `DUK_EXEC_SUCCESS` again maybe?

blueprint/core/blueprint_EcmascriptEngine.cpp
```diff
@@ -188,66 +99,30 @@ namespace blueprint
...
-        result = readVarFromDukStack(ctx, -1);
+        auto result = readVarFromDukStack(ctx, -1);
         duk_pop(ctx);
 
         return result;
     }
 
     //==============================================================================
-    void EcmascriptEngine::registerNativeMethod (const juce::String& name, NativeFunction fn, void* stash)
+    void EcmascriptEngine::registerNativeMethod (const juce::String& name, juce::var::NativeFunction fn)
```
JoshMarler, line 118: I'm so hyped that this is now a thing!

blueprint/core/blueprint_EcmascriptEngine.cpp
```diff
@@ -556,12 +446,23 @@ namespace blueprint
...
 
                             // Clean the result and the stash off the top of the stack
                             duk_pop_2(ctx);
 
+                            // Here too, we know that we have potentially just pushed temporaries
+                            // to the stack. Duktape will eventually get around to garbage collection,
+                            // but specifically where we have LambdaHelper allocations for these
+                            // temporaries, we want to keep our release pool fairly small. For that
+                            // reason, we manually invoke a garbage collection pass right away.
+                            duk_gc(ctx, 0);
```
JoshMarler, line 464: Cool. I wonder if it would be worth adding some little helper function wrapped in an `#ifdef DEBUG_RELEASE_POOL` or something which dumps the size of the release pool at certain points and prints whenever `removeLambdaHelper` is called ? Might be a useful thing to toggle on whenever similar such performance issues arise.

JoshMarler:
@nick-thompson, FYI I've pulled this down to start testing and in both my app and the GainPlugin I'm seeing an error at startup. Will investigate further.

blueprint/core/blueprint_EcmascriptEngine.cpp
nick-thompson: Same :D

blueprint/core/blueprint_EcmascriptEngine.cpp
nick-thompson: Yea I had written that actually, but I removed it because I was seeing a constant release pool size in the GainPlugin example with the canvas animation, so the repetitive logs became not much value after the first few lines.

Think it'd be worth adding back in still?

nick-thompson:
Huh... what's the error you see? I have it building and running fine on MacOS.

JoshMarler:
Hey Nick, I'm not at my desk at the moment but it looked to be intermittent. Running the Gain Example a few times under xcode reproduced an eval error at startup.

I'll dive into it again when I get back to the office.

nick-thompson:
Yea, very interested thanks. Do you know offhand what version of Xcode you were running?

JoshMarler:
I think 11.2 but will sanity check when back.

nick-thompson:
Hm, I just tried it with Xcode11.2.1 on MacOS 10.14.4, all looks good on my end. Let me know the error details when you get a chance

JoshMarler:
@nick-thompson, 

OK so, 

This is definitely an intermittent thing. I cleaned the GainPlugin example and am seeing a syntax error reported from duktape on first line of the bundle:

```
Error in script evaluation: SyntaxError: source decode failed (line 1)
    at [anon] (
:1) internal
    at [anon] (../../../../../blueprint/blueprint/duktape/src-noline/duktape.c:82601) internal
```

I've not worked out the exact cause yet. I can only seem to reproduce this in production builds intermittently, as if duktape takes occasional offence to using an exclamation mark for the immediately invoked bundle function (not present in debug bundles). 

I'm running Xcode Version 11.6  on 10.15.2. 

I'll try stripping the Canvas timer paints and stuff out again to see if I can workout what is actually going on here. It's very odd.

JoshMarler:
@nick-thompson, also, I'm wondering about the API for things like `dispatchEvent` calls from a timer. What is the expected way for a user to stop dispatching events when a bundle error occurs when in a hot-reload workflow? We no longer wrap `dispatchEvent` in a `bundleValid` check. With `ReactApplicationRoot` always catching recoverable errors how should users be disabling and re-enabling their event dispatching? 

Should a timer be stopped in `beforeBundleEval` and restarted in `afterBundleEval` ?

blueprint/core/blueprint_ReactApplicationRoot.h
```diff
@@ -195,35 +189,31 @@ namespace blueprint
             // Clear error state from previous js evals
             errorText.reset();
 
-            if (!bundleValid)
```
JoshMarler, line 198: This bundleValid check was originally here to stop us from calling `registerNativeRenderingHooks` for a second time which results in duplicate function entries in the duktape stack. I think removing this opens us up for that issue again in hot-reload scenarios? 

i.e. you make a valid edit which causes no error so the stack is not cleared. Then we call evaluate in response to the bundle change and re-register our "native rendering hooks"

blueprint/core/blueprint_ReactApplicationRoot.h
```diff
@@ -195,35 +189,31 @@ namespace blueprint
...
-            auto result = engine.evaluate(bundle);
+                auto result = engine.evaluate(bundle);
 
-            bundleValid = result != EcmascriptEngine::EvaluationError;
+                if (afterBundleEval)
+                    afterBundleEval(bundle);
 
-            if (bundleValid && afterBundleEval)
-                afterBundleEval(bundle);
+                if (hotReloadEnabled)
```
JoshMarler, line 204: This hotReload block should probably be moved up above the try-catch block. It won't trigger and evaluation until the file actually changes. With this block inside the try catch block users cannot fixup their bundle if there is an error on initial load as the bundle will never be added to the watcher due to the new exception approach.

JoshMarler: i.e. 

```
        {
            JUCE_ASSERT_MESSAGE_THREAD

            // Clear error state from previous js evals
            errorText.reset();
            
            if (hotReloadEnabled)
            {
                jassert(bundleWatcher);

                if (!bundleWatcher->watching(bundle))
                    bundleWatcher->watch(bundle);
            }

            try {
                // Register internal React.js backend rendering methods
                registerNativeRenderingHooks();

                if (beforeBundleEval)
                    beforeBundleEval(bundle);

                auto result = engine.evaluate(bundle);

                if (afterBundleEval)
                    afterBundleEval(bundle);

                return result;
            } catch (const EcmascriptEngine::Error& err) {
                handleBundleError(err);
                return juce::var();
            }
        }
```

blueprint/core/blueprint_EcmascriptEngine.cpp
```diff
@@ -273,34 +150,38 @@ namespace blueprint
...
     juce::var EcmascriptEngine::invoke (const juce::String& name, const std::vector<juce::var>& vargs)
     {
-        // Evaluate the target string on the context, leaving the result on the stack
-        // TODO: Return specific error code if name not in stack?
-        if (!detail::safeEvalString(ctx, name, errorHandler))
-            return juce::var::undefined();
-
-        // Ensure requested function exists on the stack
-        duk_require_function(ctx, -1);
+        try {
```
JoshMarler, line 153: Should we indent brace here as per juce style guide?

i.e. 

```
try
{
 // stuff
}
catch (const auto &e)
{
// stuff
}
```

blueprint/core/blueprint_EcmascriptEngine.cpp
JoshMarler: It might not be that essential now. Assuming the weird issues I'm seeing currently aren't to do with the release pool!

JoshMarler: Hmm. Maybe not. I think the duk_context dump wrapped under some ifdef toggle could certainly be useful for user bug reports. Maybe some release pool dump out is still useful? I think a separate PR for more debug tools maybe rather than bloating out this set of changes though

nick-thompson:
Hey @JoshMarler, thank you for the excellent, detailed review. There's a lot to reply to here, I'm going to try to go piece by piece!



> This is definitely an intermittent thing. I cleaned the GainPlugin example and am seeing a syntax error reported from duktape on first line of the bundle:
> 
> ```
> Error in script evaluation: SyntaxError: source decode failed (line 1)
>     at [anon] (
> :1) internal
>     at [anon] (../../../../../blueprint/blueprint/duktape/src-noline/duktape.c:82601) internal
> ```
> 
> I've not worked out the exact cause yet. I can only seem to reproduce this in production builds intermittently, as if duktape takes occasional offence to using an exclamation mark for the immediately invoked bundle function (not present in debug bundles).
> 
> I'm running Xcode Version 11.6 on 10.15.2.
> 
> I'll try stripping the Canvas timer paints and stuff out again to see if I can workout what is actually going on here. It's very odd.
> 
> EDIT: On some runs I see this variant
> 
> ```
> Error in script evaluation: SyntaxError: invalid token (line 1)
>     at [anon] (
> :1) internal
>     at [anon] (../../../../../blueprint/blueprint/duktape/src-noline/duktape.c:84078) internal
> ```
> 
> What is really odd is that I'll sometimes get an eval error at startup but then if I just `touch` the js bundle (i.e. no actual content/js change) hot reload kicks in and everything works fine. Always seems to be on the initial bundle eval in some runs.

Ok so let's start here, and I'd like to defer the conversation about a global CanvasContext a little bit. My thought is that if we can get the garbage collection on temporaries working effectively, then temporary canvas contexts is the right way to go. If we can't manage to get GC performance to an acceptable level, let's revisit the global context idea.

So, to the errors above, the first thing this makes me wonder is if the bundle watcher is scooping up up the contents of the file before webpack has finished writing the whole bundle? Is that even a thing? I'm curious if you could write some little debug check in there to save the contents of the `juce::File` off to a separate temporary, and maybe when you see these eval errors we'll find that the temp file only has partial contents. This would explain why, when you touch the file later, the evaluation totally succeeds.

> Also, for what it's worth, I tested my app using the above `touch` trick with the bundle watcher change and it seems very laggy all of a sudden.

So this then makes me think that once we get past the eval errors, we'll be dealing with this head-on. It's odd, because with the GainPlugin example, I'm seeing totally solid performance with the gc even in debug mode. But, I think you're probably dealing with a far more complicated app that has many Canvas elements... which would mean that, on every JUCE paint cycle, we have several calls to install canvas temporaries, run the draw call, garbage collect, repeat. I can imagine that stacking up... this would actually be a good case for debouncing the garbage collection via juce's AsyncUpdater. That way we can run multiple draw calls and defer the GC until the last one has finished. 

Let's start with that– let's see if we can resolve the eval errors, and then I'll update this to debounce the GC and we'll see how that impacts performance.

blueprint/core/blueprint_EcmascriptEngine.cpp
nick-thompson: Yea, I should really add a clang-format/clang-tidy step :)

JoshMarler: Heh. Funnily enough we've just been looking at the same thing!

blueprint/core/blueprint_EcmascriptEngine.cpp
nick-thompson: Ok yea, sounds good! I think your app is more complicated than what I'm working against, so I'd be curious to hear from you as to what info you'd want to see in that dump out.

blueprint/core/blueprint_ReactApplicationRoot.h
nick-thompson: Ah, of course, great catch

JoshMarler:
@nick-thompson, 

Awesome. 

So, 

>So, to the errors above, the first thing this makes me wonder is if the bundle watcher is scooping up up the contents of the file before webpack has finished writing the whole bundle? Is that even a thing?

Yes possible and this has been a thing before but the error appeared to occur on initial eval before the watcher is actually even given the file. I'll sanity check this with some printouts and if this is the case I shall fix it!

In regards to my current apps performance, yes the app is of a reasonable size. I'm pushing meter events at 45hz for a number of tracks within a mixer. Each meter is rendered using a canvas view. I've sanity checked to ensure only the meter component is actually being rendered by React. 

Let me see if I can get the error out of the way and then I'll dig in a bit further.

blueprint/core/blueprint_ReactApplicationRoot.h
nick-thompson: Ahh yes, another great catch. Hm, interesting. On hot reload should we reset the engine? I.e. clear the value stack and reset the release pool?

JoshMarler: Yeah I think we discussed this briefly during the debugger work. We wan't to keep the Engine and the duk ctx alive to persist debugger connections during reloads (for debugging bundle load etc.). So we could expose some sort of `reset()` or `clear()` function on the engine to do this? Obviously this should all go into `handleBundleChanged` so that we re-register ViewManager and everything else etc.

nick-thompson:
> @nick-thompson, also, I'm wondering about the API for things like `dispatchEvent` calls from a timer. What is the expected way for a user to stop dispatching events when a bundle error occurs? i.e. in a hot-reload workflow. We no longer wrap `dispatchEvent` in a `bundleValid` check. With `ReactApplicationRoot` always catching recoverable errors how should users be disabling and re-enabling their event dispatching?
> 
> Should a timer be stopped in `beforeBundleEval` and restarted in `afterBundleEval` ? I guess this probably makes sense. In which case it might be worth updating the class docs to reflect this situation better.

Mhm, another great question. What do you think would be the ideal user experience here? My initial thought is that it would be cool for ReactApplicationRoot to just internally treat a `dispatchEvent` call as a no-op if the bundle is invalid. Might that cause problems in production?

I thought I could avoid the `bundleValid` flag because the errors would immediately short-circuit to the `catch` blocks, but obviously I overlooked a few cases!

blueprint/core/blueprint_EcmascriptEngine.cpp
JoshMarler: Sure. I'll have a play. Dumping the context and checking it's state certainly helped me resolve a few errors last time I was digging through here.

nick-thompson:
> In regards to my current apps performance, yes the app is of a reasonable size. I'm pushing meter events at 45hz for a number of tracks within a mixer. Each meter is rendered using a canvas view. I've sanity checked to ensure only the meter component is actually being rendered by React.

I really think this shouldn't be a problem. Let's get to the bottom of it, we'll make it run smoothly ;)

JoshMarler:
> > @nick-thompson, also, I'm wondering about the API for things like `dispatchEvent` calls from a timer. What is the expected way for a user to stop dispatching events when a bundle error occurs? i.e. in a hot-reload workflow. We no longer wrap `dispatchEvent` in a `bundleValid` check. With `ReactApplicationRoot` always catching recoverable errors how should users be disabling and re-enabling their event dispatching?
> > Should a timer be stopped in `beforeBundleEval` and restarted in `afterBundleEval` ? I guess this probably makes sense. In which case it might be worth updating the class docs to reflect this situation better.
> 
> Mhm, another great question. What do you think would be the ideal user experience here? My initial thought is that it would be cool for ReactApplicationRoot to just internally treat a `dispatchEvent` call as a no-op if the bundle is invalid. Might that cause problems in production?
> 
> I thought I could avoid the `bundleValid` flag because the errors would immediately short-circuit to the `catch` blocks, but obviously I overlooked a few cases!

For me just running a no-op is fine, an assertion could be argued over I guess. The alternative could be to pass an eval result through to `afterBundleEval` so the user can switch off a timer or whatever to avoid dispatching more things. To be honest though, a no-op feels like it makes the most sense. 

Related to this, I can't remember how we felt about `ReactApplicationRoot` catching all errors in Release build? I'm fine with it. Thought it worth raising though.

blueprint/core/blueprint_ReactApplicationRoot.h
nick-thompson: Ugh, of course, that totally slipped my mind. Very glad I have you looking over this right now :D

Yea let's add a `reset()` that keeps the duk_context alive but resets the value stack. I'll update shortly

JoshMarler:
@nick-thompson, there was a check added for a race on the bundle watcher before:

```
               // In some instances webpack rebuilds temporarily result in an empty
                    // bundle file. We do not want to trigger evaluation callbacks when the
                    // bundle file is empty as this is likely to cause errors when users attempt
                    // to call javascript functions from C/C++. Calling eval/call in duktape with
                    // an empty js file/string does not result in an error and simply pushes undefined
                    // to the top of the duktape stack.
                    if (lmt > wb.bundleLastModifiedTime && wb.bundle.loadFileAsString().isNotEmpty())
``` 

Only against empty though. It needs to be more thorough and avoid reading a partially written file if this is what is happening (suspect so as it's the only possible race here as far as I can tell)

nick-thompson:
> For me just running a no-op is fine, an assertion could be argued over I guess. The alternative could be to pass an eval result through to `afterBundleEval` so the user can switch off a timer or whatever to avoid dispatching more things. To be honest though, a no-op feels like it makes the most sense. I guess the main annoyance here is that we loose original error messages because the "red screen of doom" has it's content replaced with dispatch failures as soon as the bundle is no longer valid.

:+1: I'm with you, let's go for the no-op. I think we can avoid the dispatch errors overwriting the red screen content

> Related to this, I can't remember how we felt about `ReactApplicationRoot` catching all errors in Release build? I'm fine with it. Thought it worth raising though.

Another save! Man, seriously thank you. Yes, let's catch and show the red screen in debug, but throw in release.

blueprint/core/blueprint_ReactApplicationRoot.h
JoshMarler: Heh. No worries, glad to be of service on this end for once! Normally you correcting my hacks!

nick-thompson:
> @nick-thompson, hmmm. That might be the bundle watcher theory down the pan as I just turned it off and hit the same error

Interesting! Any chance you can get the debugger to stop on the syntax error? Or does duktape just totally bail?

JoshMarler:
> > @nick-thompson, hmmm. That might be the bundle watcher theory down the pan as I just turned it off and hit the same error
> 
> Interesting! Any chance you can get the debugger to stop on the syntax error? Or does duktape just totally bail?

Yeah I'll see if I can fire vscode debugger up and catch this.

blueprint/core/blueprint_EcmascriptEngine.h
```diff
@@ -102,68 +128,52 @@ namespace blueprint
...
         struct LambdaHelper {
-            LambdaHelper(juce::var::NativeFunction fn);
+            LambdaHelper(juce::var::NativeFunction fn, uint32_t _id);
 
             static duk_ret_t invokeFromDukContext(duk_context* ctx);
             static duk_ret_t callbackFinalizer (duk_context* ctx);
 
             juce::var::NativeFunction callback;
-            juce::Uuid                id;
+            uint32_t id;
```
JoshMarler, line 156: Hah. I was just musing, with the rate I'm pushing meter events through and creating temporaries I wonder how long it would take for this to overflow ...

nick-thompson: Yea, I thought about the same :)

My thought was basically that _most_ of the lambdas going into the release pool will be temporaries, which means that at the time the `uint32_t` wraps, we'll be fine to just start clobbering. Glad you said something though, because I'm realizing that blatantly clobbering existing entries would mean overwriting persistent functions (i.e. `registerNativeMethod` callbacks). A simple fix would be to write a `getNextId()` helper which just returns `nextId + 1` IFF the release pool is empty at that id, otherwise increment again. This, then, would only become problematic if you have so many persistent functions that you have no room left in the map for a block of temporaries (i.e. the ~30 methods of the canvas context). I think we can be reasonably sure that case would never happen!

JoshMarler: Fair enough, so just continue with a circular wrap and increment past any collisions when needed. Sounds like a plan. Heh. That feels like the sort of thing that warrants a comment inline so we understand what is going on in 6 months time!

nick-thompson: > Fair enough, so just continue with a circular wrap and increment past any collisions when needed. Sounds like a plan. Heh. That feels like the sort of thing that warrants a comment inline so we understand what is going on in 6 months time!

For sure hah. Also, realistically I should probably just benchmark this against `juce::Uuid`. If there's no significant advantage to the int then the UUID is probably easier to think about. I'm remembering that part of why I made the switch was so that I could compare `nextId` to the size of the pool, to see how many lambdas had been handed out vs how many were still allocated. But that's a debug trick, not a necessary concern in the design of this

> How do you feel about the whole `lightfunc` thing? They smell like they could be a winnder for the temporary function object's registered via `pushVarToDukStack` and `readVarFromDukStack`. I'm trying to wrap my head around whether their limitations would actually cause problems in practice here. It doesn't feel like it?

Yea, I'm with you, I just haven't taken a stab at it in here yet. Let me poke at it while I'm still in here; I think it could be worthwhile

JoshMarler: >I'm remembering that part of why I made the switch was so that I could compare nextId to the size of the pool, to see how many lambdas had been handed out vs how many were still allocated. But that's a debug trick, not a necessary concern in the design of this

Sneaky :-)

>Yea, I'm with you, I just haven't taken a stab at it in here yet. Let me poke at it while I'm still in here; I think it could be worthwhile

Yeah no worries. I'm happy to assist in anway with this. As far as I understand it the provide all the functionality we need for Blueprint's temporary function object's but require much less over head and no GC. It's fair to say this would require a new approach in regards to removing lightfuncs from the stack once the temporary objects are done with I assume.

JoshMarler: @nick-thompson, if we haven't decided on the performance improvement approach we could always pull the error handling and NativeFunction improvement out into their own PR/PRs ? I'm starting to think that might make sense as the performance stuff could be a dive down a big old rabbit hole?

nick-thompson:
Hey @JoshMarler just pushed an update addressing your feedback. I also commented out the `duk_gc` calls, and I will follow up soon with a separate PR exploring lightfuncs, and we'll move our performance related discussion there.

I found the strange eval error... I was taking a `char*` to a `juce::String` temporary when trying to `duk_pcompile` the file. Ouch, heh. Should be all good now, let me know what you think

blueprint/core/blueprint_ReactApplicationRoot.h
```diff
@@ -394,6 +420,9 @@ namespace blueprint
             JUCE_ASSERT_MESSAGE_THREAD
             jassert(handler);
 
+            // TODO: Here we have a case of `invoke` where the callback is not a temporary. Perhaps
+            // we want two different `invoke` methods: one for immeidately cleaning up temporaries,
+            // and one which persists native functions. `pinvoke`? `invokeWithPersistentNativeFunctions`?
             return engine.invoke("__BlueprintNative__.subscribe", eventType, std::move(handler));
```
JoshMarler, line 426: Funnily enough for "reasons" I've ended up not using this and instead have registered a `__MyApp__.raiseReques`.

blueprint/core/blueprint_ReactApplicationRoot.h
```diff
@@ -240,10 +250,23 @@ namespace blueprint
         {
             JUCE_ASSERT_MESSAGE_THREAD
 
+            // We early return here in the event that we're currently showing the red error
+            // screen. This prevents subsequent errors caused by dispatching events with an
+            // incorrect engine state from overwriting the first error message.
+            if (errorText)
```
JoshMarler, line 256: Cool so we're not going for a no-op here then? we just avoid the secondary error case. Meaning users need to be very careful about only dispatching events once AfterBundleEval is successful ? That's fine with me. In some ways it makes things a little more cumbersome from an API perspective, I could have various things listening to "engine" objects and dispatching events to JS in response. 

This is already sort of clear in the `ReactApplicationRoot::AfterBundleEval` docs

JoshMarler:
@nick-thompson, looks good now. I added one minor comment around the `dispatchEvent` API but I think it's all cool.

blueprint/core/blueprint_ReactApplicationRoot.h
nick-thompson: Ah interesting! Do you think we should back this 2-way event emitter out?

blueprint/core/blueprint_ReactApplicationRoot.h
nick-thompson: Hm, I guess I thought this line implemented no-op behavior? That is, if you try to dispatch events while we know the engine to be in an error state (because we set the error text and it has not yet been cleared), we just early return (no-op).

If a user dispatches before the bundle eval, I suppose this will error... do you think we should avoid that case?

My thought was basically use the `afterBundleEval` to install initial state, then afterwards just dispatch events as usual.

blueprint/core/blueprint_ReactApplicationRoot.h
JoshMarler: Yeah I'm not sure. I guess it's a more "fluxy" architecture if the user has to register specific actions. It may be worth it. I might have been guilty of suggesting the wrong thing!

blueprint/core/blueprint_ReactApplicationRoot.h
JoshMarler: Yeah that's fine with me. It might just be one of those "be aware of X" things to note in the docs. It's reasonably clear from the `ReactApplicationRoot` header docs. I've no issues with it staying like this. I think as blueprint evolves and there are more community baked "binding" solutions (i.e. ValueTree, JSON patch etc.) this sort of concern might become less of an issue.

nick-thompson: Let's keep this one going in #92, because when that merges in it will break the current `subscribe` implementation. I'll mark this resolvd for now

nick-thompson: Whoops, meant for this comment to be on the above comment on the `subscribe` method!

PR #92, opened 2020-08-21
Distinguish temporary and persistent native callbacks with lightfuncs

nick-thompson:
@JoshMarler this is a follow up to #89, adding one more commit that distinguishes temporary native callbacks (e.g. canvas context methods) from persistent native callbacks (e.g. registerNativeMethod), using a limited pool of lightfuncs for the former and our duk_finalizer approach for the latter, with duktape running it's gc whenever it deems appropriate.

After a very cursory profile of the GainPlugin, this looks to me like a 5x performance speedup on the canvas paints, which is awesomeee.

Take a look, let me know if you like the approach, and I'd also be very curious to hear a profile with your app!

### DIFF
--- a/blueprint/core/blueprint_ReactApplicationRoot.h
+++ b/blueprint/core/blueprint_ReactApplicationRoot.h
@@ -418,9 +418,6 @@ namespace blueprint
             JUCE_ASSERT_MESSAGE_THREAD
             jassert(handler);
 
-            // TODO: Here we have a case of `invoke` where the callback is not a temporary. Perhaps
-            // we want two different `invoke` methods: one for immeidately cleaning up temporaries,
-            // and one which persists native functions. `pinvoke`? `invokeWithPersistentNativeFunctions`?
             return engine.invoke("__BlueprintNative__.subscribe", eventType, std::move(handler));
         }
 
--- a/blueprint/core/blueprint_EcmascriptEngine.cpp
+++ b/blueprint/core/blueprint_EcmascriptEngine.cpp
@@ -20,6 +20,14 @@ namespace blueprint
             throw EcmascriptEngine::FatalError(msg);
         }
 
+        static juce::String getContextDump(duk_context* ctx)
+        {
+            duk_push_context_dump(ctx);
+            auto ret = juce::String(duk_to_string(ctx, -1));
+            duk_pop(ctx);
+            return ret;
+        }
+
         static void safeCall(duk_context* ctx, const int numArgs)
         {
             if (duk_pcall(ctx, numArgs) != DUK_EXEC_SUCCESS)
@@ -27,7 +35,7 @@ namespace blueprint
                 const juce::String stack = duk_safe_to_stacktrace(ctx, -1);
                 const juce::String msg = duk_safe_to_string(ctx, -1);
 
-                throw EcmascriptEngine::Error(msg, stack);
+                throw EcmascriptEngine::Error(msg, stack, getContextDump(ctx));
             }
         }
 
@@ -38,7 +46,7 @@ namespace blueprint
                 const juce::String stack = duk_safe_to_stacktrace(ctx, -1);
                 const juce::String msg = duk_safe_to_string(ctx, -1);
 
-                throw EcmascriptEngine::Error(msg, stack);
+                throw EcmascriptEngine::Error(msg, stack, getContextDump(ctx));
             }
         }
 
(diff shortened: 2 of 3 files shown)


nick-thompson:
Ok just finished a big rebase onto master after merging #89, and I went in and fixed up everything you mentioned as well. This is ready to merge, but I want to wait until we can confirm that there's no long-running lockup issue. Let me know when you get any more info on that.

I'm also going to open one more inline discussion as I mentioned in #89 on the `subscribe` functionality

blueprint/core/blueprint_ReactApplicationRoot.h
```diff
@@ -418,9 +418,6 @@ namespace blueprint
             JUCE_ASSERT_MESSAGE_THREAD
             jassert(handler);
 
-            // TODO: Here we have a case of `invoke` where the callback is not a temporary. Perhaps
-            // we want two different `invoke` methods: one for immeidately cleaning up temporaries,
-            // and one which persists native functions. `pinvoke`? `invokeWithPersistentNativeFunctions`?
             return engine.invoke("__BlueprintNative__.subscribe", eventType, std::move(handler));
```
nick-thompson, line 421: Ok, carrying on our convo from #89 here. As it is now, this commit will break this functionality because `engine.invoke` treats the callback as temporary, and deallocates it as soon as its slot in the internal temporary release pool is needed.

Based on the convo we started on the other PR, I guess the first question is probably: is this API even necessary? Because in one sense, (and I believe this is what you ended up doing @JoshMarler) if you want events from javascript, you can always just make your own `appRoot.registerNativeMethod("__MyApp__", "dispatchEventToNative"...)` and wire it up into your native code however you need.

I suppose I'd say that as we approach beta, it would help us to lean towards a minimal API, because if we announce a beta and users start working with blueprint more regularly, then introducing breaking API changes thereafter (e.g. removing something like approot.subscribe) becomes really hard. And introducing new API features is never a breaking change.

That said, we put this in originally for a reason, so obviously it has some value. A follow-up question then would be (if we decide to keep the API) how to implement it, as right now this would break

JoshMarler: Reading this on a phone so no further investigations but from experience my opinion is that this was poorly conceived request/change on my part. Initial the simplicity of this approach seemed appealing but I now feel it's API bloat. Specifying specific actions (be that a "dispatchEvent" function or something else) actually makes more sense to me and I think probably enforces more robust designs. Your native event/action handlers could take a variety of forms. If a completely generic 2-way data bind is what you need you can accomplish it very easily with the current functionality.

JoshMarler: Personally I think the sooner we pull it out the better. Maybe provide one clear foundational mechanism from native->JS communication and visa versa. Various patterns will begin to emerge in the wild no doubt and these can be documented as useful idioms as and when appropriate.

nick-thompson: Ok cool, yea, I'm with you. I think you're right that we'll see patterns emerge in the wild, and if any such pattern shows up that really starts to make sense as part of blueprint proper, we can always integrate it at that time. For now, we'll keep the API slim. I'll pull this out as part of this PR

nick-thompson:
Hey @JoshMarler have you had a chance to look into the long-running lockup issue any more?

JoshMarler:
Hey Nick, apologies I've been stuck on a couple of completely unrelated bugs. I'll be able to take a pass at this tomorrow. However, my app locked up this morning and that was using the current master so you can probably merge this down if you're not seeing similar issues on your end.

nick-thompson:
No worries! Interesting, ok, good to know. Curious if there's something else in here going awry...

JoshMarler:
Yeah for sure. I'm on leave today and not at my desk but I'll fire it up tomorrow and dig into this thing. I've been fighting obscure TS/Babel issues. Turns out authorising a TS component library that works in duktape is a bit of a pain !

nick-thompson:
> I've been fighting obscure TS/Babel issues for a few days so I've ignored this. Turns out authorising a TS component library that works in duktape is a bit of a pain !

Hmm... interesting. You've now gone the furthest (I believe) into the realm of typescript+blueprint+duktape. I wonder, leaning on your experience with it so far, should we consider leaving out Typescript integration until after we reach a solid beta milestone?

JoshMarler:
I reckon I'll have an answer for you on that in a couple of days! I literally have one problem left to solve which is basically that the Babel polyfills/transforms don't seem to be getting applied despite being present in the final bundle. I.e. Array.from.and friends. Might strike up a chat with you tomorrow on Discord just incase you have any ideas. You might be better at understand webpacks bundle output!

nick-thompson:
Sounds good!

Issue #15, opened 2019-06-26
Open question: Event support


nick-thompson:
As it stands, Blueprint doesn't really have event support. There's a non-standard implementation of pushing events to JavaScript by registering global functions in your app bundle and calling them from C++, with one special case for finding view instances and pushing events right to them.

A proper implementation would catch events on visible view elements and create something like React's SyntheticEvent, then bubble that event up the virtual React element tree. React has a library for doing exactly this, but it is not a public library and likely way too unstable to try to rely on just yet. We could attempt to implement our own, but it would likely be incomplete for a while.

Another option is to provide a global EventBridge in Blueprint, which is already kind of implemented, and have every React element that needs events subscribe to the event bridge and filter events by some known id. This skips the question of bubbling and goes more the route of something like Rxjs, but still would benefit from some kind of SyntheticEvent wrapper. In this case, we would need to find a way for a React element to access the internal id of a given view from the renderer, so that it knows what to filter on. Not sure exactly how to provide that, and I'm also questioning whether that's the right idea in the first place... seems like it could be a design flaw.

I'll spend some time studying the React Native implementation and proposing ideas here

JoshMarler:
@nick-thompson , at a point where this can be closed after the recent events work ? Or want to leave this open until we have covered all remaining events ? I.e. mouse over, mouse leave.

nick-thompson:
Yes, definitely. With the above linked issues, we have pretty well resolved the open questions here. For documentation's sake: view events are dispatched into js much like the generic `dispatchEvent` hook, and those events are caught by an event replayer which takes the `ViewId` associated with the event and looks up the `ViewInstance` in the registry, then dispatches the event to that viewInstance and bubbles it up the parent tree. There is a lot of work to be done still enumerating the different event types and totally fleshing out the SyntheticEvent interface, but we've got our approach down, so closing this higher level issue.

PR #95, opened 2020-08-30
Clean up view event dispatching (#44)

nick-thompson:
Been wanting to get to this one for a while. There's one more change I want to make that gets us closer to React's SyntheticEvent territory, but I don't want to go too far before we have the right plan in place (#15).

Also these changes are based on #92 – planning to merge that one before this.

@JoshMarler let me know if this fixes the issue you were seeing

### DIFF
--- a/blueprint/core/blueprint_View.cpp
+++ b/blueprint/core/blueprint_View.cpp
@@ -144,33 +144,37 @@ namespace blueprint
         auto w = cachedFloatBounds.getWidth();
         auto h = cachedFloatBounds.getHeight();
 
-        if (auto root = findParentComponentOfClass<ReactApplicationRoot>())
-            root->dispatchViewEvent(getViewId(), "Measure", w, h);
+        dispatchViewEvent("onMeasure", w, h);
     }
 
     void View::mouseDown (const juce::MouseEvent& e)
     {
-        if (auto root = findParentComponentOfClass<ReactApplicationRoot>())
-            root->dispatchViewEvent(getViewId(), "MouseDown", e.x, e.y);
+        dispatchViewEvent("onMouseDown", e.x, e.y);
     }
 
     void View::mouseUp (const juce::MouseEvent& e)
     {
-        if (auto root = findParentComponentOfClass<ReactApplicationRoot>())
-            root->dispatchViewEvent(getViewId(), "MouseUp", e.x, e.y);
+        dispatchViewEvent("onMouseUp", e.x, e.y);
     }
 
     void View::mouseDrag (const juce::MouseEvent& e)
     {
         auto mouseDown = e.mouseDownPosition;
-
-        if (auto root = findParentComponentOfClass<ReactApplicationRoot>())
-            root->dispatchViewEvent(getViewId(), "MouseDrag", e.x, e.y, mouseDown.x, mouseDown.y);
+        dispatchViewEvent("onMouseDrag", e.x, e.y, mouseDown.x, mouseDown.y);
     }
 
     void View::mouseDoubleClick (const juce::MouseEvent& e)
     {
-        if (auto root = findParentComponentOfClass<ReactApplicationRoot>())
-            root->dispatchViewEvent(getViewId(), "MouseDoubleClick", e.x, e.y);
+        dispatchViewEvent("onMouseDoubleClick", e.x, e.y);
+    }
+
+    bool View::keyPressed (const juce::KeyPress& key)
+    {
+        dispatchViewEvent("onKeyPress", key.getKeyCode());
+
+        // TODO: `dispatchViewEvent` should take something like a SyntheticEvent
+        // and provide methods for stopPropagation/cancelBubble, which we can
+        // use here to indicate a real return value.
+        return false;
     }
 }
--- a/blueprint/core/blueprint_View.cpp
+++ b/blueprint/core/blueprint_View.cpp
@@ -11,6 +11,44 @@
 namespace blueprint
 {
 
+    namespace detail
+    {
+
+        /** A little helper for DynamicObject construction. */
+        juce::var makeViewEventObject (const juce::NamedValueSet& props)
+        {
+            auto* o = new juce::DynamicObject();
+
+            for (auto& pair : props)
+                o->setProperty(pair.name, pair.value);
+
+            return juce::var(o);
+        }
+
+        /** Another little helper for DynamicObject construction. */
+        juce::var makeViewEventObject (const juce::MouseEvent& me)
+        {
+            // TODO: Get all of it!
+            return makeViewEventObject({
+                {"x", me.x},
+                {"y", me.y},
+                {"screenX", me.getScreenX()},
+                {"screenY", me.getScreenY()},
+            });
+        }
+
+        /** And another little helper for DynamicObject construction. */
+        juce::var makeViewEventObject (const juce::KeyPress& ke)
+        {
+            // TODO: Get all of it!
+            return makeViewEventObject({
+                {"key", juce::String(ke.getTextCharacter())},
+                {"keyCode", ke.getKeyCode()},
+            });
+        }
+
+    }
+
     //==============================================================================
     ViewId View::getViewId()
     {
@@ -38,6 +76,9 @@ namespace blueprint
                 default:     setInterceptsMouseClicks (true,  true);   break;
             }
         }
+
+        if (name == juce::StringRef("onKeyPress"))
+            setWantsKeyboardFocus(true);
 
         if (name == juce::StringRef("opacity"))
             setAlpha(static_cast<float> (value));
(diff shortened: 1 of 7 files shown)


blueprint/core/blueprint_View.h
```diff
@@ -72,6 +72,25 @@ namespace blueprint
...
+        //==============================================================================
+        /** Invokes, if exists, the respective view event handler. */
+        template <typename... Ts>
+        void dispatchViewEvent (const juce::String& eventType, Ts... args)
+        {
+            JUCE_ASSERT_MESSAGE_THREAD
+
+            if (props.contains(eventType) && props[eventType].isMethod())
+            {
+                std::vector<juce::var> vargs { args... };
```
JoshMarler, line 87: You can use std::array here if you wish:

```
std::array<juce::var, sizeof...(args)> vargs = { args... };
```

JoshMarler:
@nick-thompson, Looks great so far. Will hold out for the second drop of changes.

blueprint/core/blueprint_View.h
nick-thompson: Ah, that's nice! Will do

nick-thompson:
Ok, went a little farther than I had originally anticipated, but I ended up finding a solution that I really like for #15. The commit I just added changes view event dispatching to (1) occur at the `blueprint::View` level just by calling the correct prop key, if it exists then (2) on the javascript side we have another `ViewInstance`, and here we inject shims for view events that catch the native event object, wrap it in the correct `SyntheticEvent`, and bubble it up the view tree, so that by the time the event ever reaches user code they will only see a `SyntheticEvent` instance, and they can expect event bubbling just as in the web DOM. This allows us to drop the juce-specific `interceptClickEvents`, enabling user code that will be much more familiar and portable.

It's incomplete yet... more events to cover and have to flesh out the SyntheticEvent classes, but I really like the pattern. @JoshMarler what do you think?

examples/GainPlugin/Source/jsui/src/Meter.js
```diff
@@ -37,10 +37,10 @@ class Meter extends Component {
     });
   }
 
-  _onMeasure(width, height) {
+  _onMeasure(e) {
```
JoshMarler, line 40: Nice. One thing that would be nice here is some doc or TS definition for a MeasureEvent or anything non-standard. As I understand it, for the other event types Typescript provides typings (provided Blueprint continues to confirm to the W3C) from the dom lib which users could rely on. 

https://microsoft.github.io/PowerBI-JavaScript/interfaces/_node_modules_typedoc_node_modules_typescript_lib_lib_dom_d_.mouseevent.html

blueprint/core/blueprint_View.cpp
```diff
@@ -144,37 +185,70 @@ namespace blueprint
...
     {
-        dispatchViewEvent("onMouseUp", e.x, e.y);
+        dispatchViewEvent("onMouseUp", detail::makeViewEventObject(e));
     }
 
     void View::mouseDrag (const juce::MouseEvent& e)
     {
-        auto mouseDown = e.mouseDownPosition;
-        dispatchViewEvent("onMouseDrag", e.x, e.y, mouseDown.x, mouseDown.y);
+        // TODO: mouseDrag isn't a dom event... is it?
```
JoshMarler, line 206: Closest thing looks like https://www.w3.org/TR/DOM-Level-3-Events/#event-type-mousemove

blueprint/core/blueprint_View.cpp
```diff
@@ -11,6 +11,44 @@
...
+            auto* o = new juce::DynamicObject();
+
+            for (auto& pair : props)
+                o->setProperty(pair.name, pair.value);
+
+            return juce::var(o);
+        }
+
+        /** Another little helper for DynamicObject construction. */
+        juce::var makeViewEventObject (const juce::MouseEvent& me)
```
JoshMarler, line 29: This is neat.

examples/GainPlugin/Source/jsui/src/Meter.js
JoshMarler: Glad I held off with the components lib! This adds some significant but nice changes!

packages/juce-blueprint/esm/lib/BlueprintBackend.js
```diff
@@ -24,12 +24,73 @@ if (typeof window !== 'undefined') {
...
+  stopPropagation() {
+    this.bubbles = false;
+  }
+
+  preventDefault() {
+    this.defaultPrevented = true;
+  }
+}
+
+class SyntheticMouseEvent extends SyntheticEvent {
```
JoshMarler, line 61: Ahh OK. You've typed the events here.

packages/juce-blueprint/esm/lib/BlueprintBackend.js
```diff
@@ -24,12 +24,73 @@ if (typeof window !== 'undefined') {
...
+  const k = key.toLowerCase();
+
+  return k === 'onkeydown' ||
+    k == 'onkeyup' ||
+    k == 'onkeypress';
+}
+
+class SyntheticEvent {
+  constructor(props) {
+    this.bubbles = true;
```
JoshMarler, line 46: No idea whether it's worth making this cancelBubble for parity? https://developer.mozilla.org/en-US/docs/Web/API/Event/cancelBubble

blueprint/core/blueprint_View.cpp
```diff
@@ -144,37 +185,70 @@ namespace blueprint
...
     {
-        dispatchViewEvent("onKeyPress", key.getKeyCode());
+        dispatchViewEvent("onKeyPress", detail::makeViewEventObject(key));
+
+        // We always inform the underlying juce::Component that we've consumed the event,
+        // because we manually bubble a SyntheticEvent wrapper up the javascript view tree.
+        // However, because the React app may be only a subtree of the overall app architecture,
+        // we skip up to the ReactApplicationRoot parent and restart the keypress event propagation
+        // up there.
+        if (auto* parent = findParentComponentOfClass<ReactApplicationRoot>())
```
JoshMarler, line 224: Struggling to wrap my head around this part. Could you give a little more explanation?

JoshMarler:
> Ok, went a little farther than I had originally anticipated, but I ended up finding a solution that I really like for #15. The commit I just added changes view event dispatching to (1) occur at the `blueprint::View` level just by calling the correct prop key, if it exists then (2) on the javascript side we have another `ViewInstance`, and here we inject shims for view events that catch the native event object, wrap it in the correct `SyntheticEvent`, and bubble it up the view tree, so that by the time the event ever reaches user code they will only see a `SyntheticEvent` instance, and they can expect event bubbling just as in the web DOM. This allows us to drop the juce-specific `interceptClickEvents`, enabling user code that will be much more familiar and portable.
> 
> It's incomplete yet... more events to cover and have to flesh out the SyntheticEvent classes, but I really like the pattern. @JoshMarler what do you think?

This looks awesome Nick. Took me a while to workout what was going wrong as it's been a long time since I last thought about dom events etc. I'd be happy to drop over some TS defs for the event types. Curious how a browser compatible shim would look for those.

packages/juce-blueprint/esm/lib/BlueprintBackend.js
nick-thompson: Yea, IIRC `cancelBubble` is deprecated, and `bubbles` should be a read-only property. I'm tempted to say we can ignore adding deprecated features?

nick-thompson: Oh, jk, it is not deprecated! It's just an alias to `stopPropagation()`. Ok, yea, should update

examples/GainPlugin/Source/jsui/src/Meter.js
nick-thompson: Yea, we should definitely enumerate what these event objects carry. The goal is for the `SyntheticEvent` to roughly match the `UIEvent` spec, and the `SyntheticMouseEvent` to match `MouseEvent`, etc. So ideally we could just use the typescript defs there, but we're a long way away from parity, so I wonder if typescript would yell at us the whole time

blueprint/core/blueprint_View.cpp
nick-thompson: Yep, so basically juce does keypress event bubbling itself, walking up the `Component*` heirarchy. But we do that manually ourselves inside the juce-blueprint backend so that we can better shim the UIEvent spec, so we return `true` here to make sure juce's Component implementation stops, otherwise we'd get event bubbling twice. Now, if we just simply stopped here, any `juce::Component` starting with ReactApplicationRoot or higher will never receive keypress events that originated somewhere in the React tree, but our js bubbling will stop at the first child of the react root (because it knows of no parent higher). That means, for example, we might catch a Ctrl+D keypress on some component in the react tree, bubble it up the javascript heirarchy where maybe nothing happens, and then stop. That would be undesired, because we want ReactAppRoot to still be able to catch that and enable debugging. So this line, 224, is at a point where we've already completed our view event bubbling in js land, we don't want juce::Component to bubble the same event to the same tree twice, but we do want it to bubble starting from the ReactAppRoot.

examples/GainPlugin/Source/jsui/src/Meter.js
JoshMarler: Yeah. I think it would be fine depending on what we use. For example I've actually used to dom CanvasRenderingContext2D typing in my code and everything works fine. In this case though maybe adding definitions for `SyntheticMouseEvent` and friends makes the most sense so that we write to those interfaces. Not sure how you feel about this for the core components. I'll hold off till this is merged and then drop you a PR ASAP.

blueprint/core/blueprint_View.cpp
JoshMarler: Awesome. Makes total sense. Thanks.

nick-thompson:
> This looks awesome @nick-thompson . Took me a while to workout what was going on as it's been a long time since I last thought about dom events etc. I'd be happy to drop over some TS defs for the event types once this goes in, will need them for my base components. Curious how a browser compatible shim would look for SyntheticEvent?

Yea this one took me a while to come up with. I think it's a really solid direction though, quite happy with it. As for events, react-dom actually uses its own SyntheticEvent class, so for a browser shim the ideal move would be just to reimplement `View` in the web:

```js
function View(props) {
  if (__executingInWebEnvironment__) {
    return React.createElement('div', props, props.children);
  }
  
  // We're not in web, assume our renderer knows how to handle a 'View'
  return React.createElement('View', props, props.children);
}
```

Then we just need to catch the top level renderer:

```js
Blueprint.render(<MyApp />, rootInstance);

// Where...
function render(el, root) {
  if (__executingInWebEnvironment__) {
    return reactdom.render(el, root);
  }

  .. else do blueprint native stuff
}
```

With this, at the end of the day when you write a component using `View`, your event handlers will receive a SyntheticEvent no matter where you're running, but in web you'll get one made by ReactDOM, and in native you'll get one made by Blueprint.

There's much more to flesh out there, but that's the basic gist of it.

> Also, how far do you plan on going with this at this stage? Is the pretty much the scope of this PR? I ask as I'd like to throw over the core components ASAP once this goes in. Obviously this is a breaking change for the core components I've written so far so it would be good to get them moved up into Blueprint if you're game so that they can move with the event API changes more formally.

Yea, I'll stop here. There's a ton more to do (enumerating the DOM events and the SyntheticEvent API), but I believe this change is stable and, although breaking in API changes, provides th same feature set as before.

So if you're good with it, I'll go ahead and merge

nick-thompson:
Merged in 7defca5087f51571b85f1e7f447e6ed1bfa1eba9, forgot to update the base branch on this PR.

Issue #97, opened 2020-09-08
Remove native subscribe/unsubscribe hooks


nick-thompson:
See discussion here: https://github.com/nick-thompson/blueprint/pull/92#discussion_r475249628

This is follow up work from a previous PR, just cleaning up a now unstable/deprecated API.

nick-thompson:
Merged in #99.

PR #99, opened 2020-09-11
Move remaining JS juce-blueprint files to TS

JoshMarler:
@nick-thompson here is the beginnings of the full switch to TS. It all appears to be working with the gain example. It's got a couple breaking changes for existing code in that users will have to update their build setup to use the new babel configuration. I can add a TS example app at some point soon so we have examples of both approaches. 

Couple things left to tidy before this can be merged still but thought I'd fire over so you can start looking at everything as I'm no JS build system expert. 

- Need to remove the C++ EventBridge subscribe bindings as I've removed the JS side

- Should the dependencies set as Webpack `externals` also be peerDependencies in juce-blueprint package.json? I don't think they need to be but need a quick read up on this. 

- Need to add some sort of copy step for the `polyfill.js` file so that it is added to the `dist` directory on build. To get things working quickly I changed the GainPlugin example's `index.js` file to import from source. 

Will get all the above sorted.


examples/GainPlugin/Source/jsui/package.json
```diff
@@ -3,43 +3,21 @@
   "version": "0.1.0",
   "private": true,
   "dependencies": {
-    "@babel/runtime-corejs3": "^7.10.3",
```
JoshMarler, line 6: This may be nonsense but I had a set of configurations and dependency versions working well in my current project so I copied some stuff over for good measure.

examples/GainPlugin/Source/jsui/src/index.js
```diff
@@ -1,4 +1,4 @@
-import "juce-blueprint/cjs/lib/polyfill";
+import "juce-blueprint/src/lib/polyfill";
```
JoshMarler, line 1: Should be changed to `juce-blueprint/dist/lib/polyfill` unless there is a better way to do this? I just need to sort out copy step in webpack.

packages/juce-blueprint/babel.config.js
```diff
@@ -0,0 +1,20 @@
+module.exports =  {
+  presets: [
+    "@babel/preset-env",
+    "@babel/preset-react",
+    "@babel/preset-typescript",
+  ],
+  plugins: [
+    "@babel/plugin-proposal-class-properties",
+    "@babel/plugin-proposal-object-rest-spread",
```
JoshMarler, line 9: Not sure if the spread plugin is actually required here ... recommended in most setup tutorials.

packages/juce-blueprint/package.json
```diff
@@ -2,48 +2,37 @@
...
   "description": "BlueprintPlugin user interface in React.",
-  "main": "./cjs/juce-blueprint.js",
-  "types": "./cjs/juce-blueprint.d.ts",
-  "scripts": {
-    "build": "npm run build:types && npm run build:js",
-    "build:types": "tsc --emitDeclarationOnly --outFile ./cjs/juce-blueprint.d.ts",
-    "build:js": "babel esm --out-dir cjs --extensions \".js,.ts\"",
-    "init": "node scripts/init.js"
-  },
+  "main": "./dist/index.js",
```
JoshMarler, line 5: Is dist the best option here? Should it be `umd` ?

packages/juce-blueprint/src/components/Canvas.ts
```diff
@@ -0,0 +1,116 @@
+import React, { Component } from 'react';
```
JoshMarler, line 1: Arguably we should have some file/class description comment up at the top here. Not sure what pattern you would prefer moving forward for the individual component files?

packages/juce-blueprint/src/index.tsx
```diff
@@ -0,0 +1,45 @@
...
+export { default as NativeMethods } from './lib/NativeMethods';
+export { default as EventBridge } from './lib/EventBridge';
+
+export * from './components/View';
+export * from './components/ScrollView';
+export * from './components/Canvas';
+export * from './components/Text';
+export * from './components/Image';
+
+let __renderStarted = false;
```
JoshMarler, line 13: We could move the below out into `Blueprint.ts` and have `index.tsx` literally just be a barrel file but I think that would require users to use a named import for `Blueprint` in things like `index.js` etc.

packages/juce-blueprint/src/lib/BlueprintRenderer.ts
```diff
@@ -19,30 +23,26 @@ const HostConfig = {
    *  Really only using this and `getChildHostContext` for enforcing nesting
    *  constraints, such as that raw text content must be a child of a <Text>
    *  element.
-   *
-   *  @param {Container} rootContainerInstance
    */
-  getRootHostContext(rootContainerInstance) {
+  getRootHostContext(rootContainerInstance: ViewInstance): HostContext {
```
JoshMarler, line 27: Hopefully all the typings here are sensible ... I wasn't totally sure. TS seems happy.

packages/juce-blueprint/src/lib/BlueprintRenderer.ts
```diff
@@ -125,22 +110,20 @@ const HostConfig = {
   /** During a state change, this method will be called to identify the set of
    *  properties that need to be updated. This is more-or-less an opportunity
    *  for us to diff our props before propagating.
-   *
```
JoshMarler, line 128: I'm assuming removing the JSDoc param types makes sense here? They don't seem to add anything with the introduction of TS. Correct me if I'm wrong though @nick-thompson and I'll back this out.

packages/juce-blueprint/src/lib/SyntheticEvents.ts
```diff
@@ -0,0 +1,69 @@
...
+
+  constructor(props: any) {
+    super(props);
+
+    this.keyCode = props.keyCode;
+    this.key = props.key;
+  }
+}
+
+export default {
```
JoshMarler, line 52: Does this feel like the write approach for SyntheticEvents? Add the helpers to the `default` export?

packages/juce-blueprint/tsconfig.json
```diff
@@ -1,11 +1,19 @@
...
+    "noImplicitAny": false,
     "strict": true,
-    "esModuleInterop": true
+    "strictNullChecks": true,
+    "skipLibCheck": true,
+    "jsx": "react"
   },
-  "include": ["./esm/**/*"]
+  "include": ["./src/**/*"],
+  "exclude": ["./src/lib/polyfill.js"]
```
JoshMarler, line 18: Linters in various IDE's complained about this file without excluding polyfill.js. Maybe a .ts extension in the include pattern is better?

packages/juce-blueprint/webpack.config.js
```diff
@@ -0,0 +1,59 @@
...
+          rules: [
+            {
+              test: /\.tsx?$/,
+              exclude: [
+                path.resolve(__dirname, 'dist'),
+                path.resolve(__dirname, 'node_modules')
+              ],
+              use: [
+                {
+                  loader: "awesome-typescript-loader",
```
JoshMarler, line 36: Using ATL as it seems to be the best option from my experiments. It generates all declarations, compiles/type-checks and then transpiles through babel and it's reasonably quick.

examples/GainPlugin/Source/jsui/src/index.js
```diff
@@ -1,5 +1,3 @@
-import "juce-blueprint/cjs/lib/polyfill";
```
JoshMarler, line 1: import for `polyfill.js` has been moved to the module barrel file.

packages/juce-blueprint/package.json
```diff
@@ -2,48 +2,40 @@
...
+    "awesome-typescript-loader": "^5.2.1",
     "chalk": "^2.4.2",
     "fs-extra": "^8.1.0",
-    "typescript": "^3.9.6"
+    "typescript": "^3.9.6",
+    "webpack": "^4.27.1",
+    "webpack-cli": "^3.1.2"
   },
   "dependencies": {
+    "@babel/runtime-corejs3": "^7.11.2",
```
JoshMarler, line 28: `@babel/runtime-corejs3` is specified in the GainPlugin example dependencies. Should we make this a peer dependency to reduce bundle size etc?

packages/juce-blueprint/src/components/Canvas.ts
```diff
@@ -0,0 +1,116 @@
...
+    this._onMeasure = this._onMeasure.bind(this);
+    this._onDraw = this._onDraw.bind(this);
+
+    this.state = {
+      width: 0,
+      height: 0
+    };
+  }
+
+  _onMeasure(e: any) {
```
JoshMarler, line 84: This _onMeasure callback  was still `_onMeasure(width, height)` which caused some issues in my app after the SyntheticEvents change so I've just fixed as part of this PR.

packages/juce-blueprint/src/components/Canvas.ts
```diff
@@ -0,0 +1,116 @@
...
+      width: e.width,
+      height: e.height
+    });
+
+    if (typeof this.props.onMeasure === 'function') {
+      this.props.onMeasure(e.width, e.height);
+    }
+  }
+
+  _onDraw(ctx: any) {
```
JoshMarler, line 95: It's possible to type this against CanvasRenderingContext2D from lib dom. Worth doing now do you think? Or should we wait till we can provide the full API (or at least the majority of it)?

packages/juce-blueprint/template/src/index.js
```diff
@@ -1,5 +1,3 @@
-import "juce-blueprint/cjs/lib/polyfill";
```
JoshMarler, line 1: Once this goes in I can do a follow up PR with a TS project template. We can maybe provide templates for both TS and JS users and have juce-blueprint provide an `init-js` command and a `init-ts` command or just have the `init` command take a `useTS` flag or something. Then the `init.js` script can just switch on the template directory to copy based on that flag, i.e. `template-js` or `template-ts`. We can update the docs accordingly

JoshMarler: Maybe the set of example apps could be the existing GainPlugin for JS, some other simple plugin for TS like DelayPlugin and then the "ComponentShowcase" app which can be either.

blueprint/core/blueprint_ReactApplicationRoot.h
```diff
@@ -371,33 +371,6 @@ namespace blueprint
          **/
         BundleEvalCallback afterBundleEval;
 
-        //==============================================================================
```
JoshMarler, line 374: Removed this stuff here as easier to type EventBridge without this stuff (less ts-ignore statements etc.). We can close this once this PR is merged: https://github.com/nick-thompson/blueprint/issues/97

JoshMarler:
@nick-thompson. OK. I think this should be good for a review now. A few minor discussion points inline but otherwise everything seems to be working pretty well!

packages/juce-blueprint/src/components/Canvas.ts
JoshMarler: Added in https://github.com/nick-thompson/blueprint/pull/99/commits/3d2fe005116dfdb14b58f2a8afe2d704f0faeb3d. Let me know what you think @nick-thompson

JoshMarler: It would be great to type `View` props and avoid having to replicate `onMeasure` everywhere so that this sort of thing can be trapped a bit more easily.

packages/juce-blueprint/src/components/Canvas.ts
nick-thompson: Awesome, yea I think big +1 for typing it as `CanvasRenderingContext2D`, but that's a cool idea about the `BlueprintCanvasRenderingContext2D`. Maybe we start with the latter, and as we build out support for the remaining methods we make a plan to switch to the standard `CanvasRenderingContext2D` type def?

packages/juce-blueprint/template/src/index.js
nick-thompson: :+1: definitely, all sounds great

blueprint/core/blueprint_ReactApplicationRoot.h
nick-thompson: Oh sweet, thank you!

packages/juce-blueprint/src/components/Canvas.ts
JoshMarler: Yeah that sounds sensible to me. I leave it for the moment and add that as a follow up if that's cool. I think we probably want an `add-additional-typings` PR which brings in all the ViewProps etc.

nick-thompson:
Awesome @JoshMarler thank you! I'll pull it down and test it out locally shortly

packages/juce-blueprint/babel.config.js
nick-thompson: :+1: Happy to have it available

packages/juce-blueprint/package.json
nick-thompson: Don't think it matters really, `dist` is cool with me

packages/juce-blueprint/src/components/Canvas.ts
nick-thompson: Yes! I don't have any pattern in mind, but let's start rolling those in on a future PR

packages/juce-blueprint/src/index.tsx
nick-thompson: Yea, this is cool with me

packages/juce-blueprint/package.json
nick-thompson: Riiight, interesting. I'd be curious to see how much it actually inflates bundle size. I think let's not worry about it for now, but keep it on our radar

packages/juce-blueprint/src/index.tsx
```diff
@@ -0,0 +1,52 @@
...
+      container._rootContainer = __preferredRenderer.createContainer(container, false, false);
+    }
+
+    // Update the root Container
+    // TODO: callback in __preferredRenderer.updateContainer is not optional.
+    // @ts-ignore
+    return __preferredRenderer.updateContainer(element, container._rootContainer, null, callback);
+  },
+
+  enableMethodTrace() {
```
nick-thompson, line 44: This whole thing was really just a debug tool for myself. I use it to get details about which HostConfig methods React is calling and when. The HostConfig is now mostly implemented, so we're nearly to a point where we can just remove this. Really just think we'll need this here when upgrading React and the reconciler to support their fiber implementation and to see if we can get hooks for committing a batch of updates.

All of this is future work, just highlighting what this is for and that it should probably go away shortly.

packages/juce-blueprint/src/lib/BlueprintBackend.ts
```diff
@@ -1,12 +1,19 @@
...
 /* global __BlueprintNative__:false */
 
-let __rootViewInstance = null;
+let __rootViewInstance: ViewInstance | null = null;
 
 if (typeof window !== 'undefined') {
   // This is just a little shim so that I can build for web and run my renderer
   // in the browser, which can be helpful for debugging my renderer implementation.
+
+  //@ts-ignore
```
nick-thompson, line 14: This thing will also go away in the future when we add a proper web shim :)

packages/juce-blueprint/src/lib/BlueprintRenderer.ts
```diff
@@ -1,10 +1,14 @@
...
-import BlueprintBackend from './BlueprintBackend';
+import BlueprintBackend, {ViewInstance, RawTextViewInstance} from './BlueprintBackend';
 
 import invariant from 'invariant';
 
+type HostContext = {
+  isInTextParent: boolean;
+};
 
+//TODO: This should really be types against ReactReconciler.HostConfig with the generics typed out.
```
nick-thompson, line 11: :+1:

packages/juce-blueprint/src/lib/BlueprintRenderer.ts
nick-thompson: Definitely seems like soemthing we'll want to type against react-reconciler if we can. Otherwise we might want to just `any` all the places we touch the reconciler and try to establish our type inference above it

packages/juce-blueprint/src/lib/BlueprintRenderer.ts
nick-thompson: Yup, LGTM

packages/juce-blueprint/src/lib/NativeMethods.ts
```diff
@@ -1,8 +1,9 @@
 
+//@ts-ignore
```
nick-thompson, line 2: Yea... this file should maybe go away too. It was meant to protect methods installed from native, but the user will obviously be able to figure out for themselves if their native method installation is broken

packages/juce-blueprint/src/lib/SyntheticEvents.ts
nick-thompson: Yea, these should be implementaion details, only used when we're trying to decide what to bubble up the tree. Let's hide them from the user

packages/juce-blueprint/package.json
JoshMarler: Yeah maybe pop this down as a thing to investigate. From my point of view the only concern is reducing page load times where possible

packages/juce-blueprint/src/lib/SyntheticEvents.ts
JoshMarler: Heh. So happy as they are for now or further hiding? Can move helpers file for internal import of desired, I'm not sure what TS/JS module beat practice dictates here.

nick-thompson: Yea, we can definitely leave them here for now. In a future PR, let's pull these out into a `utils/EventHelpers.js` or some such file, then any of our internal code that needs it can just `import { isMouseEvent } from "utils/EventHelpers"` or so, and that way these helpers won't end up in the user-facing package because our barrel file won't import or re-export any of the helpers, and any of our other code that relies on them will be importing, but not exporting them.

nick-thompson:
@JoshMarler looks good to me! I've tested it out locally and everything looks good. I'm sure we'll be iterating on the minor details, but I'd love to merge now to establish the new structure. We all good to merge?

JoshMarler:
Yeah go for it. I'll run over the PR afterwards and create the issues we need to tidy the minor details up. I think the only possible thing is the measure event ? I'll be at my desk in just over an hour or so to finish off some other bits so feel free to merge or I can tidy up later this eve.

packages/juce-blueprint/src/lib/SyntheticEvents.ts
JoshMarler: Yeah totally. I'll fire up an issue for that as a minor refactor.

JoshMarler:
FYI I was thinking the component showcase app we discussed could be a typescript project which would give a good example of fully types components. Maybe that, JS GainPlugin and some other TS Plugin. Hopefully serve as a good set of starter/example apps.

JoshMarler:
@nick-thompson, Alrighty. I've removed `SyntheticMeasureEvent` and squashed everything down. Should be good to merge.

nick-thompson:
Awesome! Great work @JoshMarler thank you. I agree, I'll go ahead and close those two

PR #101, opened 2020-09-12
Clean up stashed function references when corresponding var dies

nick-thompson:
Hey @JoshMarler check this out, this is what I was talking about yesterday with cleaning up persistent references in the duktape global stash when the corresponding `var` from `readVarFromDukStack` dies. Haven't thoroughly tested it yet myself but wanted to throw this up here so you could check it out if you're interested

### DIFF
--- a/blueprint/core/blueprint_EcmascriptEngine.cpp
+++ b/blueprint/core/blueprint_EcmascriptEngine.cpp
@@ -491,23 +491,38 @@ namespace blueprint
 
                 if (duk_is_function(ctx, idx) || duk_is_lightfunc(ctx, idx))
                 {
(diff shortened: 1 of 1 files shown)


JoshMarler:
@nick-thompson awesome! Thanks for this. Looks like it should do the job. I'm just trying to clear a build error at the moment and as soon as I have that out of the way I'll give this an actual test. Looks sane to me though

JoshMarler:
@nick-thompson, FYI this is looking good. Appears to have resolved the memory leak my end! Nice one.

nick-thompson:
Great, very glad to hear it! I'll go ahead and merge now

PR #106, opened 2020-09-22
Remove juce build artefacts from example project

JoshMarler:
@nick-thompson, minor thing to remove unwanted stuff from example project.


nick-thompson:
Yea, understandable :) I keep these things in here intentionally so that an end user needs nothing other than Xcode or VSCode (and, in this case, node/npm) to get an example project up and running.

I realize it's largely cruft, but I think I'd like to keep 'em in until we make our move to juce6/CMake. Once the examples are cmake-ified we can drop all this as well as the projucer files, and generally clean up the repo a bit

JoshMarler:
Ahhh on fair enough. I hadn't appreciated it was a conscious choice. Nothing to see here!

PR #109, opened 2020-10-03
.png image support 

zigen:
related to #14
# summary
this PR allows you to show PNG image by data url.
For example `<Image source={require("./zigen.png")}/>`

Screen Shot 2020-10-03 at 21 32 00

I'll implement jpeg/gif later.

PR #110, opened 2020-10-04
jpeg/gif image support

zigen:
the rest of #109 

this PR allows showing jpeg/gif images, but the animated gif is not supported.

Screen Shot 2020-10-04 at 12 14 53

### DIFF
--- a/blueprint/core/blueprint_ImageView.h
+++ b/blueprint/core/blueprint_ImageView.h
@@ -34,7 +34,7 @@ namespace blueprint
             {
                 juce::String source = value.toString();
 
-                if (source.startsWith("data:"))
+                if (source.startsWith("data:image/"))
                 {
                     auto drawableImg = std::make_unique<juce::DrawableImage>();
                     const juce::Image img = loadImageFromDataURL(source);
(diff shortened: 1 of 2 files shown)


nick-thompson:
Awesome, thanks @zigen!

I'm reading over the code here and thinking we can probably simplify some of this if we use juce's URL class: https://docs.juce.com/master/classURL.html. We can take that string from the `source` prop, construct a URL, check if it's valid, then branch our behavior based on `getScheme()`. Where scheme == data, we can split out the mime type and branch thereupon as well. This for example opens the door to using the `getLocalFile()` method (https://docs.juce.com/master/classURL.html#a2aa2e8a4033cf040c4c6273dbf3f6cdc) as well for supporting the `file://` scheme quite easily. What do you think?

zigen:
@nick-thompson thank you for reviewing!
using `juce::URL` class seems absolutely better than my implementation.
let me rebase and fix it!

zigen:
I tried to use `juce::URL` to get the scheme from the given `source` property, but I couldn't to get the scheme because `juce:URL.getScheme` returns empty string for data url.

`juce::URL.getScheme` calls `findEndOfScheme` method. this method checks the index of `://`, but data url's format is like this `data:[mimeType],<data>`, so data url doesn't contain `://`.
https://github.com/juce-framework/JUCE/blob/master/modules/juce_core/network/juce_URL.cpp#L264
https://github.com/juce-framework/JUCE/blob/master/modules/juce_core/network/juce_URL.cpp#L360

finally `juce:URL` class cannot used for parsing data url (e.g. `"data:image/png;base64,iVBORw0KG...`)
so current my implementation seems fine and I just only rebased.

blueprint/core/blueprint_ImageView.h
```diff
@@ -34,7 +34,7 @@ namespace blueprint
             {
                 juce::String source = value.toString();
 
-                if (source.startsWith("data:"))
+                if (source.startsWith("data:image/"))
```
zigen, line 37: `juce::URL` class cannot handle data url properly because data-url doesn't contain `://` inside the url.
we can use `juce::URL` after this if-clause for `file://` or `http(s)://` protocol

JoshMarler:
@zigen, @nick-thompson , I need to add png support with the `file://` scheme for my project pretty soon. I was going to take a crack at this first thing tomorrow but want to check if you're planning on adding `file://` support as part of this merge @zigen ? If not I'll take a stab at it tomorrow. 

@nick-thompson given `juce::URL` doesn't support data URLs is this good to be merged so that I can continue on top and  support loading from the file system? Happy to look at fetching images over network as well but file support is the thing I need imminently.

zigen:
@JoshMarler I'm not planning to support `file://` scheme on this PR.

@nick-thompson could you review this again?

JoshMarler:
@zigen, OK thanks! I'll take a look at it in that case. Will hold out till this afternoon to see if Nick is going to merge this change down.

blueprint/core/blueprint_ImageView.h
nick-thompson: Urgh, that's annoying. I would love to see JUCE's URL class adhere to spec, maybe I'll open a bug with them

blueprint/core/blueprint_ImageView.h
```diff
@@ -81,20 +81,60 @@ namespace blueprint
...
             {
-                throw std::logic_error("Image failed to convert data url.");
+                throw std::runtime_error("Image failed to convert data url.");
             }
 
-            std::unique_ptr<juce::InputStream> inputStream;
-            inputStream.reset(new juce::MemoryInputStream(outStream.getData(), outStream.getDataSize(), false));
+            juce::MemoryInputStream inputStream (outStream.getData(), outStream.getDataSize(), false);
+
+            auto mimeType = source.substring(5,semiIndex);
```
nick-thompson, line 104: So this will get a little tripped up I think. The `[;base64]` token is optional in a data URL if the representation of the data is textual. For us this almost doesn't matter, except that SVGs do indeed have textual representations: `data:image/svg+xml,<svg><path>etc</path></svg>`.

My current implementation just accepts raw svg strings and passes them to the Drawable below, but that's very unusual, and I suspect users will more likely expect to pass a proper data URL even for their SVGs. Really, my current approach takes advantage of the DOM hack that SVG is valid HTML :) So we should change that to be more in line with the spec.

So, generally, I think the approach here should be (ofc this is pseudo code because we don't have a URL class like this)
```
URL u(sourceString);
if (u.scheme() == "data:") { return handleDataURL(u); }
if (u.scheme() == "file:") { return handleFileURL(u); }
... same for http/https
```
Then `handleDataURL` should search for a `;base64` token. If it finds a `;base64`, remove it, and decode the remaining data string into a byte array. Then carry on to decoding based on the mime type as you've done here already.

That all said, I understand that the goal of this PR is to support jpeg/gif images in data URLs, and the intention was not to correct my existing SVG hack :D. And with that in mind, let's go ahead and merge this, and I'll open up a second issue to handle this step for SVG Data URLs and @zigen if you're interested in tackling it, that'd be great!

blueprint/core/blueprint_ImageView.h
```diff
@@ -81,20 +81,60 @@ namespace blueprint
...
+            if (fmt == nullptr)
+            {
+                throw std::runtime_error("Unsupported format.");
+            }
 
-            auto fmt = std::make_unique<juce::PNGImageFormat>();
-            return fmt->decodeImage(*inputStream.get());
+            if (!fmt->canUnderstand(inputStream))
+            {
+                throw std::runtime_error("Cannot understand the image.");
```
nick-thompson, line 114: This is good practice, we should start adopting this more regularly. And ReactAppRoot or the ViewManager should try/catch runtime_errors and relay them back into the ecmascript engine as a javascript error. This will make sure the native app runs uninterrupted while a user is writing javascript... currently we have a lot of "if you write the wrong javascript, the native app will just crash" heh

blueprint/core/blueprint_ImageView.h
zigen: thank you for reviewing!
I totally agree that `;base64` token handling. 
we should handle other types of base64 encoded data url.

I'm interested in it, let me try it later (perhaps in this month)

nick-thompson: Sounds great!

PR #112, opened 2020-10-06
package.json tweaks after GainPlugin source depth move

olidacombe:
Hi, super glad to see cmake underway!

It seems some `package{,-lock}.json` bumps are needed after the elimination of `GainPlugin/Source`.


nick-thompson:
Nice catch, thank you!

PR #114, opened 2020-10-09
Callback helper fix

nick-thompson:
@JoshMarler per our conversation yesterday, this applies a proper fix for managing the duk_context with shared_ptrs and weak_ptrs, let me know what you think


JoshMarler:
@nick-thompson Looks good to me. I think my hot-reload bug is a separate issue which I'll take a look at so don't let me block you on this if you want to get it merged.

Issue #108, opened 2020-09-24
'Starting a new Project' documentation missing informations


nicolas-ra:
Hello, 

I have been trying to setup blueprint in a blank new project generated with Juice v6.0.1.

This section from https://nick-thompson.github.io/blueprint/#/New_Project : 

```
class MainComponent   : public Component
{
public:
    //==============================================================================
    MainComponent()
    {
        addAndMakeVisible(appRoot);
        setSize(400, 300);
    }

    ~MainComponent();

    //==============================================================================
    void paint (Graphics&) override;

    void resized() override
    {
        appRoot.setBounds(getLocalBounds());
    }

private:
    //==============================================================================
    blueprint::ReactApplicationRoot appRoot;

    JUCE_DECLARE_NON_COPYABLE_WITH_LEAK_DETECTOR (MainComponent)
};
``` 

Does not say which file it should be in. After looking in my project I came to the conclusion that it is `PluginEditor.h`.

But my file is not quite the same as the one from the documentation.
My "`MainComponent`" (which is called `NewProjectAudioProcessorEditor`) doesn't inherit from `Component` but from `juce::AudioProcessorEditor` which seems to create the following issue when integrating the code :
<img width="968" alt="Capture d’écran 2020-09-24 à 22 19 36" src="https://user-images.githubusercontent.com/49728072/94202442-f0c5f280-febd-11ea-9d34-dde0b6bc0689.png">
 
Here is the full class  (`PluginEditor.h`) generated by JUCE v6.0.1 : 

```
#pragma once

#include <JuceHeader.h>
#include "PluginProcessor.h"

//==============================================================================
/**
*/
class NewProjectAudioProcessorEditor  : public juce::AudioProcessorEditor
{
public:
    NewProjectAudioProcessorEditor (NewProjectAudioProcessor&);
    ~NewProjectAudioProcessorEditor() override;

    //==============================================================================
    void paint (juce::Graphics&) override;
    void resized() override;

private:
    // This reference is provided as a quick way for your editor to
    // access the processor object that created it.
    NewProjectAudioProcessor& audioProcessor;

    JUCE_DECLARE_NON_COPYABLE_WITH_LEAK_DETECTOR (NewProjectAudioProcessorEditor)
};
```

I would really appreciate if someone could point me to the right direction. I have strong experience in React, but I'm very new to C++.

Thanks !

nicolas-ra:
EDIT : 

[...]

PR #115, opened 2020-10-10
[Breaking] Fix evaluate() overload for non-explicit File constructor

nick-thompson:
@JoshMarler what do you think of this?

This fixes the issue we've been seeing in e.g. #108, which is caused by an implicit construction of a `juce::File` when trying to use `ReactApplicationRoot::evaluate` with a string. This is a breaking change to differentiate evaluating a file from evaluating a string, but I'm not sure of any other way to do it.

### DIFF
--- a/blueprint/core/blueprint_AppHarness.h
+++ b/blueprint/core/blueprint_AppHarness.h
@@ -70,7 +70,6 @@ namespace blueprint
          *     };
          *
          *     harness.watch(myAppBundle);
-         *     appRoot.evaluate(myAppBundle);
          *
          *     // Not strictly required if JUCE_DEBUG set.
          *     harness.start();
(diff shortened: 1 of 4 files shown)


JoshMarler:
@nick-thompson, looks ok. I'm surprised overload resolution picks 'juce::File' there though ... Clearly a standard thing I need to ready up on. I assumed the best match would be selected by compiler. The only other suggestion I have is to use 'juce::StringRef' to potentially avoid the breaking change

nick-thompson:
Iiinteresting... would StringRef solve the problem too? I'll admit, I don't really know what StringRef is for.

JoshMarler:
Basically like std::string_view, thinking about it though you'd probably hit the exact same problem heh. That was a silly suggestion. I really feel like the compiler should be preferring the juce::String overload .... Surely that's a better match !?

PR #117, opened 2020-10-21
Support file:// URL loading in Image view.

JoshMarler:
@nick-thompson, this change adds `file://` url loading support to `<Image>`. I've left image downloads as unsupported at the moment but have put a comment inline detailing possible approach. I'd be happy to add the as a follow up if you're happy for this to go in. 

My feeling is we probably want a `juce::ThreadPool` and then we can use `juce::URL::downloadToFile` with the `Image` class acting as a `DownloadTask::Listener`. Once the download tasks completes our listener `isFinished()` call can probably just do something like using `juce::MessageManager::callAsync` to assign the downloaded image to `drawable`. Let me know what you think.

### DIFF
--- a/blueprint/core/blueprint_ImageView.h
+++ b/blueprint/core/blueprint_ImageView.h
@@ -14,7 +14,6 @@
 
 namespace blueprint
 {
-
     //==============================================================================
     /** The ImageView class is a core view for drawing images within Blueprint's
         layout system.
@@ -23,6 +22,9 @@ namespace blueprint
     {
     public:
         //==============================================================================
+        static inline juce::Identifier sourceProp = "source";
+
+        //==============================================================================
         ImageView() = default;
 
         //==============================================================================
@@ -30,26 +32,32 @@ namespace blueprint
         {
             View::setProperty(name, value);
 
-            if (name == juce::Identifier("source"))
+            if (name == sourceProp)
             {
-                juce::String source = value.toString();
+                const juce::String source    = value.toString();
+                const juce::URL    sourceURL = source;
 
-                if (source.startsWith("data:image/"))
+                if (sourceURL.isWellFormed())
                 {
                     auto drawableImg = std::make_unique<juce::DrawableImage>();
-                    const juce::Image img = loadImageFromDataURL(source);
-                    drawableImg->setImage(img);
+                    drawableImg->setImage(loadImageFromURL(sourceURL));
                     drawable = std::move(drawableImg);
-                    return;
                 }
-
-                // handle svg image
-                drawable = std::unique_ptr<juce::Drawable>(
-                    juce::Drawable::createFromImageData(
-                        source.toRawUTF8(),
-                        source.getNumBytesAsUTF8()
-                    )
-                );
+                else if (source.startsWith("data:image/")) // juce::URL does not currently handle Data URLs
+                {
+                    auto drawableImg = std::make_unique<juce::DrawableImage>();
+                    drawableImg->setImage(loadImageFromDataURL(source));
+                    drawable = std::move(drawableImg);
+                }
+                else // If not a URL treat source prop as inline SVG/Image data
+                {
+                    drawable = std::unique_ptr<juce::Drawable>(
+                            juce::Drawable::createFromImageData(
+                                    source.toRawUTF8(),
+                                    source.getNumBytesAsUTF8()
+                            )
+                    );
+                }
             }
         }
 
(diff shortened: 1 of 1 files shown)


blueprint/core/blueprint_ImageView.h
```diff
@@ -30,26 +32,32 @@ namespace blueprint
...
-                );
+                else if (source.startsWith("data:image/")) // juce::URL does not currently handle Data URLs
+                {
+                    auto drawableImg = std::make_unique<juce::DrawableImage>();
+                    drawableImg->setImage(loadImageFromDataURL(source));
+                    drawable = std::move(drawableImg);
+                }
+                else // If not a URL treat source prop as inline SVG/Image data
+                {
+                    drawable = std::unique_ptr<juce::Drawable>(
```
nick-thompson, line 54: Eventually this case should go away and this `else` block should probably be a `runtime_error`. We'll move SVG data up into the `data:` URL category

blueprint/core/blueprint_ImageView.h
```diff
@@ -58,40 +66,81 @@ namespace blueprint
...
+                    const juce::String errorString = "Unable to load image file: " + imageFile.getFullPathName();
+                    throw std::logic_error(errorString.toStdString());
+                }
+
+                return image;
+            }
+
+            if (url.isProbablyAWebsiteURL(url.toString(false)))
+            {
+                //TODO: What approach should we take here?
```
nick-thompson, line 109: Yeaa good thoughts here. I like the ThreadPool idea, thinking we could basically just do something like

```
tp.addJob([this, url]() -> void {
  juce::MemoryBlock mb;
  url.readEntireBinaryStream(mb);

  // Will have to check the URL to see which image format to use I think?
  auto image = juce::ImageFileFormat::loadFrom(mb.data(), mb.size());

  // Will have to grab a mutex here
  drawable = std::move(image):
});
```

Then there's the question of who owns the thread pool? Each `Image` could have its own with 1 thread, but that seems like it would scale poorly. Another option is for ReactAppRoot to carry 1 threadpool (4 threads or something) and for all Image classes to `findParentComponentOfType<ReactAPplicationRoot>()->addThreadPoolJob(...etc)`. If that latter idea is easy enough to pull off, I think that scales well.

Lastly, placeholder– I'm not sure we should do anything. I think the user has enough control here to do what they like, for example:

```
<View>
  <Image source="file:///path/to/placeholder.png" {...styles.image} />
  <Image source="http://www.slowloadingimage.com/x" {...styles.image} />
</View>
```
In this example, the `file://` image will show up immediately, and will show until the network one draws over it. We could extend this by dispatching a `load` event on `Image` when the network request succeeds. This is a known pattern in web (https://idiallo.com/javascript/img-detect-loaded, couldn't find the spec in my brief search), and so the user here could render two images (placeholder + network image), wait for the load event, and when handling the load event remove the placeholder image from its tree:

```
tp.addJob([this, url]() -> void {
  juce::MemoryBlock mb;
  url.readEntireBinaryStream(mb);

  // Will have to check the URL to see which image format to use I think?
  auto image = juce::ImageFileFormat::loadFrom(mb.data(), mb.size());

  juce::MessaageManager::callAsync([this, img = std::move(image)]() {
    drawable = std::move(image):
    if (props.hasProperty("onload") && props["onload"].isNativeFunction()) {
      props["onload"]();
    }
  });
});
```

nick-thompson (approved):
Awesome @JoshMarler this is great. I'm happy to merge as is, but take a look at the comments below and let me know if you want to take a stab at the network loading

blueprint/core/blueprint_ImageView.h
JoshMarler: I'll run over your earlier comments regarding this at the weekend and see if I can tidy that up.

blueprint/core/blueprint_ImageView.h
JoshMarler: @nick-thompson , Yeah totally, I almost suggested and event/callback on `Image` rather than placeholder prop. I think that's nicer. 

Yeah `ThreadPool` seems the way to go and mutex rather than `MessageManager::callAsync` makes more sense! (I almost never use the juce threading primitives as we have our own monstrous threading system). 

`ReactApplicationRoot` owning the `ThreadPool` seems reasonable. I guess it depends if we envisage other use cases for the pool in future otherwise we could always consider a static `ThreadPool` instance on `Image` as `Image::downloadPool` or something?

If you're happy to merge this now that would be awesome as I'd could use it for some things. I'll tackle the file download in a separate PR (probs on Saturday as in the middle of other things).

nick-thompson: > If you're happy to merge this now that would be awesome as I'd could use it for some things. I'll tackle the file download in a separate PR (probs on Saturday as in the middle of other things).

Yea I'll go ahead and merge now!

> ReactApplicationRoot owning the ThreadPool seems reasonable. I guess it depends if we envisage other use cases for the pool in future, otherwise we could always consider a static ThreadPool instance on Image as Image::downloadPool or something? I haven't actually checked whether a static ThreadPool is a good idea/safe mind you.

Good question... I'm kind of hesitant to use static instances of anything because I want blueprint to comfortably & safely work within the context of plugins in a DAW. That miiiiight be safe for this particular case anyway, but I guess I've kinda developed the habit of "no static instances anywhere" as a plugin author

> Yeah ThreadPool seems the way to go and mutex rather than MessageManager::callAsync makes more sense! (I almost never use the juce threading primitives as we have our own monstrous threading system).

Yea I like the mutex, but actually if we're going to invoke the `props["onload"]` callback that will have to happen on the message thread anyway. So, might need the `callAsync` either way

JoshMarler: Ahh yeah wasn't thinking. ThreadPool between multiple instances might not be very smart 😆

PR #118, opened 2020-10-22
Add simple Button component

JoshMarler:
@nick-thompson , here is a very simple `Button` component conforming to the styles we discussed a while back. Thought it would be nice to get this in. If you're happy with the general approach I will get `Slider` sent over next and then update the Gain example to use the new slider with a `Canvas` draw routine and I can add a simple `Button` example which just bypasses the effect or mutes etc.

### DIFF
--- a/blueprint/core/blueprint_View.cpp
+++ b/blueprint/core/blueprint_View.cpp
@@ -252,7 +252,15 @@ namespace blueprint
         if (auto *parent = findParentComponentOfClass<ReactApplicationRoot>())
         {
             std::vector<juce::var> vargs { getViewId(), eventType, e };
-            parent->engine.invoke("__BlueprintNative__.dispatchViewEvent", vargs);
+
+            try
+            {
+                parent->engine.invoke("__BlueprintNative__.dispatchViewEvent", vargs);
+            }
+            catch (const EcmascriptEngine::Error& err)
+            {
+                parent->handleRuntimeError(err);
+            }
         }
     }
 
--- a/blueprint/core/blueprint_View.cpp
+++ b/blueprint/core/blueprint_View.cpp
@@ -13,20 +13,37 @@ namespace blueprint
 
     namespace detail
     {
+        juce::var getMouseEventRelatedTarget(const juce::MouseEvent& e, const blueprint::View& view)
+        {
+            juce::Component *topParent              = view.getTopLevelComponent();
+            const juce::MouseEvent topRelativeEvent = e.getEventRelativeTo(topParent);
+
+            juce::Component *componentUnderMouse = topParent->getComponentAt(topRelativeEvent.x, topRelativeEvent.y);
+
+            if (auto v = dynamic_cast<blueprint::View*>(componentUnderMouse))
+                return v->getViewId();
+
+            // return null relatedTarget if event occurred from a non-View component.
+            return {};
+        }
 
         /** A little helper for DynamicObject construction. */
-        juce::var makeViewEventObject (const juce::NamedValueSet& props)
+        juce::var makeViewEventObject (const juce::NamedValueSet& props, const blueprint::View& view)
         {
             auto* o = new juce::DynamicObject();
 
             for (auto& pair : props)
                 o->setProperty(pair.name, pair.value);
 
+            o->setProperty("target", view.getViewId());
+
+            //TODO: Add timeStamp.
+
             return juce::var(o);
         }
 
         /** Another little helper for DynamicObject construction. */
-        juce::var makeViewEventObject (const juce::MouseEvent& me)
+        juce::var makeViewEventObject (const juce::MouseEvent& me, const blueprint::View &view)
         {
             // TODO: Get all of it!
             return makeViewEventObject({
@@ -34,28 +51,29 @@ namespace blueprint
                 {"y", me.y},
                 {"screenX", me.getScreenX()},
                 {"screenY", me.getScreenY()},
-            });
+                {"relatedTarget", getMouseEventRelatedTarget(me, view)}
+            }, view);
         }
 
         /** And another little helper for DynamicObject construction. */
-        juce::var makeViewEventObject (const juce::KeyPress& ke)
+        juce::var makeViewEventObject (const juce::KeyPress& ke, const blueprint::View &view)
         {
             // TODO: Get all of it!
             return makeViewEventObject({
                 {"key", juce::String(ke.getTextCharacter())},
                 {"keyCode", ke.getKeyCode()},
-            });
+            }, view);
         }
 
     }
 
     //==============================================================================
-    ViewId View::getViewId()
+    ViewId View::getViewId() const
     {
         return juce::DefaultHashFunctions::generateHash(_viewId, INT_MAX);
     }
 
-    juce::Identifier View::getRefId()
+    juce::Identifier View::getRefId() const
     {
         return _refId;
     }
@@ -188,33 +206,33 @@ namespace blueprint
         dispatchViewEvent("onMeasure", detail::makeViewEventObject({
             {"width", w},
             {"height", h}
-        }));
+        }, *this));
     }
 
     void View::mouseDown (const juce::MouseEvent& e)
     {
-        dispatchViewEvent("onMouseDown", detail::makeViewEventObject(e));
+        dispatchViewEvent("onMouseDown", detail::makeViewEventObject(e, *this));
     }
 
     void View::mouseUp (const juce::MouseEvent& e)
     {
-        dispatchViewEvent("onMouseUp", detail::makeViewEventObject(e));
+        dispatchViewEvent("onMouseUp", detail::makeViewEventObject(e, *this));
     }
 
     void View::mouseDrag (const juce::MouseEvent& e)
     {
         // TODO: mouseDrag isn't a dom event... is it?
-        dispatchViewEvent("onMouseDrag", detail::makeViewEventObject(e));
+        dispatchViewEvent("onMouseDrag", detail::makeViewEventObject(e, *this));
     }
 
     void View::mouseDoubleClick (const juce::MouseEvent& e)
     {
-        dispatchViewEvent("onMouseDoubleClick", detail::makeViewEventObject(e));
+        dispatchViewEvent("onMouseDoubleClick", detail::makeViewEventObject(e, *this));
     }
 
     bool View::keyPressed (const juce::KeyPress& key)
     {
-        dispatchViewEvent("onKeyPress", detail::makeViewEventObject(key));
+        dispatchViewEvent("onKeyPress", detail::makeViewEventObject(key, *this));
 
         // We always inform the underlying juce::Component that we've consumed the event,
         // because we manually bubble a SyntheticEvent wrapper up the javascript view tree.
@@ -231,23 +249,10 @@ namespace blueprint
     {
         JUCE_ASSERT_MESSAGE_THREAD
 
-        if (props.contains(eventType) && props[eventType].isMethod())
+        if (auto *parent = findParentComponentOfClass<ReactApplicationRoot>())
         {
-            std::array<juce::var, 1> vargs { e };
-            juce::var::NativeFunctionArgs nfArgs (juce::var(), vargs.data(), static_cast<int>(vargs.size()));
-
-            try
-            {
-                std::invoke(props[eventType].getNativeFunction(), nfArgs);
-            }
-            catch (const EcmascriptEngine::Error& err)
-            {
-                if (auto* parent = findParentComponentOfClass<ReactApplicationRoot>())
-                    return parent->handleRuntimeError(err);
-
-                // If we coudln't find a parent that can handle it, rethrow
-                throw err;
-            }
+            std::vector<juce::var> vargs { getViewId(), eventType, e };
+            parent->engine.invoke("__BlueprintNative__.dispatchViewEvent", vargs);
         }
     }
 
--- a/examples/GainPlugin/PluginProcessor.cpp
+++ b/examples/GainPlugin/PluginProcessor.cpp
@@ -28,7 +28,12 @@ AudioProcessorValueTreeState::ParameterLayout createParameterLayout()
                 return String(Decibels::gainToDecibels(value), 1) + "dB";
             },
             nullptr
-        )
+        ),
+        std::make_unique<AudioParameterBool>(
+            "MainMute",
+            "Mute",
+            false
+       )
     );
 
     return params;
@@ -163,6 +168,13 @@ void GainPluginAudioProcessor::processBlock (AudioBuffer<float>& buffer, MidiBuf
     // Our intense dsp processing
     gain.setValue(*params.getRawParameterValue("MainGain"));
     gain.applyGain(buffer, buffer.getNumSamples());
+
+    if (auto *muteParam  = dynamic_cast<AudioParameterBool*>(params.getParameter("MainMute")))
+    {
+        bool muted = muteParam->get();
+        if (muted)
+            buffer.applyGain(0.0f);
+    }
 
     // We'll also report peak values for our meter. This isn't an ideal way to do this
     // as the rate between the audio processing callback and the timer on which the
--- a/examples/GainPlugin/jsui/src/App.js
+++ b/examples/GainPlugin/jsui/src/App.js
@@ -3,11 +3,12 @@ import Label from './Label';
 import Meter from './Meter';
 import React, { Component } from 'react';
 import Slider from './Slider';
+import ParameterToggleButton from './ParameterToggleButton'
 import {
   Canvas,
   Image,
-  View,
   Text,
+  View,
 } from 'react-juce';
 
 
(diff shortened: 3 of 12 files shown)


packages/juce-blueprint/src/components/Button.tsx
```diff
@@ -0,0 +1,73 @@
...
+
+        setDown(true);
+    }
+
+    const handleUp = (e: SyntheticMouseEvent) => {
+        if (typeof onMouseUp === 'function')
+            onMouseUp.call(null, e);
+
+        setDown(false);
+        onClick();
```
nick-thompson, line 59: Interesting here, I'm just thinking of this now. I don't think `Button` should be the one sending the `onClick` up the tree. I think we could just remove the handling of `onClick` altogether here.

That said, the `ViewInstance` inside the js renderer should send click events with a SyntheticMouseEvent as in here: https://developer.mozilla.org/en-US/docs/Web/API/Element/click_event (see https://github.com/nick-thompson/blueprint/blob/master/packages/juce-blueprint/src/lib/BlueprintBackend.ts#L99)

For now, want to just remove the handling of `onClick` here and queue the ViewInstance work for later? Or would it be helpful to tackle now?

JoshMarler: @nick-thompson, I think it makes sense to add this now and I see your point. 

So essentially we just want to switch on `onMouseUp` as the event in the `if (SyntheticEvents.isMouseEventHandler(propKey))` check and then check if `onClick` exists as a prop and call ? 

i.e. sit closer to React's proper SynteticEvent behaviour?

JoshMarler: Oh wait it doesn't even need to do that does it. C++ should send the click event up through that chain. Fool.

JoshMarler: Out of interest. Should our view event names match React's ?

`onClick onContextMenu onDoubleClick onDrag onDragEnd onDragEnter onDragExit
onDragLeave onDragOver onDragStart onDrop onMouseDown onMouseEnter onMouseLeave
onMouseMove onMouseOut onMouseOver onMouseUp`

i.e. we use `onMouseDoubleClick` should we be using `onDoubleClick` ?

nick-thompson: Yea, let's definitely aim for parity with React's event names. And now is the time for breaking changes :) Glad we're doing this now, it'll be nice to minimize expected breaking changes once we announce beta.

> So essentially we just want to switch on onMouseUp as the event in the if (SyntheticEvents.isMouseEventHandler(propKey)) check and then check if onClick exists as a prop and call ?
>
> i.e. sit closer to React's proper SynteticEvent behaviour?

Exactly. With `onClick` though there's slightly more complicated behavior. I'll elaborate a bit here on how this part works to hopefully clarify why this one's complicated and what we can do about it. Feel free to skip over this paragraph if you already know it :). So, blueprint ships only a few core components (View, Image, Canvas, ScrollView... etc), and all of these components natively derive from `blueprint::View`. So, no matter what your application looks like, by the time it hits the BlueprintRenderer, pretty much everything is just a `View`. To track the rendered instances, we make this `ViewInstance` class internally. Here, when we see that you're trying to supply an `onMouseDown`, handler, we hijack it. We put our _own_ `onMouseDown` handler over the native boundary, and hang onto the one you gave us. Then, from the native side, we listen for juce's Component mouse callbacks and try to invoke the `onMouseDown` that we handed over to native: https://github.com/nick-thompson/blueprint/blob/master/blueprint/core/blueprint_View.cpp#L196. When that happens, we catch the call, run the `onMouseDown` that you gave us, but then apply our own behavior for bubbling this up the `ViewInstance` tree, as you see in React/React-Native and normal web DOM stuff. So in the naive case, if juce had an `onClick` method of `juce::Component`, we could just run the same play to wire this up for our `onClick` prop.

Now the reason this one in particular is complicated, is because an `onClick` is (according to the spec) only fired when we've seen both a mouse down _and_ a mouse up. Imagine your app running on your desktop, clicking somewhere on your desktop (not on your app), holding the button down, moving your mouse over your app, and releasing the mouse button. We'll fire a mouseUp, but not a mouseDown (we didn't see the mouseDown, your desktop did). So we should not fire an onClick. More complicated: per the spec, we should fire an onClick from a parent `ViewInstance` even when one of the children saw the mouse down.

So one way we could implement this is to give _every_ `ViewInstance` a default `onMouseDown`/`onMouseUp` and let them each do their own bookkeeping to see if they should fire an `onClick`. That could get costly, and we really don't want all those functions getting set onto the native side. Another way would be to create a top level "book keeper," maintain our existing functionality, but have our `mouseDown`/`mouseUp` handlers tell the book keeper "hey I just got a mouse down/up." Then the book keeper identifies when to run an `onClick`, and invokes its own bubble mechanism up the appropriate `ViewInstance` subtree. This would be the preferred way, imo, because it creates the least overhead. (FWIW this is, as far as I've discerned from a very quick look over react's internal implementation, how React's synthetic event library works too: with a top level listener and this notion of "replaying" events over the correct child nodes– https://github.com/facebook/react/blob/master/packages/react-dom/src/events/ReactDOMEventReplaying.js)

What do you think?

JoshMarler: @nick-thompson , So I'd like to try tackling this soon. 

I just want to clarify my understanding on this. 

Basically we're saying that in `ViewInstance` if a `setProperty` call comes through for a mouse event then we can check if that event is `onMouseDown` / `onMouseUp` and then call a function to register that event for the given `ViewInstance` with some `EventBookKeeper` class ?

I.e. 

```
// in setProperty
if (SyntheticEvents.isMouseEventHandler([propKey)) {
  if (isMouseUpOrDown(propKey)) {
    eventBookKeeper.notify(this._id, propKey);
  }
}
```

In the above `eventBookKeeper` (or whatever we call it) would store the previous id/event pair and on notify if the previous event belongs to the same view instance fetch the instance from the root container and call `instance.bubbleViewEvent` (or something along those lines). 

I'm sure the event tracking will get more complex than this as we go along and we'll need to define how something like a `mouseDrag` within a single view followed by a release should behave etc (imagining some sort of id -> event stack map or something). 

Is this crude description at all in line with what you're describing?

nick-thompson: Yea, so this is indeed fairly tricky.

In general you have the right idea: if we equip every single `ViewInstance` with a default handler (currently we're using that noop function https://github.com/nick-thompson/blueprint/blob/master/packages/juce-blueprint/src/lib/BlueprintBackend.ts#L50), and use that wrapped callback to notify our bookkeeper, then the bookkeeper will be able to definitively say "we saw a mouse down over here, and a mouseup over here, which means we should run a click event from here up." In that case, yea, I think the code snippet you posted above is about right.

Off the bat, I wonder if this is an expensive way to do it. For example, in a large application, do we really want default noop functions for several mouse events on _every_ view instance just so that we can bubble? (Remember, all of those prop functions go over the native boundary). Or would it perform better to have a more generic "dispatchViewEvent" that goes directly to the bookkeeper, who is then responsible for replaying the event up the tree from the child emitting the event. That would simplify the bookkeeping and bubbling too, since we don't need to wrap every prop function with some wrapper code for bubbling. This is actually how the view event stuff was first written, and I scooped it out when prop functions became supported. Now that we're considering proper event bubbling, I wonder if that was a wrong move :)

Anyways! I think we should start with the proposed method two paragraphs up, with the goal of just correctly satisfying the onclick event. If we have that behavior correct, then user code would never have to change if we reimplemented the view stuff along the lines of the proposal one paragrah up.

JoshMarler: @nick-thompson , Yeah sorry just whacked that example in without thinking but of course it wouldn't happen in that exact location as you'd have to persist some callback/handler. Gotcha. 

Having `dispatchViewEvent` push events into some "book keeping" class does sound like a sensible way to do this. I agree on the overhead concerns and I'm all for anything that helps to reduce memory usage inside duktape etc. 

Agree'd though. I can implement the first approach just to satisfy `onClick` and then allow us to get `Button` and the like into the core components. 

I'll take a quick look over the `dispatchViewEvent` approach just in case it looks like something I can get out of the way in a reasonable amount of time!

examples/GainPlugin/PluginProcessor.cpp
```diff
@@ -28,158 +28,170 @@ AudioProcessorValueTreeState::ParameterLayout createParameterLayout()
...
+//==============================================================================
+GainPluginAudioProcessor::GainPluginAudioProcessor()
      : AudioProcessor (BusesProperties()
                        .withInput  ("Input",  AudioChannelSet::stereo(), true)
                        .withOutput ("Output", AudioChannelSet::stereo(), true)),
-       params(*this, nullptr, JucePlugin_Name, createParameterLayout())
-{
-}
-
-GainPluginAudioProcessor::~GainPluginAudioProcessor()
```
JoshMarler, line 46: This file seemed to be the only one lurking with dos/windows line endings and my editor has converted it. Let me know if that's an issue. Everything else seems to use unix line endings.

examples/GainPlugin/PluginProcessor.cpp
```diff
@@ -28,158 +28,170 @@ AudioProcessorValueTreeState::ParameterLayout createParameterLayout()
...
+    // this code if your algorithm always overwrites all the output channels.
+    for (auto i = totalNumInputChannels; i < totalNumOutputChannels; ++i)
+        buffer.clear (i, 0, buffer.getNumSamples());
 
     // Our intense dsp processing
-    gain.setValue(*params.getRawParameterValue("MainGain"));
+    gain.setValue(*params.getRawParameterValue("MainGain"));
     gain.applyGain(buffer, buffer.getNumSamples());
 
+    if (auto *muteParam  = dynamic_cast<AudioParameterBool*>(params.getParameter("MainMute")))
```
JoshMarler, line 172: This is the only significant change. Adding the MainMute param and the wild dsp.

blueprint/core/blueprint_View.cpp
```diff
@@ -13,49 +13,67 @@ namespace blueprint
...
+        juce::var makeViewEventObject (const juce::NamedValueSet& props, const blueprint::View& view)
         {
             auto* o = new juce::DynamicObject();
 
             for (auto& pair : props)
                 o->setProperty(pair.name, pair.value);
 
+            o->setProperty("target", view.getViewId());
+
+            //TODO: Add timeStamp.
```
nick-thompson, line 40: We could also trivially add the timestamp in js if that's easier

blueprint/core/blueprint_View.cpp
```diff
@@ -231,23 +249,10 @@ namespace blueprint
...
-            catch (const EcmascriptEngine::Error& err)
-            {
-                if (auto* parent = findParentComponentOfClass<ReactApplicationRoot>())
-                    return parent->handleRuntimeError(err);
-
-                // If we coudln't find a parent that can handle it, rethrow
-                throw err;
-            }
+            std::vector<juce::var> vargs { getViewId(), eventType, e };
+            parent->engine.invoke("__BlueprintNative__.dispatchViewEvent", vargs);
```
nick-thompson, line 255: We should still catch the error here right?

blueprint/core/blueprint_View.cpp
JoshMarler: Yeah totally. Should be a separate PR but whatever works best. Ideally the most performant of the two perhaps so that we can keep the overhead of event creation as low as possible.

examples/GainPlugin/PluginProcessor.cpp
nick-thompson: Yeeaaaa :) nice, thanks, definitely prefer unix line endings

blueprint/core/blueprint_View.cpp
JoshMarler: Ah yeah my mistake.

examples/GainPlugin/jsui/src/App.js
```diff
@@ -3,11 +3,12 @@ import Label from './Label';
 import Meter from './Meter';
 import React, { Component } from 'react';
 import Slider from './Slider';
+import ParameterToggleButton from './ParameterToggleButton'
 import {
   Canvas,
   Image,
-  View,
   Text,
+  View
```
nick-thompson, line 11: 1000% nit pick, but I prefer trailing commas everywhere in javascript. The reason is simple: trailing commas aren't a problem in js, and here for example with one import name per line, adding a _new_ import after View would involve touching the `View` line as well to add the necessary comma. That means touching 2 lines where you only needed 1, which makes the git diff slightly more complicated. For longer lines, this effect gets more impactful.

Again, supreme nitpick, but that's why you'll see me write trailing commas everywhere :)

packages/react-juce/src/components/Button.tsx
```diff
@@ -0,0 +1,104 @@
...
+    super(props);
+    this._ref = React.createRef();
+
+    this.state = {
+      down: false
+    }
+  }
+
+  handleDown = (e: SyntheticMouseEvent): void => {
+    if (this.props.onMouseDown && typeof this.props.onMouseDown === 'function')
```
nick-thompson, line 62: This can just be `if (typeof this.props.onMouseDown === 'function')`. If `this.props.onMouseDown` does not exist, then `typeof this.props.onMouseDown` will not === 'function'

packages/react-juce/src/components/Button.tsx
```diff
@@ -0,0 +1,104 @@
...
+      this.props.onMouseDown.call(null, e);
+
+    this.setState({
+      down: true
+    });
+  }
+
+  handleUp = (e: SyntheticMouseEvent): void => {
+    if (this.props.onMouseUp && typeof this.props.onMouseUp === 'function')
```
nick-thompson, line 71: Same

packages/react-juce/src/components/Button.tsx
```diff
@@ -0,0 +1,104 @@
...
+      this.props.onMouseUp.call(null, e);
+
+    this.setState({
+      down: false
+    });
+
+    if (this.props.onClick && typeof this.props.onClick === 'function') {
+      const instance = this._ref ? this._ref.current : null;
+
+      if (instance && instance.contains(e.relatedTarget)) {
```
nick-thompson, line 81: This ends up looking quite nice!!

packages/react-juce/src/components/Button.tsx
JoshMarler: Cool. Might have done that a couple times actually so will clean up.

packages/react-juce/src/lib/BlueprintBackend.ts
```diff
@@ -146,30 +160,100 @@ export class RawTextViewInstance {
...
+    }
+
+    if (eventType === "onMouseUp") {
+      __bubbleEvent(instance, eventType, event);
+
+      if (__lastMouseDownViewId && viewId === __lastMouseDownViewId) {
+        __bubbleEvent(instance, "onClick", event);
+      }
+
+      __lastMouseDownViewId = null;
```
nick-thompson, line 231: Hm... should this be inside the `if` block immediately preceding?

nick-thompson (approved):
@JoshMarler this is fantastic. Tiny things inline, but I'm quite ready to merge

packages/react-juce/src/lib/BlueprintBackend.ts
JoshMarler: Yeah arguably should be. Will tidy that up.

JoshMarler: I'm gonna go with your suggestion being right for the moment given the juce docs: 

`A mouseUp callback is sent to the component in which a button was pressed even if the mouse is actually over a different component when the button is released.`

Hopefully that holds in all cases otherwise there's a slight gotcha there if `__lastMouseDownViewId` isn't equal to `viewId`. I think it's correct in practice though. We just might need to keep an eye out if any weird edge case gets reported.

JoshMarler:
@nick-thompson, OK I've bumped the package.json versions and addressed the minor review comments. Let me know if that's all good and I can rebase if you'd prefer. Let me know if anything needs to be done in regards to the npm package update. If needed I can split the event work to a separate PR first so that the package update can go in and then the GainPlugin change with the updated dependency can follow. Whatever is easiest.

blueprint/core/blueprint_View.cpp
JoshMarler: Ah. Still have this left to do. 
Will sort that out.

JoshMarler: Just to clarify. We're talking about wrapping `engine.invoke` in `try/catch` here right?

JoshMarler: Should be sorted now

nick-thompson: Hm... maybe I'm missing something? I don't see that change

nick-thompson: Whoops, jk, I had gotten into a weird diff state on github. All good!

nick-thompson:
Merged in 3ebdf5d. Had to ammend the latest commit to bump the package.jsons again because I messed up the `npm publish`, my bad!

PR #119, opened 2020-10-22
Add documentation for debugger setup

JoshMarler:
@nick-thompson , Here's a docs addition for setting up VSCode debugger. I need to add one quick thing before it gets merged which is the tweak/hack you need to make to get this work on Windows at the moment. I hope to look at that properly at some point and submit an actual fix to the author or officially fork the extension (bit hesitant to do that). 

Anyway pushed so you can have a quick read over. We can probably add some little bit to the docs for using the standard duktape debugger interface but it my limited experience with it the dev experience was pretty bad.


JoshMarler:
I just need to fire up windows and sanity check the hack actually works still after all the TS changes and the like.

nick-thompson:
Excellent, looks great, thanks @JoshMarler! Let me know when you've checked the windows side, I'm ready to merge otherwise

JoshMarler:
@nick-thompson , OK this should be good to go. Added points for Windows in the Known Issues section. Debugging works but for now it requires a patch to the extension on my fork: 

https://github.com/JoshMarler/vscode-duktape-debug/tree/topic/fix-windows-debug-blueprint

nick-thompson:
Great! 🙌

PR #121, opened 2020-10-25
Fix bug with juce::URL parsing causing GainPlugin to crash.

JoshMarler:
@nick-thompson, this fixes the issue with the new file URL loading code that was causing the GainPlugin example to crash. `juce::URL::isWellFormed` is currently incomplete.

### DIFF
--- a/blueprint/core/blueprint_ImageView.h
+++ b/blueprint/core/blueprint_ImageView.h
@@ -14,6 +14,20 @@
 
 namespace blueprint
 {
+    namespace
+    {
+        // juce::URL::isWellFormed is currently not a complete
+        // implementation, so we have this slightly more robust check
+        // for now.
+        bool isWellFormedURL(const juce::URL &url)
+        {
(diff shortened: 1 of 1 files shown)


blueprint/core/blueprint_ImageView.h
```diff
@@ -14,6 +14,20 @@
...
+    namespace
+    {
+        // juce::URL::isWellFormed is currently not a complete
+        // implementation, so we have this slightly more robust check
+        // for now.
+        bool isWellFormedURL(const juce::URL &url)
+        {
+            return url.isWellFormed()  &&
+                   url.getScheme().isNotEmpty() &&
+                   !url.toString(false).startsWith("data");
```
nick-thompson, line 26: Heh, the `juce::URL` class is confusing, because a `data:` URL _is_ well formed!

nick-thompson:
Great, thanks!

PR #122, opened 2020-10-25
Update react-reconciler to support hooks

JoshMarler:
@nick-thompson, this bumps react-reconciler so that we can support hooks. All seems to be working my end. Maybe worth giving it a quick sanity check yourself as I've noted the occasional seemingly random crash in the GainPlugin example. I don't think it's related to the upgrade and is something that needs separate investigation.

### DIFF
--- a/examples/GainPlugin/jsui/package-lock.json
+++ b/examples/GainPlugin/jsui/package-lock.json
@@ -10,7 +10,7 @@
       "integrity": "sha512-vG6SvB6oYEhvgisZNFRmRCUkLz11c7rp+tbNTynGqc6mS1d5ATd/sGyV6W0KZZnXRKMTzZDRgQT3Ou9jhpAfUg==",
       "dev": true,
       "requires": {
-        "@babel/highlight": "7.10.4"
+        "@babel/highlight": "^7.10.4"
       }
     },
     "@babel/compat-data": {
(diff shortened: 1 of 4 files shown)


JoshMarler:
@nick-thompson , Follow up to this. I've just been looking at debugging under Windows again for the docs and noticed I'm hitting errors in the object-assign package: https://www.npmjs.com/package/object-assign

I updated to react-reconciler 0.25.1 to see if this made any difference as react-reconciler appears to be the source of the object-assign issues but I still see the same problem. Interestingly the problem is a little worse in 0.25.1 as the example crashes and prints the red screen of doom on startup, however if I then build/evaluate the bundle a second time (after first load) everything works as expected.

JoshMarler:
Quick update, built everything from scratch on Windows and fetched all dependencies from scratch. 

Error in dev mode shows `identifier 'map' undefined`. I have a suspicion this is similar to the behaviour I saw in my own app which required an alteration to the babel preset-env setup etc. I'll investigate a little further.

nick-thompson:
Iiiiinteresting... ok, yea, let me know what else you find

JoshMarler:
@nick-thompson, I suspect this could be the issue: https://github.com/svaarala/duktape/issues/1918

Though why re-evaluating the bundle causes things to work I'm not sure.

JoshMarler:
[babel.txt](https://github.com/nick-thompson/blueprint/files/5438923/babel.txt)


@nick-thompson , I think this is all down to the core-js imports in `blueprint/packages/juce-blueprint/src/index.tsx` only applying within the scope of the `juce-blueprint` modules. 

Because react is now and external/peer dependency it exists in the end user's module/bundle. I have the setup in the attached dev in my app locally at the moment. It basically uses `@babel/preset-env` with `usBuiltIns` which as I understand it makes the core-js polyfills available to the global scope, this resolves issues with 3rd party dependencies using features not currently supported by duktape. 

I'm not totally sure where the best place to use @babel/preset-env and the above settings is. I assume it should be in the app config files rather than the `juce-blueprint` package but I suspect you'll be able to advise on what the most "correct" setup here is.

JoshMarler:
Adding this as a note for myself as much as anything as I'll resolve this tomorrow. 

The following page provides a good explanation of the differences between babel preset-env and babel tansform-runtime:

https://github.com/zloirock/core-js#babel

Basically `preset-env` with `useBuiltIns: usage` will put core-js polyfills into the global scope but only add polyfills used by your bundle's sources automatically. Other polyfills like `Map` and `Set` required for 3rd party libs need to be added manually I believe. 

So, preset-env may be the way to go but it's possibly the move of react into `peerDependencies` actually means that the polyfills for `Map` and `Set` currently in `juce-blueprint` should be moved into user applications (we can update the project templates accordingly). 

I will sanity check this all tomorrow. I suspect adding `Map` and `Set` to the `GainPlugin` example's `index.js` file will probably resolve the issues I've seen at startup.

nick-thompson:
Awesome, thanks

nick-thompson:
@JoshMarler where do we stand on this one? Sanity checks all good? I'm happy to merge it if you give the thumbs up

JoshMarler:
@nick-thompson , this one needs a quick experiment to determine  whether `preset-env` with `useBuiltIns` is the right way to go. Gut feeling is this might be a good choice if it avoids users having to add imports to polyfills in their `index.js`/`index.ts` for each app but I'll confirm the definite cause of the polyfill error tomorrow and get this updated.

JoshMarler:
Updating for reference.

This is blocked by our issue with the polyfill.js implementation for setTimeout and friends. The react-reconciler update appears to be clashing with the polyfill. Plan is to implement a native setTimeoit binding with juce::MultiTimer and apply this to the global object in EcmascriotEngine before bundle evaluation. Once this is done the reconciler/hooks update can go in.

nick-thompson:
@JoshMarler now that #188 is in, I think we can move ahead with this. Mind giving it a rebase and taking it for a spin? Let's get this in for beta.

JoshMarler:
@nick-thompson, yep sure thing. Will give this a test today.

JoshMarler:
@nick-thompson, Have tested with `react-reconciler` 0.25.1 and all looking good. Perhaps let CI do a check and then this should be good to merge.

nick-thompson:
🙌 awesome, thanks!

I'll take it for a spin myself this evening as well, will merge after

PR #123, opened 2020-10-26
Add missing babel plugin dependency (recommended for TS)

JoshMarler:
@nick-thompson , Noticed on a fresh install/build from scratch this babel dependency is missing. Referenced in the `babel.config.js` for Typescript.


nick-thompson:
:+1: nice catch, thanks

PR #126, opened 2020-10-29
Support CSS function-style properties

olidacombe:
Move from non-css property names like `transform-rotate` to their css counterparts with functional values.

Allows definition of a set of valid "macro calls" for a css property (e.g. for `transform` we can define `rotate(...)`, `skew(...)`, ...) to be translated to individual blueprint property: string/float pairs (e.g. `transform: rotate(90deg) => transform-rotate: pi/4`).

Only `transform: rotate() => transform-rotate: ...` has been implemented here as an example.

The optional second argument to `getMacroCalls` specifies an array of valid functional property values (otherwise it'll accept any alphanumeric).  To develop valid functional values for `transform` further, you'd add to the array `["rotate"]` and handle in `Transform.default`.

### DIFF
--- a/packages/react-juce/src/lib/MacroProperties/Transform.ts
+++ b/packages/react-juce/src/lib/MacroProperties/Transform.ts
@@ -1,0 +1,158 @@
+import { TPropertyAssignment } from "./types";
+import { getMacroCalls } from "./util";
+import matrix from 'matrix-js';
+
+const rotateMultipliers = {
+  deg: Math.PI / 180.0,
+  grad: Math.PI / 200.0,
+  rad: 1,
+  turn: 2 * Math.PI,
+};
+
+const angleUnitMatcher = new RegExp(
+  `(-?[0-9]*\.?[0-9]+)(${Object.keys(rotateMultipliers).join("|")}|)$`
+);
+
+const toRadians = (arg: string) => {
+  // @ts-ignore
+  let [, angle, unit] = angleUnitMatcher.exec(arg);
+  let angleFloat = parseFloat(angle);
+  if (angleFloat === NaN) return NaN;
+  if (unit === "" && angleFloat !== 0) return NaN;
+  angleFloat *= rotateMultipliers[unit] || 1;
+  return angleFloat;
+}
+
+const stringArgsToFloat = (...args: string[]) => args.map(parseFloat);
+const stringArgsToRadians = (...args: string[]) => args.map(toRadians);
+
+const identity = matrix([
+  [1, 0, 0, 0],
+  [0, 1, 0, 0],
+  [0, 0, 1, 0],
+  [0, 0, 0, 1],
+]);
+
+const rotate3d = (x: number, y: number, z: number, theta: number) => {
+  const sinTheta = Math.sin(theta);
+  const cosTheta = Math.cos(theta);
+  const r = Math.sqrt(x ** 2 + y ** 2 + z ** 2)
+  if (r === 0) { return identity; }
+  const [u, v, w] = [x / r, y / r, z / r];
+
+  return matrix([
+    [cosTheta + (1 - cosTheta) * u ** 2, u * v * (1 - cosTheta) - w * sinTheta, u * w * (1 - cosTheta) + v * sinTheta, 0],
+    [v * u * (1 - cosTheta) + w * sinTheta, cosTheta + (1 - cosTheta) * v ** 2, v * w * (1 - cosTheta) - u * sinTheta, 0],
+    [w * u * (1 - cosTheta) - v * sinTheta, w * v * (1 - cosTheta) + u * sinTheta, cosTheta + (1 - cosTheta) * w ** 2, 0],
+    [0, 0, 0, 1]
+  ]);
+}
+
+const rotateX = (theta: number) => rotate3d(1, 0, 0, theta);
+const rotateY = (theta: number) => rotate3d(0, 1, 0, theta);
+const rotateZ = (theta: number) => rotate3d(0, 0, 1, theta);
+
+const scale3d = (sx = 1, sy = 1, sz = 1) => matrix([
+  [sx, 0, 0, 0],
+  [0, sy, 0, 0],
+  [0, 0, sz, 0],
+  [0, 0, 0, 1]
+]);
+
+const scale = (sx = 1, sy = 1) => scale3d(sx, sy);
+const scaleX = (sx = 1) => scale(sx);
+const scaleY = (sy = 1) => scale(1, sy);
+const scaleZ = (sz = 1) => scale3d(1, 1, sz);
+
+const skew = (alpha = 0, beta = 0) => {
+  const tanAlpha = Math.tan(alpha) || 0;
+  const tanBeta = Math.tan(beta) || 0;
+  return matrix([
+    [1 + tanAlpha * tanBeta, tanAlpha, 0, 0],
+    [tanBeta, 1, 0, 0],
+    [0, 0, 1, 0],
+    [0, 0, 0, 1],
+  ]);
+}
+
+const skewX = (theta: number) => skew(theta);
+const skewY = (theta: number) => skew(0, theta);
+
+const translate3d = (dx = 0, dy = 0, dz = 0) => matrix([
+  [1, 0, 0, dx],
+  [0, 1, 0, dy],
+  [0, 0, 1, dz],
+  [0, 0, 0, 1]
+]);
+
+const translate = (dx = 0, dy = 0) => translate3d(dx, dy, 0);
+const translateX = (dx = 0) => translate3d(dx);
+const translateY = (dy = 0) => translate3d(0, dy);
+const translateZ = (dz = 0) => translate3d(0, 0, dz);
+
+const argsTransformMap = {
+  matrix: stringArgsToFloat,
+  matrix3d: stringArgsToFloat,
+  rotate: stringArgsToRadians,
+  rotate3d: (x, y, z, t) => [...stringArgsToFloat(x, y, z), toRadians(t)],
+  rotateX: stringArgsToRadians,
+  rotateY: stringArgsToRadians,
+  rotateZ: stringArgsToRadians,
+  scale: stringArgsToFloat,
+  scale3d: stringArgsToFloat,
+  scaleX: stringArgsToFloat,
+  scaleY: stringArgsToFloat,
+  scaleZ: stringArgsToFloat,
+  skew: stringArgsToRadians,
+  skewX: stringArgsToRadians,
+  skewY: stringArgsToRadians,
+  translate: stringArgsToFloat,
+  translate3d: stringArgsToFloat,
+  translateX: stringArgsToFloat,
+  translateY: stringArgsToFloat,
+  translateZ: stringArgsToFloat,
+}
+
+const transformMatrixMap = {
+  none: identity,
+  matrix: (l11, l12, dx, l21, l22, dy) => matrix([
+    [l11, l12, 0, dx],
+    [l21, l22, 0, dy],
+    [0, 0, 1, 0],
+    [0, 0, 0, 1]
+  ]),
+  matrix3d: (m11, m12, m13, m14, m21, m22, m23, m24, m31, m32, m33, m34, m41, m42, m43, m44) => matrix([
+    [m11, m12, m13, m14],
+    [m21, m22, m23, m24],
+    [m31, m32, m33, m34],
+    [m41, m42, m43, m44],
+  ]),
+  // perspective: ()=>identity, // TODO
+  rotate: rotateZ,
+  rotateX, rotateY, rotateZ,
+  scale, scale3d, scaleX, scaleY, scaleZ,
+  skew, skewX, skewY,
+  translate, translate3d, translateX, translateY, translateZ
+};
+
+const matrixToArray = m => m().reduce((acc, v) => [...acc, ...v]);
+
+export default function (value: string): TPropertyAssignment[] {
+  const calls = getMacroCalls(value);
+  // operations performed right to left
+  // i.e. transform: translateY(20) rotate(90deg)
+  // means rotate 90 degrees, then translate up 20
+  // so we reverse in-place the calls array which initially
+  // has the translation first followed by the rotate.
+  calls.reverse()
+  //@ts-ignore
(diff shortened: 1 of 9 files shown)


packages/juce-blueprint/src/lib/MacroProperties/Transform.ts
```diff
@@ -0,0 +1,30 @@
...
+  rad: 1,
+  turn: 2 * Math.PI,
+};
+
+const angleUnitMatcher = new RegExp(
+  `(-?[0-9]*\.?[0-9]+)(${Object.keys(rotateMultipliers).join("|")}|)$`
+);
+
+export default function (value: string): TPropertyAssignment[] {
+  const calls = getMacroCalls(value, ["rotate"]);
```
nick-thompson, line 16: I feel a bit like this won't scale very neatly. I can envision this list growing to include all possible 2D transform macros, and the `if` block on line 19 being duplicated many times for each one. I also think this is a particularly interesting first view property macro, because it ends up being quite complicated.

At the end of the day, each one of these functions boils down to an Affine transform matrix. I'm wondering if our macro helpers here should ultimately return a single affine transform matrix:  `transform-matrix: [1, 0, 0, 0, 0, 1, 0, 0, 0, 0, 1, 0, 0, 0, 0, 1]` like the `matrix3d` transform function (https://developer.mozilla.org/en-US/docs/Web/CSS/transform).

Then each of these helper functions (rotate, transform, skew, etc) just operate over a matrix that eventually gets passed to native:
```
const transformFunctions = value.match(/([a-z]+\([a-z0-9,.\s]+\))/ig);

// Here transform functions would be a list like [ "rotate(90deg, 25deg)", "translateX(25)" ]
// We reduce over that list, starting with an identity matrix, and ultimately yielding some aggregate matrix
let results = transformFunctions.reduce(function(mat, funcCall) {
   let [funcName, ...args] = splitFuncCall(funcCall);

   // Here funcName would be e.g. "rotate" and "args" would be ["90deg", "25deg"]
   // We call into our `transformFunctionMap` for the `rotate` helper to act over our matrix
   return transformFunctionMap[funcName](mat, ...args);
}, makeIdentityTransformMatrix());

return ['transform-matrix', results];
```

Then we have a map for handling discrete 2d transform function macros.

```
transformFunctionMap.rotate = function(mat, args) {
   // Here we're looking to apply our ["90deg", "25deg"] args over the given `mat` matrix
}
```

The reason I think this is important to do in javascript is because if instead our goal is to ultimately boil down all of these transform helpers to a set of native properties that get set on the View class and handled their, then how do we handle this case?

```css
.hi {
  width:100px;
  height:100px;
  background-color: red;
  transform: rotate(45deg) scale(0.5) rotate(45deg);
}
```
This is a valid transform, and it's equivalent to `rotate(90deg) scale(0.5)`. But if we were to map this to [['transform-rotate', 45], ['scale', 0.5], ['transform-rotate', 45]] you can see that we would clobber the rotate property instead of stacking it.

nick-thompson (commented):
Hey @olidacombe thanks so much for diving into this! Sorry for the delayed review, I peeked at this a few days ago and realized it's actually quite complicated, and have been thinking about it for a while. I left some comments inline, let me know what you think

packages/juce-blueprint/src/lib/MacroProperties/Transform.ts
olidacombe: Ah totally yes, this is really the right way to do it in my opinion.  I'll get to that.  And your above comments look good too.

I'm guessing we build should build the full 4d matrix in javascript as suggested, accounting for what's expressible in css, even though I'm not sure we'd use the z dimension in View geometry (just throw away the second-last row/column before handing over to `juce::AffineTransform`).

nick-thompson: Great!

> I'm guessing we build should build the full 4d matrix in javascript as suggested, accounting for what's expressible in css, even though I'm not sure we'd use the z dimension in View geometry (just throw away the second-last row/column before handing over to juce::AffineTransform).

Yup, totally agreed

olidacombe: Hey @nick-thompson , I made another pass through this taking the above into account and it feels way better. Let me know what you think when you have a chance to inspect. Sorry about the editor whitespace changes etc, can pick through them if it's a problem.

packages/react-juce/src/lib/MacroProperties/Transform.ts
```diff
@@ -0,0 +1,130 @@
...
+};
+
+const matrixToArray = m => m().reduce((acc, v) => [...acc, ...v]);
+
+export default function (value: string): TPropertyAssignment[] {
+  const calls = getMacroCalls(value, Object.keys(transformFunctionMap));
+  // operations performed right to left
+  calls.reverse()
+  //@ts-ignore
+  const transformMatrix = calls.reduce((acc, { macro: f, args }) => {
```
nick-thompson, line 125: 🙌 awesome

packages/react-juce/src/lib/MacroProperties/Transform.ts
```diff
@@ -0,0 +1,130 @@
...
+  [0, 0, sz, 0],
+  [0, 0, 0, 1]
+]);
+
+const scale = argsToFloat((sx = 1, sy = 1) => scale3d(sx, sy));
+const scaleX = argsToFloat((sx = 1) => scale(sx));
+const scaleY = argsToFloat((sy = 1) => scale(1, sy));
+const scaleZ = argsToFloat((sz = 1) => scale3d(1, 1, sz));
+
+const skew = (alpha = '0', beta = '0') => {
```
nick-thompson, line 68: Not sure I understand the string type arguments here?

packages/react-juce/src/lib/MacroProperties/Transform.ts
```diff
@@ -0,0 +1,130 @@
...
+    [tanBeta, 1, 0, 0],
+    [0, 0, 1, 0],
+    [0, 0, 0, 1],
+  ]);
+}
+
+const skewX = (theta) => skew(theta);
+const skewY = (theta) => skew('0', theta);
+
+const translate3d = argsToFloat((dx = 0, dy = 0, dz = 0) => matrix([
```
nick-thompson, line 84: Not sure I understand the `argsToFloat` part here either.

I think it'd be easier to understand if all of our rotate/scale/translate/etc helpers operated over strictly numeric values of a consistent unit (i.e. radians). Any conversions that we need to do to go from the user's string argument `rotate(90deg)` to that numeric value can be done above this layer.

The other reason I think it's helpful to think that way is that `const argsToFloat = f => (...args) => f(...args.map(parseFloat));` is quite an expensive wrapper :). Most major ES6->ES5 transpilers will rewrite that touching the `arguments` object, which is a known performance in Duktape and often a deopt flag in V8/SpiderMonkey. So, liberally wrapping our helper functions in this could end up being quite costly for apps that use these `transform` properties regularly

packages/react-juce/src/lib/MacroProperties/Transform.ts
```diff
@@ -0,0 +1,130 @@
...
+  [0, 0, 1, dz],
+  [0, 0, 0, 1]
+]));
+
+const translate = argsToFloat((dx = 0, dy = 0) => translate3d(dx, dy, 0));
+const translateX = argsToFloat((dx = 0) => translate3d(dx));
+const translateY = argsToFloat((dy = 0) => translate3d(0, dy));
+const translateZ = argsToFloat((dz = 0) => translate3d(0, 0, dz));
+
+const transformFunctionMap = {
```
nick-thompson, line 96: 🙌 yeaaa this is great!

packages/react-juce/src/lib/MacroProperties/Transform.ts
```diff
@@ -0,0 +1,130 @@
...
+  skew, skewX, skewY,
+  translate, translate3d, translateX, translateY, translateZ
+};
+
+const matrixToArray = m => m().reduce((acc, v) => [...acc, ...v]);
+
+export default function (value: string): TPropertyAssignment[] {
+  const calls = getMacroCalls(value, Object.keys(transformFunctionMap));
+  // operations performed right to left
+  calls.reverse()
```
nick-thompson, line 123: Might be helpful to elaborate on the comment here with an example? I can imagine myself coming back to this 6 months from now and going "what? Why?"

packages/react-juce/src/lib/MacroProperties/util.ts
```diff
@@ -0,0 +1,35 @@
...
+  const argArray = a.trim().split(/\s*,\s*/);
+  if (argArray.length === 1 && argArray[0] === '') return [];
+  return argArray;
+}
+
+export const getMacroCalls = (
+  s: string,
+  candidates?: string[]
+): TMacroCall[] => {
+  const macroMatcher = String.raw`(${candidates ? `${candidates.join("|")}` : "[a-zA-Z0-9]+"
```
nick-thompson, line 14: Feels a bit like we're writing a code path that we don't even want to use here.

If we pass in the candidates array here by calling `Object.keys(transformFunctionMap)` then the `candidates` argument doesn't need to be optional, or treated as optional (hence no need for the ternary operator here in the regex construction). I think we could invoke an `invariant` here that the candidates array is in fact present.

But, one better, I think we could drop the candidates array altogether and just use the `[a-zA-Z0-9]+` matcher that you have here already. The result will be that we'll return a `macroCalls` array below that _might_ contain function names that we don't actually support. But that's easy enough: in the `reduce` call at the end of `Transform.ts` we can just do

```
const transformMatrix = calls.reduce((acc, { macro: f, args }) => {
  if (!transformFunctionMap.hasOwnProperty(f)) {
    return acc;
  }
  const transform = transformFunctionMap[f](...args);
  return matrix(transform.prod(acc));
}, identity);
```
That feels a little simpler to me. What do you think?

nick-thompson (commented):
Excellent @olidacombe, thank you!

I love how easily you've extended this to handle so many of the transform functions.

I have a couple further small comments below, but this feels very close to merge for me. Let me know what you think!

packages/react-juce/src/lib/MacroProperties/util.ts
olidacombe: Ah yeah, I knew there was a reason I felt a little exposed writing the call to `transformFunctionMap[f]` unchecked, in the case where we haven't passed candidates from the right `Object.keys`, we might not get a function.  I think dropping the candidates arg altogether and checking as you've written is the way to go.

packages/react-juce/src/lib/MacroProperties/Transform.ts
olidacombe: Ha, good point.  Willing to take advice on a clear comment, hopefully this amendment will do.

packages/react-juce/src/lib/MacroProperties/Transform.ts
olidacombe: I see what you mean about the mathematical functions being relieved of concern about strings.  The part about the expensive wrapper is a bit above my paygrade, so I'd appreciate your thoughts on what I've done.  I moved the string argument to float translation functions out to a map, and compose those with purely numerical args => matrix functions.  Is this actually an improvement at all in term of the transpiled code?

packages/react-juce/src/lib/MacroProperties/Transform.ts
nick-thompson: Thanks, this looks good!

packages/react-juce/src/lib/MacroProperties/Transform.ts
```diff
@@ -0,0 +1,158 @@
...
+  // means rotate 90 degrees, then translate up 20
+  // so we reverse in-place the calls array which initially
+  // has the translation first followed by the rotate.
+  calls.reverse()
+  //@ts-ignore
+  const transformMatrix = calls.reduce((acc, { macro: f, args }) => {
+    if (!transformMatrixMap.hasOwnProperty(f)) {
+      return acc;
+    }
+    const argsTransform = argsTransformMap[f] || (() => args);
```
nick-thompson, line 153: Should we bail and return `acc` if the argsTransformMap doesn't have `f`?

packages/react-juce/src/lib/MacroProperties/Transform.ts
nick-thompson: Yea, it's tricky :)

If we're running in an engine like V8, this is definitely an improvement in the transpiled code, because we're touching the `arguments` object in fewer places (which means fewer de-optimization flags).

In Duktape, we still will pay the penalty for touching that `arguments` object in those 2 places, since each transform function has to first run through it string->float/radian conversion first, but again, less touching `arguments` = better.

So to me this looks good! I think this is about as good as we're going to get without another very long, very detailed performance pass. And if we do that performance pass, we shouldn't block the feature from merging before we do.

packages/react-juce/src/lib/MacroProperties/Transform.ts
olidacombe: I say either we allow the above or abort the whole calculation and return `[]`.  Reason being the above skips nonsensical items in `transform:` to return a partial transform (all the macros that make sense, reduced), and bailing part way through the reducer also would return a partial transform.  I reckon either allow partial transforms in the form above where we skip nonsense rather than truncate at first nonsense, or don't allow partially permissible transforms at all and force the whole thing to identity.  I hope that makes some sense.  Basically I think skip is better than truncate if partial is ok, but also fine with partial not being ok.

nick-thompson:
Ok fantastic! Thanks so much @olidacombe. I left one more comment for you inline, but I'm happy to merge now. Would you mind rebase against master + squash before I do? Or I'm happy to do it myself as I merge.

nick-thompson (approved):

olidacombe:
@nick-thompson great, I'm not around right now, but can rebase later.  If you're in the business or merging right now then totally happy for you to take care of the squash.

olidacombe:
Oh 1 more thing, @nick-thompson are you ok with this killing the existing transform-rotate prop? It will break client UIs expecting that. It's simple enough to add this back in.

nick-thompson:
> Oh 1 more thing, @nick-thompson are you ok with this killing the existing transform-rotate prop?

Yes! We'll have to make these breaking changes at some point anyway if the goal is to standardize around more CSS-like standards. Now's the time for those breaks :). Merged! Thank you again

PR #127, opened 2020-10-30
Some simple additions to ScrollView

JoshMarler:
Supports things like conditional display of x/y scrollbars, setting
  scrollbar track/thumb colors, setting scrollbar width and specifying
  if scroll-on-drag is enabled.

### DIFF
--- a/blueprint/core/blueprint_ScrollView.cpp
+++ b/blueprint/core/blueprint_ScrollView.cpp
@@ -1,0 +1,100 @@
+/*
+  ==============================================================================
+
+    blueprint_ScrollView.cpp
+    Created: 10 Nov 2020 18:07pm
+
+  ==============================================================================
+*/
+
+#include "blueprint_ScrollView.h"
+
+namespace blueprint
+{
+    ScrollView::ScrollView()
+    {
+        addAndMakeVisible(viewport);
+        viewport.setScrollBarsShown(true, true);
+    }
+
+    void ScrollView::setProperty (const juce::Identifier& name, const juce::var& value)
+    {
+        View::setProperty(name, value);
+
+        auto showX = viewport.isHorizontalScrollBarShown();
+        auto showY = viewport.isVerticalScrollBarShown();
+
+        auto overflowCheck = [](const juce::String& p, const juce::String &v)
+        {
+            if (juce::StringRef("scroll") == v) { return true; }
+            if (juce::StringRef("hidden") == v) { return false; }
+
+            const juce::String e = "Invalid prop value. Prop '" + p + "' must be a string of 'hidden' or 'scroll'";
+            throw std::invalid_argument(e.toStdString());
+        };
+
+        if (name == overflowYProp)
+        {
+            showY = overflowCheck(overflowYProp.toString(), value);
+        }
+
+        if (name == overflowXProp)
+        {
+            showX = overflowCheck(overflowXProp.toString(), value);
+        }
+
+        viewport.setScrollBarsShown(showY, showX);
+
+        if (name == scrollBarWidthProp)
+        {
+            if (!props[scrollBarWidthProp].isDouble())
+                throw std::invalid_argument("Invalid prop value. Prop \'scollbar-width\' must be a number.");
+
+            const int thickness = props[scrollBarWidthProp];
+            viewport.setScrollBarThickness(thickness);
+        }
+
+        if (name == scrollbarThumbColorProp)
+        {
+            if (!props[scrollbarThumbColorProp].isString())
+                throw std::invalid_argument("Invalid prop value. Prop \'scrollbar-thumb-color\' must be a color string.");
+
+            const auto thumbColor = juce::Colour::fromString(props[scrollbarThumbColorProp].toString());
+            viewport.getVerticalScrollBar().setColour(juce::ScrollBar::thumbColourId, thumbColor);
+            viewport.getHorizontalScrollBar().setColour(juce::ScrollBar::thumbColourId, thumbColor);
+        }
+
+        if (name == scrollbarTrackColorProp)
+        {
+            if (!props[scrollbarTrackColorProp].isString())
+                throw std::invalid_argument("Invalid prop value. Prop \'scrollbar-track-color\' must be a color string.");
+
+            const auto trackColor = juce::Colour::fromString(props[scrollbarTrackColorProp].toString());
+            viewport.getVerticalScrollBar().setColour(juce::ScrollBar::trackColourId, trackColor);
+            viewport.getHorizontalScrollBar().setColour(juce::ScrollBar::trackColourId, trackColor);
+        }
+
+        if (name == scrollOnDragProp)
+        {
+            if (!props[scrollOnDragProp].isBool())
+                throw std::invalid_argument("Invalid prop value. Prop \'scoll-on-drag\' must be a bool.");
+
+            const bool scrollOnDrag = props[scrollOnDragProp];
+            viewport.setScrollOnDragEnabled(scrollOnDrag);
+        }
+    }
+
+    void ScrollView::addChild (View* childView, int index)
+    {
+        juce::ignoreUnused (index);
+        jassert (viewport.getViewedComponent() == nullptr);
+        viewport.setViewedComponent(childView, false);
+    }
+
+    void ScrollView::resized()
+    {
+        const auto &bounds = getLocalBounds();
+        viewport.setBounds(bounds);
+        View::resized();
+    }
+}
--- a/blueprint/core/blueprint_ScrollView.h
+++ b/blueprint/core/blueprint_ScrollView.h
@@ -11,7 +11,6 @@
 
 #include "blueprint_View.h"
 
-
 namespace blueprint
 {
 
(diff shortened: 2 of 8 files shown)


JoshMarler:
@nick-thompson , Couple questions inline under `TODO` comments. Mainly do we hide scrollbars by default. 

Also, What do you think to a `scrollbar-vertical-pos` and `scrollbar-horizontal-pos` these would allow user to specify whether scrollbars are on bottom/top/left/right. I don't need this right now myself but happy to add if you think it's useful.

blueprint/core/blueprint_ScrollView.h
```diff
@@ -23,25 +23,70 @@ namespace blueprint
...
             {
-                juce::Colour c = juce::Colour::fromString(props["scrollbar-thumb-color"].toString());
+                //TODO: Support CSS scrollbar-width "thin", "auto", "none" values
+                //      or stick with a pixel value for flexibility? Or both?
+                viewport.setScrollBarThickness(props[scrollBarWidthProp]);
+            }
+
+            if (name == scrollbarColorProp)
+            {
+                auto colors = juce::StringArray::fromTokens(props[scrollbarColorProp].toString(), " ", {});
```
JoshMarler, line 69: This allows a space separated pair of colors for thumb and track as per spec here: https://developer.mozilla.org/en-US/docs/Web/CSS/scrollbar-color

blueprint/core/blueprint_ScrollView.h
```diff
@@ -23,25 +23,70 @@ namespace blueprint
...
+                const auto showVertical   = props[overflowYProp] == "scroll" || props[overflowProp] == "scroll";
+                const auto showHorizontal = props[overflowXProp] == "scroll" || props[overflowProp] == "scroll";
+
+                viewport.setScrollBarsShown(showVertical, showHorizontal);
+            }
+
+            if (name == scrollBarWidthProp)
             {
-                juce::Colour c = juce::Colour::fromString(props["scrollbar-thumb-color"].toString());
+                //TODO: Support CSS scrollbar-width "thin", "auto", "none" values
```
JoshMarler, line 62: Should we support these CSS values or do we keep this as just an integer value for scrollbar width ?

blueprint/core/blueprint_ScrollView.h
```diff
@@ -23,25 +23,70 @@ namespace blueprint
     class ScrollView : public View
     {
     public:
+        //==============================================================================
+        // Props following CSS Scrollbars spec with some omissions/additions.
+        // https://developer.mozilla.org/en-US/docs/Web/CSS/CSS_Scrollbars
+        static inline juce::Identifier overflowProp       = "overflow";
```
nick-thompson, line 29: So I think our general stance on native props should be to make a minimal set of very concrete props, and defer all composition and clever syntax up to the javascript layer.

For example, native never needs to know about the idea of this `overflow` prop. At the `ScrollView` layer in the js library, we can catch an `overflow` prop and transform it to the necessary composition of `overflow-x` and `overflow-y`.

blueprint/core/blueprint_ScrollView.h
```diff
@@ -23,25 +23,70 @@ namespace blueprint
...
         ScrollView()
         {
             addAndMakeVisible(viewport);
-            // TODO: Set these based on props...
-            viewport.setScrollBarsShown(true, false);
+
+            //TODO: Make scrollbars hidden by default so that
+            //      explicit overflow settings required? Or show on
+            //      default and overflow: hidden can be applied?
+            viewport.setScrollBarsShown(false, false);
```
nick-thompson, line 44: I think visible by default, and let `overflow: hidden` if user desires. I believe that matches web behavior

blueprint/core/blueprint_ScrollView.h
nick-thompson: Per my comment above, let's have the native API deal only in fixed integer values, but in the javascript layer we can allow `thin` and `auto` and map those to integer values at that layer

blueprint/core/blueprint_ScrollView.h
nick-thompson: Similar here too... let's have our native props be `scrollbar-thumb-color` and `scrollbar-track-color` and defer the composition by way of space separated pairs up to the js layer

nick-thompson:
Looks great! A few comments inline, but mostly 1 driving idea

nick-thompson:
Oh @JoshMarler let's also, while we're here, add some `throw std::invalid_argument` lines for unexpected property values so we can start building out robust error handling there. Don't need to go crazy with it, but let's chip away at this whenever we do work on native view props

blueprint/core/blueprint_ScrollView.h
JoshMarler: Yep so lose `overflowProp` here and maintain the minimal `overflowXProp` and `overflowYProp`. JS converts `overflow` to desired composition for C++/native?

blueprint/core/blueprint_ScrollView.h
JoshMarler: Cool. Any thoughts on sensible values?

blueprint/core/blueprint_ScrollView.h
JoshMarler: Just to make sure we're both using the same language here ... :-) 

By "native props" you mean the props C++ side? i.e. C++ holds a `scrollbarThumbColorProp` and a `scrollbarTrackColorProp`, JS `ScrollView` intercepts `scrollbar-color` changes, splits the string and passes as `scrollbar-thumb-color` and `scrollbar-track-color` to C++ ? 

So we maintain this prop name: https://developer.mozilla.org/en-US/docs/Web/CSS/scrollbar-color ?

JoshMarler:
> Oh @JoshMarler let's also, while we're here, add some `throw std::invalid_argument` lines for unexpected property values so we can start building out robust error handling there. Don't need to go crazy with it, but let's chip away at this whenever we do work on native view props

Yeah perfect. Will deal with this all tomorrow. `invalid_argument` for bad prop types sounds like a sensible things to start adopting. Probably makes sense across the board in things like the `CanvasView` function bindings, i.e. if nonsense arguments are passed to `lineTo` etc.

blueprint/core/blueprint_ScrollView.h
nick-thompson: Yup exactly

blueprint/core/blueprint_ScrollView.h
nick-thompson: Ermmm what's the default scrollbar width, like 5px? Maybe 5px and 3px or something? I think it's probably fairly innocuous

JoshMarler:
@nick-thompson , I've pushed some additions to parse/intercept props in `ScrollView.ts`. Let me know what you think. I think the only remaining thing is to add some more error checking to `ScrollView.cpp`. I think I'll follow up a separate PR with a generalised `isPropValid` system as I want to think about that a little more. For now shall we just have a check on property type ? i.e `juce::var::isBool`, and value is within range for string/numeric props? Some of this is already done in JS at `ScrollView.ts`.

packages/react-juce/src/components/ScrollView.ts
```diff
@@ -1,20 +1,94 @@
...
+    switch(scrollBarWidthProp) {
+      // JUCE's look and feel V2 uses 18px as default scrollbar thickness.
+      case 'auto':
+        props['scrollbar-width'] = 18;
+        return;
+      case 'thin':
+        props['scrollbar-width'] = 9;
+        return;
+      case 'none':
+        props['overflow'] = 'hidden';
```
JoshMarler, line 37: This is roughly the spec of 'none'. Is enabling `scrollOnDrag` the right behaviour here?

blueprint/core/blueprint_ScrollView.cpp
```diff
@@ -0,0 +1,102 @@
...
+    {
+        addAndMakeVisible(viewport);
+        viewport.setScrollBarsShown(true, true);
+    }
+
+    void ScrollView::setProperty (const juce::Identifier& name, const juce::var& value)
+    {
+        View::setProperty(name, value);
+
+        if (name == overflowYProp)
```
nick-thompson, line 24: Hmmm thinking we can maybe shorten this up a bit too:

```
auto showX = viewport.isHorizontalScrollBarShown();
auto showY = viewport.isVerticalScrollBarShown();

if (name == overflowYProp)
{
    showY = std::invoke([](const juce::String& p) {
        if ("scroll" == p) { return true; }
        if ("hidden" == p) { return false; }
        throw std::invalid_argument("Invalid prop value. Prop \'overflow-y\' must be a string of 'hidden' or 'scroll'");
    }, props[name]);
}

if (name == overflowXProp)
{
    same...
}

// This ends up being a noop if we didn't enter the above if blocks
viewport.setScrollBarsShown(showX, showY);
```

That was the closest I could come up with for a pattern-matching like syntax, which I think would be great here. What do you think?

packages/react-juce/src/components/ScrollView.ts
nick-thompson: Yea, I like that

packages/react-juce/src/components/ScrollView.ts
```diff
@@ -1,20 +1,94 @@
...
-  return React.createElement('ScrollView', props, child);
+  let {
+    overflow,
+    "scrollbar-color": scrollbarColor,
+    "scrollbar-width": scrollBarWidth,
+    ...other
+  } = props;
+
+  if (typeof scrollbarColor !== 'undefined')
+    parseScrollbarColorProp(scrollbarColor, other);
```
nick-thompson, line 82: I generally try to avoid passing objects into functions and expecting a mutation to that object. I find that pattern creates headaches more often than not.

I'd prefer these functions return their own objects, with their own props subset, and here we could `other = Object.assign(other, parseScrollbarColorProp(scrollbarColor);`

It feels nit picky in this case, but I think this pattern of pure functions is worth being nit picky about :)

nick-thompson (commented):
This is great, thanks @JoshMarler. I'm pretty much ready to merge, but check inline I have a couple stylistic comments

packages/react-juce/src/components/ScrollView.ts
JoshMarler: Yeah totally. I was being lazy there, I remember thinking "ergh" when I wrote that .... Immutable. Object.assign tip make sense. I'm still not thinking in full JS!

blueprint/core/blueprint_ScrollView.cpp
JoshMarler: Yeah nice. Maybe pull that lambda out to avoid duplication. Ok cool will take that approach.

JoshMarler: @nick-thompson , the only issue with that approach is if you JUST set overflow-x you'll still see the vertical scrollbar until you explicitly set overflow-y to "hidden". Do we think that's an issue?

JoshMarler: de-duped version looks like: 

```
        auto showX = viewport.isHorizontalScrollBarShown();
        auto showY = viewport.isVerticalScrollBarShown();

        auto overflowCheck = [](const juce::String& p, const juce::String &v)
        {
            if (juce::StringRef("scroll") == v) { return true; }
            if (juce::StringRef("hidden") == v) { return false; }

            juce::String e = "Invalid prop value. Prop '" + p + "' must be a string of 'hidden' or 'scroll'";
            throw std::invalid_argument(e.toStdString());
        };

        if (name == overflowYProp)
        {
            showY = overflowCheck(overflowYProp.toString(), value);
        }

        if (name == overflowXProp)
        {
            showX = overflowCheck(overflowXProp.toString(), value);
        }
```

JoshMarler:
@nick-thompson , Pushed and update addressing your comments. One comment inline about  explicitly turning off overflow-x/overflow-y.

blueprint/core/blueprint_ScrollView.h
nick-thompson: :+1: Obviously coming back to this comment way late, but I can see in the code change that we were on the same page!

blueprint/core/blueprint_ScrollView.cpp
nick-thompson: > @nick-thompson , the only issue with that approach is if you JUST set overflow-x you'll still see the vertical scrollbar until you explicitly set overflow-y to "hidden". Do we think that's an issue?

Nope, no issue by me. That makes sense to me as default behavior for the `<ScrollView>`

nick-thompson (approved):

PR #128, opened 2020-11-05
Work In Progress: Adds onclick event bubbling behaviour to match DOM

JoshMarler:
element behaviour.

### DIFF
--- a/packages/juce-blueprint/src/lib/BlueprintBackend.ts
+++ b/packages/juce-blueprint/src/lib/BlueprintBackend.ts
@@ -1,11 +1,16 @@
-import SyntheticEvents, 
-       { SyntheticEvent, 
-         SyntheticMouseEvent, 
-         SyntheticKeyboardEvent } from './SyntheticEvents' 
+import SyntheticEvents, {
+    SyntheticEvent,
+    SyntheticMouseEvent,
+    SyntheticKeyboardEvent
+} from './SyntheticEvents'
 
 /* global __BlueprintNative__:false */
 
-let __rootViewInstance: ViewInstance | null = null;
+type Instance = ViewInstance | RawTextViewInstance;
+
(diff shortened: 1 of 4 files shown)


JoshMarler:
@nick-thompson , this is a work in progress for the `onCick` event bubbling changes we have discussed to support the new `Button` component etc. 

Behaviour appears to be correct. I can see a very slight delay in UI response on mouseUp events which I suspect we can improve upon but perhaps we should just get things in and stable first as it's very minor. The LCA algorithm I've used is an adaptation of the naive implementation found here: 

https://stackoverflow.com/questions/10011857/how-to-find-lowest-common-ancestor-of-a-nary-tree

We should be able to improve upon this if required. 

I'm going to do some further testing now with things like button component etc. I've been testing this is a scrollable "data-table" and behaviour appears to be correct. If you get a chance to test this also that would be much appreciated.

packages/juce-blueprint/src/lib/BlueprintBackend.ts
```diff
@@ -157,19 +145,116 @@ export default {
...
+
+    if (root.getViewId() === lhs.getViewId() || root.getViewId() === rhs.getViewId())
+        return root;
+
+    let i = 0;
+    let lastLCA: Instance | null = null;
+
+    // RawTextViewInstance does not hold a child array so
+    // treat it as empty here.
+    if (!(root instanceof RawTextViewInstance)) {
```
JoshMarler, line 185: This could just be a check on whether _children is undefined but I think it's better to be explicit here for the moment.

JoshMarler:
@nick-thompson, I'm going to close this and merge the event work into: https://github.com/nick-thompson/blueprint/pull/118

Will us Button and Slider in the GainPlugin example as a test bed for this behaviour.

PR #132, opened 2020-11-14
Add slider core component

JoshMarler:
@nick-thompson , This is a work in progress. The gesture map and draw basics all work. At the moment the slider doesn't have internal state. I think you mentioned wanting to support both a value prop and some internal state if a user doesn't supply a value prop or something along those lines? 

I also need to clean up the actual draw functions, there's a little quirk with some fill overlap on the linear draw routines and I'm rubbish at geometry and draw code generally so the rotary slider is totally wrong at the moment. If you're a graphics wiz and think you could do a rotary slider canvas routine in like 5 minutes I'd be most grateful for the help! If not I'll soldier on at some point today. 

I've followed the approach I took with the Button PR so that there is a `ParameterSlider` composing ontop of the base `Slider` component.


examples/GainPlugin/jsui/src/App.js
```diff
@@ -1,14 +1,17 @@
 import AnimatedFlexBoxExample from './AnimatedFlexBox';
 import Label from './Label';
 import Meter from './Meter';
-import React, { Component } from 'react';
-import Slider from './Slider';
 import ParameterToggleButton from './ParameterToggleButton'
+import ParameterSlider from './ParameterSlider'
+
+import React, { Component } from 'react';
```
JoshMarler, line 7: Not sure why that moved! will sort.

examples/GainPlugin/jsui/src/ParameterSlider.js
```diff
@@ -0,0 +1,92 @@
...
+            defaultValue: state.defaultValue,
+            value: state.currentValue,
+          });
+        }
+    }
+
+    render() {
+        const {
+          parameterId,
+          sliderStrokeColor,
```
nick-thompson, line 67: Looks to me like we don't need these as props? These values are captured by closure into the `onDraw` handler yea?

examples/GainPlugin/jsui/src/ParameterSlider.js
```diff
@@ -0,0 +1,92 @@
...
+          sliderFillColor,
+          onDraw,
```
nick-thompson, line 69: No need to unpack `onDraw` and `mapDragGestureToValue` if we're just gonna pass them through below

packages/react-juce/src/components/Slider.tsx
```diff
@@ -0,0 +1,284 @@
...
+import { View } from './View';
+import { SyntheticMouseEvent } from '../lib/SyntheticEvents';
+
+// Some simple helpers for slider drag gesture -> value mapping
+const _linearHorizontalGestureMap = (mouseDownX: number,
+                                     mouseDownY: number,
+                                     sensitivity: number,
+                                     valueAtDragStart: number,
+                                     dragEvent: SyntheticMouseEvent): number => {
+    let value = 0.0;
```
nick-thompson, line 13: Just for brevity, looks like we can omit this line, and collapse line 16 and 17 into a single `return Math.max(0, Math.min(1, ...etc))`?

packages/react-juce/src/components/Slider.tsx
```diff
@@ -0,0 +1,284 @@
...
+    value = Math.max(0.0, Math.min(1.0, valueAtDragStart + dx * sensitivity));
+    return value;
+}
+
+const _linearVerticalGestureMap = (mouseDownX: number,
+                                   mouseDownY: number,
+                                   sensitivity: number,
+                                   valueAtDragStart: number,
+                                   dragEvent: SyntheticMouseEvent): number => {
+    let value = 0.0;
```
nick-thompson, line 25: Brevity, same

packages/react-juce/src/components/Slider.tsx
```diff
@@ -0,0 +1,284 @@
...
+
+        ctx.lineWidth = lineWidth;
+        ctx.strokeStyle = strokeColor;
+        ctx.fillStyle = fillColor;
+
+        ctx.beginPath();
+        ctx.moveTo(x + radius, y);
+        ctx.arc(x, y, radius, 0, 2 * Math.PI * arcSize);
+
+        const rotateAngle = ((180 + 180 * (1 - arcSize)) * Math.PI) / 180;
```
nick-thompson, line 143: Not sure I follow the rotate part here. Could we just `ctx.arc` once for the rotary slider track, then `ctx.arc` again on top of it for the rotary slider fill?

packages/react-juce/src/components/Slider.tsx
```diff
@@ -0,0 +1,284 @@
...
+            }
+        }
+
+        if (typeof this.props.onMouseDrag === 'function') {
+            this.props.onMouseDrag(e);
+        }
+    }
+
+    _onDraw = (ctx: CanvasRenderingContext2D) => {
+        if (typeof this.props.onDraw === 'function') {
```
nick-thompson, line 243: No need to make our own `_onDraw` here if the only goal is to call the incoming `props.onDraw`. We can just unpack `props` into the `Canvas` below and if there's an onDraw there, it will get passed through.

That said, I think the slider should probably have some default `onDraw` in which case this would be the case to do it....

```
if (typeof this.props.onDraw === 'function') {
  return this.props.onDraw(....)
}

// We'll fall through to here if we don't have an on draw
_rotaryDrawHelper(ctx, this.state.width, this.state.height, this.props.value);
```

packages/react-juce/src/components/Slider.tsx
JoshMarler: Ah yeah don't pay too much attention to this bit of drawing code :-). I copy pasted a half finished thing from elsewhere to start twiddling. Yeah totally. My issue was that it'd be nice to have an actual arc outline the same as the track fill color if possible but my software drawing skills are totaly garbage. Two arcs is the straightforward approach.

nick-thompson (changes requested):
Very cool, this looks great. Feels like we're basically at the fun part now... once the native foundation is solid enough we can live primarily in js land and just make cool stuff 💥 

Couple little things in line

examples/GainPlugin/jsui/src/ParameterSlider.js
JoshMarler: The color prop.  Yeah sorry that's a left over mistake. This is still a bit of a work in progress

packages/react-juce/src/components/Slider.tsx
JoshMarler: Yeah OK cool so keep `_onDraw` but just for the purposes of a fallback.

JoshMarler:
> Very cool, this looks great. Feels like we're basically at the fun part now... once the native foundation is solid enough we can live primarily in js land and just make cool stuff 
> 
> Couple little things in line

Yeah totally, starting to shape up nicely. 

One question, you mentioned a while back wanting internal state if the user doesn't supply a `value` prop to `Slider`. Is this still the case?

nick-thompson:
> One question, you mentioned a while back wanting internal state if the user doesn't supply a value prop to Slider. Is this still the case?

Yea, that inclination is really just to make the Slider feel a bit more familiar to `<input type="range">` for people coming from js/web. With those, you can just slap a slider there and drag it if you want, but its internal value state is decoupled from the react tree. In our case, we could basically implement the Slider such that if there is no `onChange` callback prop we just `this.setState()` with the new slider value. Then on render we read from `this.props.value` if its there or fall back to `this.state.value`. This would mean applying some default props for onDraw and the gesture map handler, which is kinda meh: do we default to rotary? Linear?

If you want to tackle that, go for it. But I'm also happy to kick the can down the road on it for now :)

examples/GainPlugin/jsui/src/App.js
```diff
@@ -51,13 +54,24 @@ class App extends Component {
...
     return (
       <View {...styles.container}>
         <View {...styles.content}>
           <Image source={require('./logo.png')} {...styles.logo} />
-          <Slider paramId="MainGain" {...styles.knob}>
-            <Label paramId="MainGain" {...styles.label} />
-          </Slider>
+          <ParameterSlider
+            paramId="MainGain"
+            sliderStrokeColor={sliderColor}
```
nick-thompson, line 68: Looks like `ParameterSlider` doesn't need these stroke/fill props, since that info is captured in the `Slider.drawLinearVertical` closure

JoshMarler:
@nick-thompson , OK would you mind running over current state of the branch. Just two small things to discuss/decide. 

1. In terms of style, for optional props like `sensitivity` we want to fallback to a "default prop". What approach do we want to use for this in our core class components? (Not sure what the current best practice for default props is). 

2. For some reason draw performance is choppy when falling back to internal state if no `value` prop is supplied. Have I done something obviously stupid here?

packages/react-juce/src/components/Slider.tsx
```diff
@@ -143,16 +143,16 @@ const _drawRotarySlider = (trackColor,
 }
 
 export interface SliderProps {
-  value: number
+  value?: number
   sensitivity?: number
   onChange: (value: number) => void
```
nick-thompson, line 148: `onChange` could be optional here too right? If you don't pass a `value` and defer the whole operation to the `Slider` itself, maybe you don't care about change callbacks?

packages/react-juce/src/components/Slider.tsx
```diff
@@ -206,6 +206,12 @@ export class Slider extends Component<PropsWithChildren<SliderProps | any>, Slid
   private _mouseDownX = 0;
   private _mouseDownY = 0;
 
+  static defaultProps = {
```
nick-thompson, line 209: 🔥

packages/react-juce/src/components/Slider.tsx
JoshMarler: Yeah I wondered about that. Is it valid to have a Slider that basically does nothing?

packages/react-juce/src/components/Slider.tsx
```diff
@@ -242,22 +248,14 @@ export class Slider extends Component<PropsWithChildren<SliderProps | any>, Slid
   _onMouseDrag(e: SyntheticMouseEvent) {
     let value = 0.0;
 
-    if (this.props.hasOwnProperty('mapDragGestureToValue') && typeof this.props.mapDragGestureToValue === 'function') {
+    if (typeof this.props.mapDragGestureToValue === 'function') {
```
nick-thompson, line 251: I wonder if we should flip this around:

```
if (typeof this.props.mapDragGesture !== 'function') {
  console.warn('You passed a weird mapping function! Abort!');
  return;
}
```

And then just carry on as usual below?

JoshMarler: Ah yeah good shout. We could adopt that across the board for non-native props like this.

packages/react-juce/src/components/Slider.tsx
nick-thompson: Totally! Maybe you're just trying to mock something up for your designer and you don't want to go the full step to wire it to any real parameters yet?

packages/react-juce/src/components/Slider.tsx
JoshMarler: Do you think log or use `invariant` ?

nick-thompson (approved):
This is doooope. Nitpicky things inline but I'm pretty much ready to merge

packages/react-juce/src/components/Slider.tsx
nick-thompson: Hm good point, let's use `invariant`

packages/react-juce/src/components/Slider.tsx
JoshMarler: Very good point. Coool. Optional onChange it is.

JoshMarler:
> This is doooope. Nitpicky things inline but I'm pretty much ready to merge

Awesome thanks for that. I'll clean up those little points and then sort the package updates. We all cool with the questionable `Canvas` performance for the moment? I'll follow this up and deal with that.

nick-thompson:
> Awesome thanks for that. I'll clean up those little points and then sort the package updates. We all cool with the questionable `Canvas` performance for the moment? I'll follow this up and deal with that.

Ah, yea, cool with it! Looking at the code here I can't see why the performance would struggle when controlled by internal state... that can definitely be a follow-up

JoshMarler:
@nick-thompson , Just pushed up what should be the final version. If you're happy to give this a quick run over to confirm you're happy with those last things I'll publish to npm and then we can hit merge.

nick-thompson (approved):
:+1: LGTM! Let's npm publish + merge

JoshMarler:
Should be good to go!

PR #135, opened 2020-11-18
Improve shadowview performance

olidacombe:
Pick up from https://github.com/nick-thompson/blueprint/pull/83 with a view to reduced layout computation calls, and seeking the most efficient layout property validation/assignment we can get.

### DIFF
--- a/blueprint/core/blueprint_ShadowView.h
+++ b/blueprint/core/blueprint_ShadowView.h
@@ -11,6 +11,8 @@
 
 #include "blueprint_View.h"
 
+#define BP_SPREAD_SETTER_PERCENT(setter) setter, setter##Percent
+#define BP_SPREAD_SETTER_AUTO(setter) BP_SPREAD_SETTER_PERCENT(setter), setter##Auto
 
 namespace blueprint
 {
@@ -89,6 +91,58 @@ namespace blueprint
         }
     };
 
+    template <typename Setter, typename ...Args>
+    const auto getYogaNodeFloatSetter(Setter setter, Args... args) {
+      return [=](const juce::var& value, YGNodeRef node) {
+        if(value.isDouble()) {
+          setter(node, args..., (float) value);
+          return true;
+        }
+        return false;
+      };
+    }
+
+    template <typename Setter, typename SetterPercent, typename ...Args>
+    const auto getYogaNodeDimensionSetter(Setter setter, SetterPercent setterPercent, Args... args) {
+      return [=, floatSetter = getYogaNodeFloatSetter(setter, args...)](const juce::var& value, YGNodeRef node) {
+        if (floatSetter(value, node))
+          return true;
+        if (value.isString() && value.toString().contains("%"))
+        {
+          juce::String strVal = value.toString().retainCharacters("-1234567890.");
+          setterPercent(node, args..., strVal.getFloatValue());
+          return true;
+        }
+        setter(node, args..., YGUndefined);
+        return true;
+      };
+    }
+
+    template <typename Setter, typename SetterPercent, typename SetterAuto, typename ...Args>
+    const auto getYogaNodeDimensionAutoSetter(Setter setter, SetterPercent setterPercent, SetterAuto setterAuto, Args... args) {
+      return [=, nonAutoSetter = getYogaNodeDimensionSetter(setter, setterPercent, args...)](const juce::var& value, YGNodeRef node) {
+        if (value.isString() && value.toString() == "auto") {
+          setterAuto(node, args...);
+          return true;
+        }
+        return nonAutoSetter(value, node);
+      };
+    }
+
+    template <typename Setter, typename EnumMap>
+    const auto getYogaNodeEnumSetter(Setter setter, EnumMap &map) {
+      return [=](const juce::var& value, YGNodeRef node) {                       \
+        const auto val = map.find(value);
+        if(val == map.end()) {
+          // TODO catch further up to add the key at which we tried
+          // to set this enum property to the message and rethrow
+          throw std::invalid_argument("Invalid property: " + value.toString().toStdString());
+        }
+        setter(node, val->second);
+        return true;
+      };
+    }
+
     //==============================================================================
     /** The ShadowView class decouples layout constraints from the actual View instances
         so that our View tree and ShadowView tree might differ (i.e. in the case of raw
@@ -111,7 +165,7 @@ namespace blueprint
 
         //==============================================================================
         /** Set a property on the shadow view. */
-        virtual void setProperty (const juce::Identifier& name, const juce::var& newValue);
+        virtual bool setProperty (const juce::String& name, const juce::var& newValue);
 
         /** Adds a child component behind the existing children. */
         virtual void addChild (ShadowView* childView, int index = -1)
--- a/blueprint/core/blueprint_ViewManager.cpp
+++ b/blueprint/core/blueprint_ViewManager.cpp
@@ -48,18 +48,16 @@ namespace blueprint
         return id;
     }
 
-    void ViewManager::setViewProperty(ViewId viewId, const juce::Identifier& name, const juce::var& value)
+    void ViewManager::setViewProperty(ViewId viewId, const juce::String& name, const juce::var& value)
     {
         const auto& [view, shadow] = getViewHandle(viewId);
 
-        view->setProperty(name, value);
-        shadow->setProperty(name, value);
-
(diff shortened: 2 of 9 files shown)


blueprint/core/blueprint_ViewManager.cpp
```diff
@@ -48,18 +48,16 @@ namespace blueprint
...
-        shadow->setProperty(name, value);
-
-        // For now, we just assume that any new property update means we
-        // need to redraw or lay out our tree again. This is an easy future
-        // optimization.
-        performRootShadowTreeLayout();
-        view->repaint();
+        // ShadowView::setProperty returns true when a layout prop
+        // has been set.  Otherwise set on the view and repaint
+        if(!shadow->setProperty(name, value)) {
```
JoshMarler, line 57: Excellent. Based on what I've seen under profiler with the current implementation this will be a big win as will removing the use of `juce::Identifier`.

blueprint/core/blueprint_ShadowView.h
```diff
@@ -89,6 +91,54 @@ namespace blueprint
...
+        return false;
+      };
+    }
+
+    template <typename Setter, typename SetterPercent, typename ...Args>
+    const auto getYogaNodeDimensionSetter(Setter setter, SetterPercent setterPercent, Args... args) {
+      return [=, floatSetter = getYogaNodeFloatSetter(setter, args...)](const juce::var& value, YGNodeRef node) {
+        if (floatSetter(value, node))
+          return true;
+        if (value.isString() && value.toString().trim().contains("%"))
```
JoshMarler, line 110: Neat! Like the macro based setter switching.

JoshMarler:
Looks great @olidacombe! I'll pull and give it a test shortly. Are we merging as is for now or did you particularly want to tackle to traversal of dirty layout nodes first?

olidacombe:
Thanks for taking a look @JoshMarler , sorry I haven't had a chance to do the profiling I wanted yet. In short, because our maps are small, I wanted to compare unordered_map to map since hashing may be slow enough that quick-to-bail comparison binary search in std::map might actually win. My hunch is unordered_map will triumph. If you are able to check it out that would be great.

I'm happy to call this a performance pass without dirty node traversal for now since I was wondering if keeping track of a latest common ancestor at "touch time" might be quicker than a tree walk of some kind on resetAfterCommit. But that would probably involve a few ShadowView data structure mods and is probably not worth fussing about, calculating once on resetAfterCommit should already be a big win.

JoshMarler:
@olidacombe, Ran a very quick profile but I'll do more thorough one tomorrow and A/B. Getting late. 

Looks like a significant improvement though. One additional win to top it off based on profiler:

64% of the time in `View::setFloatBounds` is spent in `juce::Identifier` construction. Adding some static identifier instances for `transform-rotate` rotate and friends would finish this off nicely. 

i.e. 

```
// Use a static juce::Identifer instance here as is encouraged rather than constructing Identifier from a String literal. 
props.contains(transformRotateProp)
```

I suspect the trade off between `std::map` and `std::unordered_map` will be far less significant than what you have already achieved and removing the dynamic creation of `juce::Identifier` instances.

olidacombe:
Thanks @JoshMarler that's really helpful given my current time constraints. Yes static `Identifier`s in `setFloatBounds` are a no-brainer, good spot.  https://github.com/nick-thompson/blueprint/pull/126 is making some touches there to replace `transform-rotate` altogether so I'll either add your suggestion after that is merged, or introduce the static identifier over on that branch if it's held back from merge any longer.  Fingers crossed I will get a chance to profile tonight and turn up any more oversights.

JoshMarler:
Awesome @olidacombe, Yeah no problem. @nick-thompson, based on what I've seen under the profiler I reckon a replacement of `juce::Identifier` as prop keys might be well worth doing. I suspect flat string comparisons would be better and then we could consider/investigate hashing approaches for property keys. In the office on other things and the moment but will try to get some more profile data today. Think this is good to merge though

JoshMarler:
@nick-thompson , happy for me to do a quick final sanity check of this on Win, Mac and Linux under the example plugin and then merge?

blueprint/core/blueprint_ShadowView.h
```diff
@@ -89,6 +91,54 @@ namespace blueprint
         }
     };
 
+    template <typename Setter, typename ...Args>
+    const auto getYogaNodeFloatSetter(Setter setter, Args... args) {
+      return [=](const juce::var& value, YGNodeRef node) {
+        if(value.isDouble()) {
```
nick-thompson, line 97: Do you think we should throw here?

```
if (!value.isDouble()) {
  throw std::invalid_argument("Layout properties must have numeric values.")
}
setter(node....);
return true;
```
That way we can catch this and bubble the error into js?

blueprint/core/blueprint_ShadowView.h
nick-thompson: Same! This is slick.

One quick nitpick– probably can skip the `trim()` there? It shouldn't affect finding a `"%"`

blueprint/core/blueprint_ShadowView.h
```diff
@@ -89,6 +91,54 @@ namespace blueprint
...
+        return nonAutoSetter(value, node);
+      };
+    }
+
+    template <typename Setter, typename EnumMap>
+    const auto getYogaNodeEnumSetter(Setter setter, EnumMap &map) {
+      return [=](const juce::var& value, YGNodeRef node) {                       \
+        try {
+          setter(node, map.at(value));
+        } catch(const std::out_of_range& e) { /* TODO log something */ }
```
nick-thompson, line 137: Similar here, I'm thinking we should throw from these helper setters

nick-thompson:
@olidacombe awesome work, this is really great. And I'm loving the early benchmark numbers here, thanks @JoshMarler!

> @nick-thompson , happy for me to do a quick final sanity check of this on Win, Mac and Linux under the example plugin and then merge?

Yes, please!

I just took a pass, my only thought inline is about bubbling runtime errors up and into javascript. We don't necessarily have to do it here and now but I think that's the direction we want to be moving.

Then, on a higher level, I do think we'll want to this dirty-checking pass in `resetAfterCommit` to identify where and when we to redo the layout calculation, but as @olidacombe said, that can be in a later PR.

This is awesome guys, thanks 🙏

nick-thompson (approved):

blueprint/core/blueprint_ShadowView.h
olidacombe: @nick-thompson I'm not actually sure.  If we did simply the above change, we'd break percentage based values (since we'd throw before they got tried).  If we attempt to parse % first, similarly we'd never call the base setter with a `YGValue{0.0f, YGUndefined}`, which is what currently happens if your value isn't a double or a string with `%`.  This might be fine, but I basically just converted the previous macros to something maximally template-based and logically equivalent.

I'm down with throwing on invalid values, but it might take a bit of thought (e.g. also we'd have to decide when to throw in the % case, since the existing call to juce::String::getFloatValue won't throw).

blueprint/core/blueprint_ShadowView.h
olidacombe: @nick-thompson Yes, I'll get this one in for sure

JoshMarler:
>Then, on a higher level, I do think we'll want to this dirty-checking pass in resetAfterCommit to identify where and when we to redo the layout calculation, but as @olidacombe said, that can be in a later PR.

Yeah, this is the next real killer according to the profiler. Lights up a lot in things like the `ScrollView`. Happy to take a crack at this next week. I'm busy this evening but I'll give this a final test run on remaining platforms first thing tomorrow and merge. Great work @olidacombe.

olidacombe:
Thanks @JoshMarler , much appreciated.  Apologies, I've tweaked a couple of things and squashed to be merge ready, which I realised afterwards obscures the latest changes.  But basically:
- use a static `Identifier` in `View::setFloatBounds`
- throw from `getYogaNodeEnumSetter` setters on invalid values

blueprint/core/blueprint_ShadowView.h
nick-thompson: Ahh ok, good point. Let's keep it as is then!

We can always revisit later if we find an elegant way to do it, but I think falling back to "ignore" behavior is also ok, since that's what we see in web.

JoshMarler:
@olidacombe , I'm going to close this and merge from my fork with necessary npm bumps on top of your commits.

Issue #65, opened 2020-06-18
Refine the "Multiple Bundles" story/behaviour


JoshMarler:
There are numerous areas of code in `ReactApplicationRoot` and `EcmascriptEngine` which may need to behave differently when blueprint supports the loading/evaluation of multiple bundles, i.e. for polyfill behaviour etc. 

In particular, things like error handling and hot-reloading will need careful consideration. If a single bundle causes and evaluation error, does this mean we need to reload all bundle? There may be cases for tracking/storing the bundles loaded by callers etc. 

Initially it might be good to define what all the use-cases of multiple bundle evaluation are and go from there.

nick-thompson:
:+1: Definitely. Do you think it's fair to move forward for now assuming everything is a single bundle? Until we hit a first concrete case for multiple bundles? I can guess at a few cases where multiple bundles might come in handy, but I can't make any promises that anybody will actually use it :)

JoshMarler:
Yeah. I've been hitting thoughts/considerations around multiple bundles a lot whilst looking at hot-reload and debug support. It's less pressing now I'm abandoning the idea of not resetting EcmascriptEngine in the event of a reload/error. 

I made the BundleWatcher class in ReactApplicationRoot able to handle multiple bundles for the sake of some "forward planning". I think it makes sense to assume users will load single bundles for the moment. For an initial Beta release maybe it makes sense to document this, i.e. Blueprint currently only supports loading a single bundle per ReactApplicationRoot/EcmascriptEngine instance. 

It's then probably a thing thought should just be considered and kept in the back of our minds during Blueprint development. i.e. we want to at least attempt to keep multiple bundles a possibility in the future and not make it an impossible task down the line.

nick-thompson:
Done as of #137!

PR #137, opened 2020-11-22
Refactor ReactApplicationRoot & HotReloadingHarness

nick-thompson:
This PR tackles a couple things with the primary goal of simplifying the responsibilities of ReactApplicationRoot, and facilitating more effective composition. In particular, you can now:

* Use ReactApplicationRoot with an EcmascriptEngine that you manage directly
* Use ReactApplicationRoot without hot reloading, if you want
* Use the new HotReloadingHarness for easy hot reloading with multiple bundle support (#65)
* Easily write and use your own HotReloadingHarness for nontrivial, app-specific reloading behavior.


blueprint/core/blueprint_FileWatcher.h
```diff
@@ -0,0 +1,97 @@
...
+
+        explicit FileWatcher (FileChangedCallback && callback)
+            : onFileChanged(std::move(callback)) {}
+
+        void start() { startTimer(50); }
+        void stop() { stopTimer(); }
+
+        void watch (const juce::File& f)
+        {
+            jassert(f.existsAsFile());
```
JoshMarler, line 33: Super minor but maybe we should assert `watch` is called on message thread just to avoid any daft timer disasters with people trying to load things on background threads etc.

blueprint/core/blueprint_FileWatcher.h
```diff
@@ -0,0 +1,97 @@
...
+            bool shouldFireChangeEvent = false;
+
+            // First step here is to be careful for empty files. Some bundlers
+            // may delete the target file just before replacing it with their
+            // new compiled result, and if we happen to poll in between we'll
+            // get a messed up result.
+            for (auto it = watchedFiles.begin(); it != watchedFiles.end();)
+            {
+                auto& f = (*it).first;
+                bool const fileExists = f.existsAsFile() && (f.getSize() > 0);
```
JoshMarler, line 64: Yeah nice. Like the file size check addition.

blueprint/core/blueprint_HotReloadingHarness.cpp
```diff
@@ -0,0 +1,86 @@
...
+        JUCE_ASSERT_MESSAGE_THREAD
+
+        fileWatcher = std::make_unique<FileWatcher>([this, &appRoot, ee]() {
+            appRoot.reset();
+            appRoot.bindNativeRenderingHooks();
+
+            // TODO: I could imagine support 4 hooks here: beforeAll, beforeEach, afterEach, beforeAll,
+            // where the middle two callbacks actually get specific references to _which_ bundle
+            if (beforeBundleEval)
+                beforeBundleEval(ee, juce::File());
```
JoshMarler, line 30: So in the default implementation we're just passing a default constructed `juce::File` to the callbacks? In which case, given we aren't making `HotReloadingHarness` a virtual class to be overridden by special implementations, should we just do away with the `juce::File` argument in those callbacks?

blueprint/core/blueprint_GenericEditor.cpp
```diff
@@ -141,20 +149,14 @@ namespace blueprint
 
     void BlueprintGenericEditor::registerAppRootCallbacks()
     {
-        appRoot.beforeBundleEval = [=] (const juce::File& bundle)
+        harness.beforeBundleEval = [=] (std::shared_ptr<blueprint::EcmascriptEngine> engine, const juce::File& bundle)
```
JoshMarler, line 152: Again. Maybe just kill `juce::File` for the default `HotReloadHarness` callbacks?

blueprint/core/blueprint_FileWatcher.h
```diff
@@ -0,0 +1,97 @@
...
+            {
+                onFileChanged();
+            }
+        }
+
+        // It's important here that we're using a std::map because we want the
+        // insertion order retained. In the event of a file change, we want to
+        // re-evaluate all bundles within the cleaned engine in the same order
+        // we started with.
+        std::map<juce::File, juce::Time>            watchedFiles;
```
JoshMarler, line 93: Yup nice. Makes sense.

blueprint/core/blueprint_HotReloadingHarness.cpp
```diff
@@ -0,0 +1,86 @@
...
+            if (afterBundleEval)
+                afterBundleEval(ee, juce::File());
+        });
+
+#if JUCE_DEBUG
+        fileWatcher->start();
+#endif
+    }
+
+    HotReloadingHarness::HotReloadingHarness(ReactApplicationRoot& appRoot)
```
JoshMarler, line 44: I wonder if an initializer_list of vector of initial files might make a nice addition to this constructor? Defaulted to empty maybe?

blueprint/core/blueprint_ReactApplicationRoot.cpp
```diff
@@ -9,7 +9,196 @@
...
+        bindNativeRenderingHooks();
+    }
+
+    ReactApplicationRoot::ReactApplicationRoot()
+        : ReactApplicationRoot(std::make_shared<EcmascriptEngine>())
+    {
+
+#if JUCE_DEBUG
+        // Enable keyboardFocus to support CTRL-D/CMD-D debug attachment.
+        setWantsKeyboardFocus(true);
```
JoshMarler, line 32: Should this not be moved up into the constructor above? i.e. enable the debug key bindings in both use cases of `ReactApplicationRoot` ?

blueprint/core/blueprint_ReactApplicationRoot.h
```diff
@@ -122,103 +48,32 @@ namespace blueprint
...
-            initViewManager();
-
-#if JUCE_DEBUG
-            enableHotReloading();
-
-            // Enable keyboardFocus to support CTRL-D/CMD-D debug attachment.
-            setWantsKeyboardFocus(true);
-#endif
-        }
+        ReactApplicationRoot(std::shared_ptr<EcmascriptEngine> ee);
```
JoshMarler, line 51: Maybe make this `explicit` ?

blueprint/core/blueprint_FileWatcher.h
JoshMarler: Wonder if a `std::for_each` might look a little nicer?

blueprint/core/blueprint_ViewManager.cpp
```diff
@@ -8,13 +8,45 @@
...
+
+    }
+
     ViewManager::ViewManager(View* rootView)
         : rootId(rootView->getViewId())
     {
         shadowViewTable[rootId] = std::make_unique<ShadowView>(rootView);
+
+        // Register the default view types
+        registerViewType("View", GenericViewFactory<View, ShadowView>());
```
JoshMarler, line 44: Nice. This is really tidy.

JoshMarler:
@nick-thompson , few comments inline for a first pass. 

Main thing is whether we can remove `shared_ptr` usage in a few places. Also adding hot-reload support to `ReactApplicationRoot` in the default case to avoid setup getting involved.

blueprint/core/blueprint_HotReloadingHarness.h
```diff
@@ -0,0 +1,142 @@
...
+         *
+         *  MyEditor()
+         *     : appRoot()
+         * {
+         *     juce::File myAppBundle("/path/to/myAppBundle.js");
+         *
+         *     // Not strictly required if JUCE_DEBUG set.
+         *     appRoot.enableHotReload(true);
+         *
+         *     appRoot.beforeBundleEval = [=](juce::File bundle)
```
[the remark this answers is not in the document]
nick-thompson: I think we can remove the arguments from this callback altogether, actually. Generally, I think if you want to do custom stuff with an ecmascript engine, we want to encourage you to make your own with a shared_ptr and let ReactAppRoot compose over it. I have a todo in here somewhere about potentially offering four callbacks:

```
onBeforeAll();
onBeforeEach(juce::File const& bundle);
onAfterEach(juce::File const& bundle);
onAfterAll();
```

If you want to do custom stuff with the engine in any of those callbacks, I think it's probably safe to assume the owner has their shared_ptr<EcmascriptEngine> handy (see the GenericEditor for example)

blueprint/core/blueprint_GenericEditor.cpp
```diff
@@ -27,7 +31,11 @@ namespace blueprint
 
         // Setup the ReactApplicationRoot callbacks and evaluate the supplied JS code/bundle
         registerAppRootCallbacks();
+        harness.watch(bundleFile);
+
+        beforeBundleEvaluated(nullptr);
```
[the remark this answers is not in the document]
nick-thompson: Hacks :) See my comment above. This will go away when we have a better answer to the above question

blueprint/core/blueprint_HotReloadingHarness.cpp
nick-thompson: Hacks again hahah sorry I should have explained this better... see my comment above about where I think we should take this, and then I'll come back and do it properly

blueprint/core/blueprint_ReactApplicationRoot.cpp
nick-thompson: Err I need to update the comments everywhere. I think ReactAppRoot by default will have no hot-reloading, but the thought is that if you want it, just make yourself a little `HotReloadingHarness` and kick it off.

And +1 to moving that up to the previous constructor, good catch

blueprint/core/blueprint_GenericEditor.cpp
```diff
@@ -27,7 +31,11 @@ namespace blueprint
 
         // Setup the ReactApplicationRoot callbacks and evaluate the supplied JS code/bundle
         registerAppRootCallbacks();
+        harness.watch(bundleFile);
```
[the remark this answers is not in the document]
nick-thompson: Yea, can do. I was thinking it would make sense to just _not_ use a HotReloading Harness in release mode, but that puts more work on the user. Plenty easy to add that case in here.

blueprint/core/blueprint_HotReloadingHarness.h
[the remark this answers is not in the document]
JoshMarler: OK yeah agreed. The above looks like a sane callback signature to me.

blueprint/core/blueprint_HotReloadingHarness.cpp
nick-thompson: Going to carry this conversation to a separate place!

nick-thompson:
Okayyy just pushed a bunch of updates, thanks for the detailed review. One more question inline for you!

blueprint/core/blueprint_GenericEditor.cpp
```diff
@@ -25,9 +29,18 @@ namespace blueprint
         for (auto& p : proc.getParameters())
             p->addListener(this);
 
-        // Setup the ReactApplicationRoot callbacks and evaluate the supplied JS code/bundle
-        registerAppRootCallbacks();
+        // Setup the hot reloading callbacks
+        harness.onBeforeAll = [this]() { beforeBundleEvaluated(); };
+        harness.onAfterAll = [this]() { afterBundleEvaluated(); };
+
+        // TODO: Should we just watch a bunch of files and then ask the harness to `start` manuall?
```
nick-thompson, line 36: @JoshMarler I want to carry the initializer_list comment idea in here.

Basically you can see here that, as written, I have to tell the harness to `watch` my bundle and then manually tell appRoot to evaluate. The annoying part there is that at his point, none of the bundles have _changed_ as far as the harness is concerned, so it doesn't run its `onBefore*`/`onAfter*` callbacks, which means I have this manual invocation of `beforeBundleEvaluated`/`afterBundleEvaluated` below. This feels a little clunky.

Here's the change I want to propose: instead of starting the harness for you by default, we ask that you first `watch` all the bundles you want watched, then call `start`. At that time, we run the callback hooks for you in the order expected and kick off that initial call to appRoot.evaluate.

That would mean that the following few lines look as follows:
```
harness.watch(bundleFile);
harness.start(); // Will kick off the first few callback hooks in order and call appRoot.evaluate
```
OR, supposing I have multiple bundle files:
```
harness.watch({bundleFileA, bundleFileB, bundleFileC});
harness.start(); // Will kick off the first few callback hooks in order and call appRoot.evaluate
```

What do you think?

JoshMarler: Yeah interesting. Think this API looks fine. It does force a rather different flow of events for debug/release builds if that is the distinction that people use to switch hot-reload on and off. So I do wonder if the GenericEditor should switch between the two approaches based on `JUCE_DEBUG`. If only to give and example of a likely setup, i.e. in release builds you still need to manual `beforeBundleEvaluated`, `evaluate`, `afterBundleEvaluated` chain?

blueprint/core/blueprint_HotReloadingHarness.h
```diff
@@ -0,0 +1,193 @@
...
+         *     harness.watch(myAppBundle);
+         *     appRoot.evaluate(myAppBundle);
+         *
+         *     // Not strictly required if JUCE_DEBUG set.
+         *     harness.start();
+         * }
+         *
+         * @endcode
+         **/
+        std::function<void(const juce::File&)> onAfterEach;
```
JoshMarler, line 152: Yeah these callbacks look great now

blueprint/core/blueprint_GenericEditor.cpp
nick-thompson: Hm, that's a great point...

The other option is to have `HotReloadingHarness` just _not_ run it's file watcher timer in release. For example:

```
void HotReloadingHarness::start()
{
    if (onBeforeAll) { onBeforeAll(); }

    for (const auto& f : fileWatcher->getWatchedFiles())
    {
        if (onBeforeEach) { onBeforeEach(f); }
        appRoot.evaluate(f);
        if (onAfterEach) { onAfterEach(f); }
    }

    if (onAfterAll) { onAfterAll(); }

#if JUCE_DEBUG
    if (fileWatcher)
        fileWatcher->start();
#endif
}
```

That way you don't _need_ to do anything as an end user, but again if you do need some particular behavior rewriting a custom HotReloadingHarness should be pretty easy

What do you think?

JoshMarler: Yeah I did think about that as an option. If we go with that option it should definitely be clearly documented and I wonder if a `HotReloadingHarness` is still the correct name at the point where it serves two purposes to some degree (i.e. saving the user boiler-plate). `AppReloadHarness` ?

JoshMarler: That's picky. It doesn't need to block.

nick-thompson: Yea, good point. Maybe just `AppHarness` since in release it won't actually reload anything?

I'm down with that!

JoshMarler: Hahaha yeah do it.

JoshMarler: Approved on my side with that addition.

PR #140, opened 2020-11-24
Remove additional warnings causing duktape compile to fail on Mac

JoshMarler:
@nick-thompson, Hit some additional errors on mac. Disabling these warnings prior to duktape include resolves.


nick-thompson:
Nice, thanks!

Issue #103, opened 2020-09-15
Could D.R.Y a lot of code with a few helper functions


julianstorer:
Hey folks, been noticing a lot of this kind of code, faffing around with juce::var::NativeFunctionArgs objects:

https://github.com/nick-thompson/blueprint/blob/2d6c9cfda6144134dd2c477af075dbde8d5f1268/blueprint/core/blueprint_CanvasView.h#L121

When I had to do a few of these lambda bindings, I wrote some helper functions:

https://github.com/soul-lang/SOUL/blob/7b8000930b43d387456e94d806931bfc5fb9d5cc/include/soul/patch/helper_classes/soul_patch_Blueprint.h#L97

..so that each bound function becomes just a normal method, with a real set of properly-typed parameters, not juce::var::NativeFunctionArgs, and it avoids all the repeated (and easy to screw-up) assertions for number of args, creating local variables for the args, etc.

You might want to look at using these helpers as a pattern in the codebase in general.

JoshMarler:
Yeah nice. Noticed these looking through the new soul stuff the other day. @nick-thompson, happy to look at refactoring this stuff at some point if you don't have the bandwidth right now

julianstorer:
Yeah, I'm too busy to do it, I'll leave it as an "exercise for the reader" :)

nick-thompson:
Definitely, yea, I saw that when I first liked over that SOUL commit. Very clean

@JoshMarler that would be amazing, thank you!

nick-thompson:
Done in #141

PR #141, opened 2020-11-24
Better ReactApplicationRoot method binding

nick-thompson:
Addresses #103 and cleans up a bunch of `jassert`/unpacking of `juce::var::NativeFunctionArgs`.

I already bumped react-juce@0.2.1 and published it because the change to the react-juce package is non-breaking either way.

LMK if this looks good to you @JoshMarler


JoshMarler:
@nick-thompson , Yeah nice looks great. Can't find any issues there at all. Much cleaner.

Issue #105, opened 2020-09-18
Need a workaround for Yoga refusing to build with fastmath


julianstorer:
Hit a bit of a gnarly problem, trying to add blueprint to Waveform. 

We build with fast-math enabled. Yoga refuses to work with that.

The messy part here is that it detects the __FASTMATH__ definition in the yoga _headers_, which are included by blueprint.h, and which therefore get pulled into the entire project. A decent workaround would be to disable the fast-math flag for just the blueprint cpp file, but that's not going to work because the headers will still fail in all the other TUs.

..which raises the question of why the duktape and yoga headers are included publicly in blueprint.h. Whenever we've embedded 3rd party libs like that in juce, we make a point of keeping them locked away to avoid potentially hundreds of user project files pulling them in when they're not actually needed.

Assuming that there's no reason why user-code would need to directly use any of the yoga/duktape types, it'd be a big improvement to both encapsulation and build times to get them out of the header, and use things like the Pimpl idiom where necessary to make the existing classes build.

This would also mean that we could either disable fast-math for just the blueprint.cpp file, or use some #pragmas inside that file, around the yoga includes, to selectively disable optimisation for that library

julianstorer:
..and just another note on just what a pain this is..

Looking for simple workarounds, the code does seem to run OK if you comment-out the yoga check and let it compile with fast-math on. Maybe there are layouts where the maths will go wrong, but it's probably something we could live with.

However, to hack it, this  line need to be removed:

```
#if defined(__FAST_MATH__)
#error __FAST_MATH__ defined, refusing to compile
#endif
```

And because that's in the yoga submodule, the only way I could possibly commit this to our project would be to have our own fork of blueprint which points to our own hacked fork of yoga, which is just a mess.

Another hack would be to `#undef __FAST_MATH__` just before this code gets included, but then the same problem I mentioned above comes into play, because this header gets pulled into ALL translation units, so this undef would affect code that does need it to be correct.

julianstorer:
Ah, hang on, I was getting yoga and duktape mixed up - the problematic check is in duk_config.h, and you seem to be including that directly, not via a submodule, so I can at least hack just my own fork of blueprint to bodge something that will compile. Whether it'll actually run all JS correctly is another matter of course..

[...]

Issue #62, opened 2020-06-16
Enable Custom Duktape Integrations


nick-thompson:
This issue sparked from discussion in PR #61: https://github.com/nick-thompson/blueprint/pull/61#discussion_r440796599

As Blueprint has grown, we've had to modify the duktape config inline to support the changes we need, which very much feels like (1) not the way duktape is meant to be integrated into a project and (2) it limits the end user's flexibility with duktape and Blueprint. For example, Duktape offers a compile configuration for low memory embedded environments, and there's no straightforward way to enable that in Blueprint as an end user.

This issue is meant to track this discussion and find a properly flexible way for integrating duktape

nick-thompson:
Following up on #105 and after #111 we're going to go ahead with enabling custom Duktape integrations via just letting users specify a custom include path to their duktape source via CMake, while providing the default include to the copy of duktape source included in the repo.

I want to note this PR– https://github.com/nick-thompson/blueprint/pull/102/files as we're merging a commit that comments out a fast math compile catch. I want to avoid nudging the default duktape install that we ship with the project, so once we have a proper implementation for custom duktape integrations we should reset that change in that file.

Also want to note that we should Pimpl the duktape implementation behind EcmascriptEngine to provide more flexibility, as tracked in #33.

Issue #33, opened 2019-12-17
Pimpl for EcmascriptEngine and V8 Implementation


nick-thompson:
Right now the EcmascriptEngine interface leaks (as in leaky abstraction) the `duk_context` as a private member variable. Long term I want the EcmascriptEngine to be a generic interface that can be implemented by any one of the various ES5+ compliant embedded js engines, and therefore the EcmascriptEngine interface should be totally implementation-independent. For that reason, I think it's a good idea to adopt the Pimpl pattern and move any duktape specifics into the pimpl class.

With that done, I'd love to take a stab at implementing the interface via V8 and providing the option to run Blueprint via V8 (both with and without the jit enabled): https://v8.dev/docs

nick-thompson:
See #105.

I'd like to drop the idea of support V8 or JavascriptCore for now (potentially can revisit after we hit a stable 1.0 release) and aim strictly for clean duktape integration with the pimpl encapsulation

JoshMarler:
@nick-thompson , Does this need to be beta?

nick-thompson:
It doesn't need to be, no. There are two issues here: one of them is Pimpl'ing the EcmascriptEngine so that all the Duktape specifics are hidden inside the Pimpl, which I believe should ease #62 and some of the issues that @julianstorer saw when trying to compile with FAST_MATH. Part of me thinks this would be a nice beta feature, but I don't know if it should hold us up.

The second issue, which perhaps I'll open a separate github issue for, is that once we have a Pimpl paradigm in place, I'd love add a second pimpl implementation on top of V8, and then add a CMake `-DBLUEPRINT_USE_DUKTAPE` and a `-DBLUEPRINT_USE_V8` or something so that users can flip 1 flag in their CMake file to choose their javascript engine. (Or even build with V8 for desktop and Duktape for mobile or such). That one is _definitely_ not beta, but I think it would be awesome and would love to look into it or maybe get some help on it shortly after beta

nick-thompson:
As of #145, the EcmascriptEngine interface has been cleaned up and is now properly Pimpled with Duktape underneath. I'll go ahead and mark this issue resolved, and open a separate issue for tracking a V8 Pimpl implementation

PR #145, opened 2020-12-02
EcmascriptEngine Duktape Pimpl

nick-thompson:
Bit of a big PR here, but most of the work is really just nudging some of EcmascriptEngine's private methods into a Pimpl struct, so much of this is renaming/re-indenting.

This PR touches on #62, #105, and #33, moving the Duktape implementation specifics entirely inside of `blueprint_EcmascriptEngine.cpp`. This should allow easier integration into user projects and better encapsulation of the Duktape dependency. While there, I changed the `includes` for duktape to use `<>` so that users can supply a custom include path if they want to integrate a particular Duktape build. We will continue to keep the existing Duktape source in the repo so that users who want the default can just `target_include_directories(${TARGET} PRIVATE blueprint)`.

@JoshMarler let me know what you think, I know you were considering custom Duktape integrations for some time.

And @julianstorer if you have time, I'd love to confirm that this solves the problem you hit in #105. I have re-enabled Duktape's fast math error in this PR to restore parity with the distributed source, but I believe you'll now have the flexibility to work around it as you need.

### DIFF
--- a/blueprint/blueprint.cpp
+++ b/blueprint/blueprint.cpp
@@ -16,31 +16,7 @@
  #error "Incorrect use of the Blueprint cpp file"
 #endif
 
-/* We're careful to include the duktape source files before the module header
- * file because `duktape.c` sets certain preprocessor definitions that enable
- * necessary features in the duktape header. We need those defines to preempt
- * the loading of the duktape header. This also, therefore, is the place for
- * custom preprocessor definitions.
- *
- * We force Duktape to use a time provider on Windows that is compatible with
- * Windows 7 SP1. It looks like W7SP1 is quite happy with plugins built with
- * the 8.1 SDK, but the GetSystemTimePreciseAsFileTime() call used in here is
- * just not supported without the 8.1 dll available.
- */
-#if defined (_WIN32) || defined (_WIN64)
-#define DUK_USE_DATE_NOW_WINDOWS 1
-#endif
-
-/*
- * For whatever reason it is necessary to define this to resolve errors caused by both
- * duktape and juce including parts of the winsock2 API. There may be a better way to
- * resolve this.
- */
-#if defined (_WIN32) || defined (_WIN64)
-#define _WINSOCKAPI_
-#endif
-
-// Disable compiler warnings for external source files (duktape & yoga)
+// Disable compiler warnings for external source files (yoga)
 #if _MSC_VER
   #pragma warning(push)
   #pragma warning(disable : 4018) // signed/unsigned mismatch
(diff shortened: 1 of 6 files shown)


JoshMarler:
@nick-thompson , all the code looks great. Question though, I could be wrong but I believe the cmake function `juce_add_module(blueprint)` will implicitly add `blueprint` as an include directory so I think you might get an include/header conflict with that approach?

Have you had a chance to test this? I realise duktape is currently in the module dir to comply with the JUCE module spec. 

I could be wrong about this but there might need to be a slightly different approach for overriding the `include <duktape/src-noline/duktape.h>` header paths. Worth a test. I can pull this branch later and test including custom `duk_config.h` etc.

nick-thompson:
> Question though, I could be wrong but I believe the cmake function juce_add_module(blueprint) will implicitly add blueprint as an include directory so I think you might get an include/header conflict with that approach?

Ah, interesting, I didn't see that but I can certainly imagine that would cause issues, and I only tested on MacOS.

One option could be to just move that `duktape/` source up a level in the directory structure?

JoshMarler:
Yeah. I mean could it be moved up to externals with the other submodules? Then we could more easily version track duktape. The issue here being I think this breaks the JUCE module format? We should probably experiment with this a bit.

nick-thompson:
> I mean could it be moved up to externals with the other submodules?

Yea but the issue with that is that we don't actually want duktape's git repository (IIRC), we want one of the tarball distributables.

> The issue here being I think this breaks the JUCE module format? We should probably experiment with this a bit.

Right... hm. Maybe that's an ok break? I mean, the only ask for users is that they add an include path to their build. But... obviously the more we add up these "one little thing" changes the harder the overall "getting started" flow is.

JoshMarler:
>Yea but the issue with that is that we don't actually want duktape's git repository (IIRC), we want one of the tarball distributables.

What's the reason for this out of interest? Can we not just set the submodule to one of these: 

https://github.com/svaarala/duktape/tree/v2-maintenance
https://github.com/svaarala/duktape/tree/v2.6-maintenance

>Right... hm. Maybe that's an ok break? I mean, the only ask for users is that they add an include path to their build. But... obviously the more we add up these "one little thing" changes the harder the overall "getting started" flow is.

Yeah, lemme check the shape of JUCE modules under cmake again. I'm not sure how much flexibility is allowed for dependency include paths etc.

nick-thompson:
See here: https://github.com/svaarala/duktape/tree/v2-maintenance#getting-started-end-user. Duktape just doesn't include the prebuilt source files in its git repository. If we submoduled, there would be an additional step for end-users to build the source distributable after cloning blueprint

JoshMarler:
> See here: https://github.com/svaarala/duktape/tree/v2-maintenance#getting-started-end-user. Duktape just doesn't include the prebuilt source files in its git repository. If we submoduled, there would be an additional step for end-users to build the source distributable after cloning blueprint

Ahhhh right OK. I hadn't appreciated this. Think we probably need to give this all a bit of a test. Configure our own duktape setup and see how easy/painful it is to do a custom include in user project.

jrlanglois:
I think I have a better way to handle all of the warnings, configurations for Duktape, and so on...

@nick-thompson If you don't mind holding off on this PR, I'm in the middle of a large clean-up of the module itself to make it more C++-like and more Unity build friendly as a whole.

nick-thompson:
@jrlanglois interesting, can you share what you've got in mind?

jrlanglois:
_Well_, it's one of those things that's elaborate and I didn't break it down entirely. Might be easier to push up a branch to explain/get a biew on what I'm concocting.

JoshMarler:
@nick-thompson, be cool to see this go in if we're tackling duktape as a separate issue. All the code looks tasty to me. Call me mad but I've been looking over embedding V8 docs ..... Minor concerns Duktape might never fully cut it with the GC overhead on our embedded targets and I'd be really really interested to see what performance with V8 or other engines is like.

nick-thompson:
@JoshMarler thanks! Yea, I'm happy to merge this and follow up separately with nudges for the includes + trying to optimize the config options for integrating a custom duktape build.

@jrlanglois does that sound ok to you?

jrlanglois:
+1 !

nick-thompson:
🚀 🚀

PR #149, opened 2020-12-04
WIP: Add improved CSS border style props to View

JoshMarler:
@nick-thompson, Would you mind giving this a quick pass to let me know if you think it's going roughly in the direction before I do any refactoring and quirky behaviour fixes. All the basics are working just fine with this approach. Bit of repetition that can be cleaned up but over all I think the approach came out pretty nicely and offers a decent level of customisation without going too mad.

### DIFF
--- a/blueprint/core/blueprint_View.cpp
+++ b/blueprint/core/blueprint_View.cpp
@@ -165,46 +165,22 @@ namespace blueprint
 
     void View::paint (juce::Graphics& g)
     {
-        if (props.contains("border-path"))
+        //TODO: Feels like this can probably be smarter.
+        if (props.contains("border-svg") && props["border-svg"].toString().isNotEmpty())
         {
-            juce::Path p = juce::Drawable::parseSVGPath(props["border-path"].toString());
+            const juce::String source = props["border-svg"].toString();
 
-            if (props.contains("border-color"))
-            {
-                juce::Colour c = juce::Colour::fromString(props["border-color"].toString());
-                float borderWidth = props.getWithDefault("border-width", 1.0);
+            auto drawable = juce::Drawable::createFromImageData(
+                    source.toRawUTF8(),
+                    source.getNumBytesAsUTF8());
 
-                g.setColour(c);
-                g.strokePath(p, juce::PathStrokeType(borderWidth));
-            }
-
-            g.reduceClipRegion(p);
+            //TODO: Additional work for opacity or is current setAlpha implementation enough?
+            drawable->drawWithin(g, getLocalBounds().toFloat(), juce::RectanglePlacement::centred, 1.0f);
         }
-        else if (props.contains("border-color") && props.contains("border-width"))
-        {
-            juce::Path border;
-            auto c = juce::Colour::fromString(props["border-color"].toString());
-            float borderWidth = props["border-width"];
-
-            // Note this little bounds trick. When a Path is stroked, the line width extends
-            // outwards in both directions from the coordinate line. If the coordinate
-            // line is the exact bounding box then the component clipping makes the corners
-            // appear to have different radii on the interior and exterior of the box.
-            auto borderBounds = getLocalBounds().toFloat().reduced(borderWidth * 0.5f);
-            auto width  = borderBounds.getWidth();
-            auto height = borderBounds.getHeight();
-            auto minLength = std::min(width, height);
-            float borderRadius = getResolvedLengthProperty("border-radius", minLength);
-
-            border.addRoundedRectangle(borderBounds, borderRadius);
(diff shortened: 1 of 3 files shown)


blueprint/core/blueprint_View.cpp
```diff
@@ -165,47 +165,23 @@ namespace blueprint
...
-            float borderRadius = getResolvedLengthProperty("border-radius", minLength);
-
-            border.addRoundedRectangle(borderBounds, borderRadius);
-            g.setColour(c);
-            g.strokePath(border, juce::PathStrokeType(borderWidth));
-            g.reduceClipRegion(border);
-        }
-
-        if (props.contains("background-color"))
+        else if (props.contains("background-color"))
```
nick-thompson, line 180: I don't think we want the `else` here right? You could have a border-path and a background-color at the same time no?

blueprint/core/blueprint_View.cpp
```diff
@@ -165,47 +165,23 @@ namespace blueprint
...
-                float borderWidth = props.getWithDefault("border-width", 1.0);
-
-                g.setColour(c);
-                g.strokePath(p, juce::PathStrokeType(borderWidth));
-            }
+            auto drawable = juce::Drawable::createFromImageData(
+                    source.toRawUTF8(),
+                    source.getNumBytesAsUTF8());
 
-            g.reduceClipRegion(p);
```
nick-thompson, line 181: I think we still want this `reduceClipRegion` to make sure background fills clip to the border path, right?

blueprint/core/blueprint_View.cpp
```diff
@@ -165,47 +165,23 @@ namespace blueprint
...
+            //TODO: Additional work for opacity or is current setAlpha implementation enough?
```
nick-thompson, line 177: I think people will probably use the `opacity` prop on the regular `View` for this anyway, so don't worry about it here

nick-thompson:
Yep, this looks like a good direction. I think we can probably refactor some of the js to clean up a bit, but "make it run, make it right, make it fast" and all that.

Before we go too hard this way though: I wonder, do you notice any performance implications when rendering, e.g. long lists of things that each have a border? We're definitely doing a bunch of small string concatenation here, which I think is one of duktape's performance warnings.

I'm also going to poke at some browser implementations of these border renderings just to make sure there's not some obvious alternative way of doing this that I'm totally missing. Will try to get to that this weekend

blueprint/core/blueprint_View.cpp
Prosolve: Right so my thinking here was that in the case of `border-svg` we would be painting the background in JS inside the overall border clipping path. Possible we could push that into native somehow. The assumption was that we would intercept the `background-color` prop when building that SVG.

blueprint/core/blueprint_View.cpp
Prosolve: Sort of answered this in the statement about background being handled in SVG clipping path I think.

Prosolve:
>Before we go too hard this way though: I wonder, do you notice any performance implications when rendering, e.g. long lists of things that each have a border? We're definitely doing a bunch of small string concatenation here, which I think is one of duktape's performance warnings.

Yeah I'm not sure. Totally possible. I originally started working on this in C++ as I wasn't too sure on how the performance  with a pure SVG or Canvas approach would be. I can try this today and take a look at how things are performing. It's certainly the cleanest approach as far as I can tell ..... might not be the fastest though.

blueprint/core/blueprint_View.cpp
JoshMarler: Right so my thinking here was that in the case of border-svg we would be painting the background in JS inside the overall border clipping path. Possible we could push that into native somehow. The assumption was that we would intercept the background-color prop when building that SVG.

blueprint/core/blueprint_View.cpp
JoshMarler: Think I've covered this in comment above.

JoshMarler:
>Before we go too hard this way though: I wonder, do you notice any performance implications when rendering, e.g. long lists of things that each have a border? We're definitely doing a bunch of small string concatenation here, which I think is one of duktape's performance warnings.

Yup. I tackled this in C++ originally as I figured if this is happening at the `View` level I should probably go for the quickest thing possible. The SVG approach is arguably the cleanest.....not sure if it will be the fastest. I can try to get some perf tests on this today.

JoshMarler:
@nick-thompson , Based on our recent discussions to move away from an SVG based approach for CSS border props I'm going to close this and will open up a different MR in due course which takes a more performant approach.

Issue #148, opened 2020-12-03
Make juce_audio_processors an optional thing.


jrlanglois:
Really, the juce_audio_processors module can be added by checking the module macro to see if it's available (this will mean optionally compiling your editor wrapper stuff with that).

![image](https://user-images.githubusercontent.com/1908886/101092141-2c003400-3587-11eb-868c-f28d4025fc1c.png)

nick-thompson:
Hmm yea very good point... it's only the `GenericEditor` that has this dependency. We could easily omit the includes for `GenericEditor` if we know that the user doesn't want to include `juce_audio_processors`.

>  by checking the module macro to see if it's available

I don't follow you here but this sounds like the trick! Can you share more about that?

jrlanglois:
Of course! https://github.com/nick-thompson/blueprint/pull/151

nick-thompson:
Merged with #151!

PR #151, opened 2020-12-08
Tidied up dependencies and made the juce_audio_processors optional.

jrlanglois:
See #148

### DIFF
--- a/blueprint/blueprint.h
+++ b/blueprint/blueprint.h
@@ -11,7 +11,7 @@
     name:               Blueprint
     description:        A React.js render backend targeting JUCE Components.
     minimumCppStandard: 17
-    dependencies:       juce_core, juce_graphics, juce_gui_basics
+    dependencies:       juce_gui_basics
     searchpaths:        ./duktape/src-noline/ ./yoga
    END_JUCE_MODULE_DECLARATION
 *******************************************************************************/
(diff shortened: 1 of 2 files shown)


blueprint/blueprint.h
```diff
@@ -11,7 +11,7 @@
     name:               Blueprint
     description:        A React.js render backend targeting JUCE Components.
     minimumCppStandard: 17
-    dependencies:       juce_core, juce_graphics, juce_gui_basics
+    dependencies:       juce_gui_basics
```
jrlanglois, line 14: The Projucer is smart enough to figure this out...

blueprint/blueprint.h
```diff
@@ -20,11 +20,11 @@
 
 #define BLUEPRINT_H_INCLUDED
 
-
-#include <juce_core/juce_core.h>
-#include <juce_graphics/juce_graphics.h>
 #include <juce_gui_basics/juce_gui_basics.h>
```
jrlanglois, line 23: This module recursively adds the headers that I removed above.

nick-thompson:
Ah! This makes a ton of sense @jrlanglois, thanks very much!

I'm quite ready to merge, just have one last question: will this break existing projects? I don't think so, right? Because those existing projects will have been forced already to add all of these juce module dependencies. When this merges, those existing projects wont' break, but will now have the opportunity to remove their dep on juce_audio_processors. Is that right?

jrlanglois:
It shouldn't have any effect on other projects!

> When this merges, those existing projects wont' break, but will now have the opportunity to remove their dep on juce_audio_processors. Is that right?

That's precisely the it: assuming a non-audio app is in the works, the user can simply pull that module out of their project now.

nick-thompson:
Excellent, then I'll go ahead and merge! Thanks very much @jrlanglois

Issue #146, opened 2020-12-02
Timers Polyfill


nick-thompson:
Currently blocking #122.

This task is for implementing the following global timer functions within our EcmascriptEngine environment:
* `setTimeout`
* `setInterval`
* `clearTimeout`
* `clearInterval`

The task should be easy enough: in the EcmascriptEngine constructor and in its `reset` method, we just need to `registerNativeMethod` for the above mentioned functions. The implementations should match the web spec (e.g. https://developer.mozilla.org/en-US/docs/Web/API/WindowOrWorkerGlobalScope/setTimeout, although lets ignore the raw code string variant) where possible, and we can use juce's MultiTimer for actually keeping track of these timers: https://docs.juce.com/master/classMultiTimer.html.

Essentially I think we can just hang onto a `std::unordered_map<int, juce::var::NativeFunction>`, and when we receive a call to `setTimeout` we pick a new timerId (auto incrementing integer), put the callback into our unordered_map, call startTimer, and return the timerId to the js function call. Then in `timerCallback` we just look up the appropriate callback in the map and invoke it. I'm hand-waiving a bit here over how to identify whether we should run once or indefinitely, and how to handle the args, but that shouldn't be too hard.

@olidacombe I'll assign this to you per our conversation on Discord, but if you don't want it feel free to unassign yourself!

Edit: Hm I guess I can't assign you? But maybe you can assign yourself? If you want to tackle it, please do try to assign the issue to yourself.

Also Edit: this should probably wait until #145 merges or it will be quite messy. Or rebase the work off of #145 initially.

olidacombe:
Hmmm, cool yes I want this, but can't see how to assign myself either...

JoshMarler:
Ha sorry just testing whether you could assign other users. Feel free to remove yourself @olidacombe !

nick-thompson:
Done in #188 🎉 !

PR #153, opened 2020-12-08
Timers Polyfill

olidacombe:
Address https://github.com/nick-thompson/blueprint/issues/146,
register native callbacks for `setTimeout`, `clearTimeout`, `setInterval`, `clearInterval`.

To get a quick overview, add something like this to GainPlugin / App.js
```js
  componentDidMount() {
    this.muteLogger = setInterval((m) => console.log(m, this.state.muted), 2000, "muted: ");
    this.muteFlasher = setInterval(()=>this.setState(({muted, ...rest})=>({muted: !muted, ...rest})), 50);
    setTimeout(()=>clearInterval(this.muteFlasher), 5000);
  }
```


olidacombe:
@nick-thompson there's definitely some validation  / `throw std::invalid_argument` to do, but getting this up here now so you can see how it looks.

blueprint/core/blueprint_ReactApplicationRoot.cpp
```diff
@@ -150,7 +150,9 @@ namespace blueprint
         DBG("");
         DBG(err.what());
 
-        errorText = std::make_unique<juce::AttributedString>(err.stack);
+        errorText = std::make_unique<juce::AttributedString>(err.what());
```
olidacombe, line 153: A bit of an audacious change, but `EcmascriptEngine::Error`s thrown with no `stack` rendered nothing in the red error despite having a message to work with.  I'll see about the right way to get a stack / more context from failed invocations and undo this.

olidacombe:
I've bumped up against an issue when errors are thrown inside callbacks (e.g. basic validation errors like `clearTimeout` wanting a numerical argument).  To illustrate:

```
  componentDidMount() {
    const flasher = setInterval(()=>this.setState(({muted, ...rest})=>({muted: !muted, ...rest})), 50);
    setTimeout(()=>{
      // Badness happens here
      clearTimeout("nonsense instead of" /* flasher */);
      this.setState({muted: false});
    }, 3000);
  }
```

The validation error from the bad call to `clearTimeout` is thrown from a `MultiTimer` callback, not within `ReactApplicationRoot::evaluate`, so it results in termination rather than a nice `ReactApplicationRoot::handleRuntimeError`.

For now I could log to the console and not throw.  But ultimately it might be nice to have a mechanism for pushing to a runtime error queue from outside `evaluate` calls, or something else suitable, that propagates to the `ReactApplicationRoot` error handler.  Beloved JS devs would surely want to see an informative red box instead of the door.

nick-thompson:
This is awesome @olidacombe thank you! Apologies for the delay, reviewing now + I'll see if I can find a nice solution to the error handling in there

blueprint/core/blueprint_TimeoutFunctionManager.h
```diff
@@ -0,0 +1,128 @@
...
+            if(f != timeoutFunctions.cend()) timeoutFunctions.erase(f);
+        }
+
+        void validateNewTimeoutArgs(const juce::var::NativeFunctionArgs& args)
+        {
+            if(args.numArguments < 2)
+            throw Error("setTimeout / setInterval require callback and interval arguments");
+            if(!(*args.arguments).isMethod())
+            throw Error("First argument to setTimeout / setInterval must be a callback");
+            // readVarFromDukStack returns a juce::var double for DUK_TYPE_NUMBER
```
nick-thompson, line 82: Right, all javascript numbers are doubles

blueprint/core/blueprint_TimeoutFunctionManager.h
```diff
@@ -0,0 +1,128 @@
...
+        JUCE_ASSERT_MESSAGE_THREAD
+
+        for(const auto &[id, timer] : timeoutFunctions)
+        {
+            stopTimer(id);
+        }
+        timeoutFunctions.clear();
+    }
+
+    juce::var clearTimeout(const juce::var::NativeFunctionArgs& args)
```
nick-thompson, line 19: With the API cleanup mentioned above, this could just be

```
juce::var clearTimeout(int timerId);
```

Delegate all of the error checking + stuff to the helper function for binding

blueprint/core/blueprint_TimeoutFunctionManager.h
```diff
@@ -0,0 +1,128 @@
+#pragma once
```
nick-thompson, line 1: Let's keep this struct inside the EcmascriptEngine.cpp file since, truly, it is an implementation detail of the Duktape implementation of the EcmascriptEngine.

I think we can slim this down a bit so it doesn't feel like such a big blob of code for that file

blueprint/core/blueprint_TimeoutFunctionManager.h
```diff
@@ -0,0 +1,128 @@
...
+            return id++;
+        }
+
+        void freeId(int id)
+        {
+            // TODO if we decide to get smart with reserveId
+            juce::ignoreUnused(id);
+        }
+
+        struct TimeoutFunction
```
nick-thompson, line 102: I think this helper could be quite simply:

```
struct TimerCallback {
    juce::var::NativeFunction fn;
    std::vector<juce::var> args;
    bool repeats;
};
```

I think we can use the default constructor and basically just drop the `call` method?

For example:

```
void timerCallback(int timerId) {
   if (timers.count(timerId) > 0) {
      auto const& tc = timers[timerId];
      std::invoke(tc.fn, juce::var::NativeFunctionArgs(juce::var(), tc.args.data(), static_cast<int>(tc.args.size()));

      if (!tc.repeats) {
        stopTimer(timerId);
        timers.erase(timerId);
      }
   }
}
```

Then we can drop pretty much all of the `TimeoutFunction` here, drop the `freeId` method, and the `clear` method.

What do you think?

blueprint/core/blueprint_EcmascriptEngine.cpp
```diff
@@ -348,12 +486,24 @@ namespace blueprint
...
+                {
+                    result = std::invoke(helper->callback, juce::var::NativeFunctionArgs(
+                        juce::var(),
+                        args.data(),
+                        static_cast<int>(args.size())
+                    ));
+                }
+                catch (Error& err)
+                {
+                    duk_push_current_function(ctx);
```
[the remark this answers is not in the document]
nick-thompson: Edit: ah, wait, I think this is point that we can do the helpful error handling? When we catch the error here, we handle it by pushing it into the duk stack:

```
    duk_push_error_object(ctx, DUK_ERR_TYPE_ERROR, "invalid argument: %d", (int) argvalue);
    return duk_throw(ctx);
```

Then I think we'll get the red screens for the case you mentioned?

nick-thompson (changes requested):
Awesome! Couple thoughts inline, mostly some ways I think we can simplify here. Left an idea for the error handling as well. Let me know what you think

blueprint/core/blueprint_EcmascriptEngine.cpp
```diff
@@ -278,6 +413,9 @@ namespace blueprint
             // Clear the LambdaHelper release pool as duktape does not call object
             // finalizers in the event of an evaluation error or duk_pcall failure.
             persistentReleasePool.clear();
+
+            timeoutsManager.reset();
```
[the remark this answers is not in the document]
olidacombe: Yeah initially I was going to do that, but I think on destruction of the replaced owned `TimeoutFunctionManager`, the data members (including our map of functions) would be destroyed before `~MultiTimer` gets called (which stops the callback timers).  So I skipped thinking about destroying `TimeoutFunctionManager`, but really you're right, we need to know that this class is torn down gracefully even when there's a single instance with the same lifetime as `EcmascriptEngine`.  I'll figure that out and switch to `unique_ptr`, it'll be way cleaner.

blueprint/core/blueprint_EcmascriptEngine.cpp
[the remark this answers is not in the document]
olidacombe: Cool, this stitches in stack / context without the hack below when a callback is called via normal bundle `ReactApplicationRoot::evaluate`, which is the original reason for me catching at this point (to grab a stack and re-throw for a more useful error).

However, still when callbacks are called via `timerCallback`s we get termination and no red screen at all.  I only see `handleRuntimeError` used in `ReactApplicationRoot.evaluate` and event dispatchers.  If an Error is thrown by a timer callback, there's nobody to hear it scream.

The example I gave before might not have been very clear.  Basically:

```
setTimeout("bunch of crap"); // red screen, all good

setTimeout(() => { // fine, arguments valid
  setTimeout("doom"); // no red screen, termination
}, 1000);
```

nick-thompson:
Ahh sorry! I had lost track that this was based against the `feature/engine-pimpl` branch. Let me see if I can restore it

nick-thompson:
Hmm @olidacombe would you mind trying to change the base branch to `master` and reopening? Or if that doesn't work I suppose open a new PR from your existing `setTimeout-et-al` branch onto blueprint's `master`. Sorry about that! I didn't realize merging my PR would close yours

PR #154, opened 2020-12-10
Fix ReactApplicationRoot hiding View::addChild

JoshMarler:
@nick-thompson, mind giving this a quick once over and then I'll bump npm. Fixies shadowing issues on mac

### DIFF
--- a/blueprint/core/blueprint_ReactApplicationRoot.cpp
+++ b/blueprint/core/blueprint_ReactApplicationRoot.cpp
@@ -56,9 +56,9 @@ namespace blueprint
         return juce::var::undefined();
     }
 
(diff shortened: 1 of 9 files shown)


nick-thompson:
Yup! This is great. Would you mind bumping our CMakeLists while we're here to turn on higher warning levels in our CI build? Hoping to prevent another break like that :)

Ready to merge when you've got the npm bump

JoshMarler:
@nick-thompson, mind if I play with warning settings in separate commit? I'm not totally sure what the correct warning flags to use along with the juce recommended helpers are. Yep cool will sort package bump.

nick-thompson:
Yup, no problem

JoshMarler:
@nick-thompson, Have bumped react-juce so good to merge.

PR #155, opened 2020-12-10
Apply statically initialised juce::Identifiers for View props

JoshMarler:
We've seen that dynamic creation of juce::Identifier instances
	when rendering view and setting properties shows up in profiling
	sessions. This tidies things up to use juce::Identifier as
	advices and initialises view property identifiers statically
	rather than creating Identifier instances from strings at
	runtime.

### DIFF
--- a/blueprint/core/blueprint_View.cpp
+++ b/blueprint/core/blueprint_View.cpp
@@ -82,7 +82,7 @@ namespace blueprint
     {
         props.set(name, value);
 
-        if (name == juce::StringRef ("interceptClickEvents"))
+        if (name == juce::StringRef (interceptClickEventsProp))
         {
             switch (static_cast<int> (value))
             {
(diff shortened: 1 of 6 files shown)


JoshMarler:
Ooops. Didn't add the TextView additions. Will push that up now

JoshMarler:
@nick-thompson, I think the main thing to decide here is whether or not we keep the declarations of all these identifiers public ?

blueprint/core/blueprint_View.cpp
```diff
@@ -113,7 +113,7 @@ namespace blueprint
 
     void View::setFloatBounds(juce::Rectangle<float> bounds)
     {
-        static const juce::Identifier transformMatrix("transform-matrix");
+        static const juce::Identifier transformMatrix(transformMatrixProp);
```
olidacombe, line 116: Should we skip this construction and just call `if(props.contains(transformMatrixProp))` below?

JoshMarler: Ah what a fool. Silly copy paste/replace. Good spot.

nick-thompson:
Yup! LGTM. I'm curious what the before/after profile looks like, if you happen to have the numbers around?

JoshMarler:
I don't right this moment but I'll get the data. This was driven by findings in a profiling session a while back which showed a fair amount of CPU time going down into identifier construction. I'll get some more up to date metrics.

PR #156, opened 2020-12-10
Only call ViewManager::performRootShadowTreeLayout from resetAfterCom…

JoshMarler:
…mit hook.

	Following olidacombes excellent performance improvements to the
	ShadowView layout code it appears we had some unnecessary calls
	to performRootShadowTreeLayout in the likes of
	ViewManager::addChild and ViewManager::removeChild. We should
	only need to peform a single root layout pass in the
	reconciler's resetAfterCommit hook and when the
	ReactRootApplicationInstance itself is resized. Removing these
	extra calls results in a massive performance improvement.

### DIFF
--- a/blueprint/core/blueprint_ViewManager.cpp
+++ b/blueprint/core/blueprint_ViewManager.cpp
@@ -110,7 +110,6 @@ namespace blueprint
                 if (auto* textShadowView = dynamic_cast<TextShadowView*>(parentShadowView))
                 {
                     textShadowView->markDirty();
-                    performRootShadowTreeLayout();
                 }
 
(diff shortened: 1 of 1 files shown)


nick-thompson:
Love it, thank you! And this is totally made possible thanks to @olidacombe's initial work spotting the `resetAfterCommit` improvement. Cheers to you both!

JoshMarler:
Yeah. I really just popped a cheery on top and stood on the shoulder of a giant there @olidacombe. The performance improvement is sliiiiick now.

olidacombe:
Thanks @nick-thompson , @JoshMarler this is a really nice project you have here!

Issue #138, opened 2020-11-22
Bug with hot reloading behavior


nick-thompson:
Our current hot reloading flow preserves a single EcmascriptEngine instance (to keep a debugger session alive if you have one attached) by resetting the engine and then re-evaluating the changed bundle. See: https://github.com/nick-thompson/blueprint/blob/master/blueprint/core/blueprint_ReactApplicationRoot.h#L420.

That's all good, but our current `reset()` behavior on EcmascriptEngine only clears the Duktape stack and the persistent lambdas: https://github.com/nick-thompson/blueprint/blob/master/blueprint/core/blueprint_EcmascriptEngine.cpp#L206.

The risk here is that we may have polluted the global javascript environment (by way of registering native functions or storing callback references for js<>native communication, etc) during the prior execution flow, and our `reset()` behavior doesn't address any global environment pollution.

So I think the ideal fix here is to find a way to properly reset the Duktape context or at least the duktape global object back to its default state.

nick-thompson:
@JoshMarler tagged this one for you since I think you're planning to look at it

nick-thompson:
Fixed in #157. Thanks @JoshMarler

PR #157, opened 2020-12-10
Fix broken hot-reload behaviour.

JoshMarler:
Addresses #138. Our efforts to persist a debug transport
	connection during a hot-reload resulted in an intermittent crash
	due to dangling references to global state. There is currently
	no sensible way to reset the global state in duktape without
	disposing of the duk_context. So, for now lets fix hot-reload
	and a future improvement will be to detect that we had an active
	debug session before reloading, after reload we'll then await
	debug connection again. To complement this, we'll need to fork
	the VSCode debug client and add support for a reconnect possibly
	parameterised with a timeout. Josh M also has designs on a
	duktape clion plugin.

### DIFF
--- a/blueprint/core/blueprint_EcmascriptEngine.cpp
+++ b/blueprint/core/blueprint_EcmascriptEngine.cpp
@@ -72,21 +72,7 @@ namespace blueprint
     //==============================================================================
     EcmascriptEngine::EcmascriptEngine()
     {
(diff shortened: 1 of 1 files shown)


nick-thompson:
❤️ 

Helpful context: https://github.com/svaarala/duktape/issues/2376
Fixes: #138

PR #183, opened 2020-12-16
Quick fix in the AppHarness for release builds

nick-thompson:


blueprint/core/blueprint_AppHarness.cpp
```diff
@@ -30,7 +30,23 @@ namespace blueprint
...
+                    appRoot.evaluate(f);
+                }
+                catch (const EcmascriptEngine::Error& err)
+                {
+                    // We may fall in here in Release builds when ReactApplicationRoot hasn't
+                    // caught the error for us to show the red screen. In this case, we call
+                    // a user supplied error handler if we have one, else just break the loop
+                    // and await the next file change event.
+                    if (onEvalError)
+                        onEvalError(err);
```
JoshMarler, line 45: We could consider re-throwing here if no `onEvalError` callback supplied otherwise we're failing silently which feels a bit odd. That being said I'd never personally have the file watch timer thread enabled in Release myself so I wouldn't hit this codepath.

blueprint/core/blueprint_AppHarness.h
```diff
@@ -39,9 +40,20 @@ namespace blueprint
...
         void start();
+
+        /** Stop any running file watch. */
         void stop();
 
+        /** Run the initial evaluation step exactly once.
+         *
+         *  It may be helpful to use AppHarness::start in DEBUG builds when you want
+         *  hot reloading for development, then switch in RELEASE builds to using
+         *  AppHarness::once to ignore file watching when you don't need it.
```
JoshMarler, line 53: This might be over the top but maybe an actual `#if JUCE_DEBUG` switch example inline would be useful?

blueprint/core/blueprint_AppHarness.cpp
nick-thompson: Updated!

Issue #170, opened 2020-12-11
Add module header checks for exceptions and a C++17 compatible toolchain.


jrlanglois:
Effectively, doing this would show a user the necessities at compile time!

```c++
//==============================================================================
#if JUCE_EXCEPTIONS_DISABLED
    #error "Blueprint module requires exceptions to be enabled!"
#endif

#if ! JUCE_CXX17_IS_AVAILABLE
    #error "Blueprint module requires a C++17 compatible toolchain!"
#endif
```

nick-thompson:
Ahh nice one, yes!

PR #184, opened 2020-12-16
Add module header checks for exceptions and a C++17 compatible toolchain.

jrlanglois:
See #170


nick-thompson:
Yup! :+1: thanks

PR #182, opened 2020-12-15
Added new exporters to the TestRunner and reworked the unit tests.

jrlanglois:
#171 

* The unit test structure is now entirely different: lots of code was broken and outdated...
* **The unit tests runner can now run JS code directly!**
* The TestRunner project has been updated with multiple exporters.
* The TestRunner now doesn't crash on exit.
* Fixed MSVC warnings.
* Removed the Build folder so as to avoid commit noise.

Note: looks like everything works just fine with VS2017!

---

This PR also adds unit test enabling control via the Projucer (and juce+cmake I'm guessing).
image

PR #188, opened 2020-12-17
setTimeout and friends

olidacombe:
Implement `setTimeout`, `setInterval`, `clearTimeout`, `clearInterval` for the duktape `EcmascriptEngine` using `juce::MultiTimer`.

For a demo, try adding this to the GainPlugin `App` component:
```js
componentDidMount() {
    this.muteLogger = setInterval((m) => console.log(m, this.state.muted), 2000, "muted: ");
    this.muteFlasher = setInterval(()=>this.setState(({muted, ...rest})=>({muted: !muted, ...rest})), 50);  
    this.muteFlashStopper = setTimeout(()=>{
        clearInterval(this.muteFlasher);
        this.setState({muted: false});
    }, 5000);
 }

componentWillUnmount() {
    clearInterval(this.muteLogger);
    clearInterval(this.muteFlasher);
    clearTimeout(this.muteFlashStopper);
}
```


blueprint/core/blueprint_EcmascriptEngine.cpp
```diff
@@ -252,8 +252,101 @@ namespace blueprint
             return result;
         }
 
+        struct TimeoutFunctionManager : private juce::MultiTimer
+        {
+            ~TimeoutFunctionManager() override {
+                for(const auto &[id, timer] : timeoutFunctions)
```
olidacombe, line 258: I know we're trying to keep things trim, but much like what was fixed in https://github.com/nick-thompson/blueprint/pull/182, I fear we do need at least this artefact from the slated `reset` method.

blueprint/core/blueprint_EcmascriptEngine.cpp
```diff
@@ -252,8 +252,101 @@ namespace blueprint
...
+                    std::vector<juce::var> args;
+                    const bool repeats;
+                };
+
+                std::map<int, TimeoutFunction> timeoutFunctions;
+        };
+
+        // IsSetter is true for setTimeout / setInterval
+        // and false for clearTimeout / clearInterval
+        template <bool IsSetter = false, bool Repeats = false, typename MethodType>
```
olidacombe, line 310: I toyed with an enum template parameter to make it clearer in calls what this bool refers to, but it just seemed a little excessive.  There are only ever going to be 4 calls to this thing anyway so a comment will do.

olidacombe: Also, I could only port so much of the cool stuff from https://github.com/nick-thompson/blueprint/commit/45584e748ce9db27c9e4ddf328f6343298d24935 over, purely because of the nature of having one type of call that wants to copy an unknown number of additional arguments over etc.  We've probably ended up at about the same code-weight here.  Nonetheless, I do slightly prefer this result.

blueprint/core/blueprint_EcmascriptEngine.cpp
```diff
@@ -342,12 +438,22 @@ namespace blueprint
...
 
+                juce::var result;
+
                 // Now we can invoke the user method with its arguments
-                auto result = std::invoke(helper->callback, juce::var::NativeFunctionArgs(
-                    juce::var(),
-                    args.data(),
-                    static_cast<int>(args.size())
-                ));
+                try
```
olidacombe, line 444: If this looks alien, I believe that currently if a persisted native function throws an Error (like `setTimeout` on bad arguments), you'd get the `Error.what()` message in the red component and nothing else.  This here is to also attach a stack trace to help find the bad call or other useful info.

blueprint/core/blueprint_ReactApplicationRoot.cpp
```diff
@@ -150,7 +150,9 @@ namespace blueprint
         DBG("");
         DBG(err.what());
 
-        errorText = std::make_unique<juce::AttributedString>(err.stack);
+        errorText = std::make_unique<juce::AttributedString>(err.what());
```
olidacombe, line 153: Actually, if you refer to the comment [above](https://github.com/nick-thompson/blueprint/pull/188/files#r545477266), before this change here you would get a completely empty red error box.

olidacombe:
@nick-thompson finally got around to putting this back together.  After the hefty (and wonderful) changes underfoot, I found it made more sense to start fresh with a new branch so I've tried to comment in the same places you raised points in the previous PR.

blueprint/core/blueprint_EcmascriptEngine.cpp
nick-thompson: :+1: this looks good!

blueprint/core/blueprint_EcmascriptEngine.cpp
```diff
@@ -342,12 +438,22 @@ namespace blueprint
...
+                    result = std::invoke(helper->callback, juce::var::NativeFunctionArgs(
+                        juce::var(),
+                        args.data(),
+                        static_cast<int>(args.size())
+                    ));
+                }
+                catch (Error& err)
+                {
+                    duk_push_error_object(ctx, DUK_ERR_TYPE_ERROR, err.what());
+                    return duk_throw(ctx);
```
nick-thompson, line 455: I'm not sure we want to return the `duk_ret_t` here? Maybe

```
(void) duk_throw(ctx);
```

?

blueprint/core/blueprint_EcmascriptEngine.cpp
```diff
@@ -252,8 +252,101 @@ namespace blueprint
...
+                "setInterval", &TimeoutFunctionManager::newTimeout
+            );
+            registerNativeTimerFunction("clearTimeout", &TimeoutFunctionManager::clearTimeout);
+            registerNativeTimerFunction("clearInterval", &TimeoutFunctionManager::clearTimeout);
+        }
+
         void reset()
         {
+            // Clear out any timer callbacks
+            timeoutsManager = std::make_unique<TimeoutFunctionManager>();
```
nick-thompson, line 348: :+1:

blueprint/core/blueprint_EcmascriptEngine.cpp
```diff
@@ -252,8 +252,101 @@ namespace blueprint
...
+            void timerCallback(int id) override
+            {
+                const auto f = timeoutFunctions.find(id);
+                if(f != timeoutFunctions.cend())
+                {
+                    const auto cb = f->second;
+                    std::invoke(cb.f, juce::var::NativeFunctionArgs(juce::var(), cb.args.data(), static_cast<int>(cb.args.size())));
+                    if(!cb.repeats)
+                    {
+                        stopTimer(id);
```
nick-thompson, line 288: Maybe here we can just do:

```
if (!cb.repeats) {
    (void) clearTimeout(id);
}
```

?

blueprint/core/blueprint_EcmascriptEngine.cpp
```diff
@@ -252,8 +252,101 @@ namespace blueprint
...
+                const auto f = timeoutFunctions.find(id);
+                if(f != timeoutFunctions.cend())
+                    timeoutFunctions.erase(f);
+                return juce::var();
+            }
+
+            int newTimeout(const juce::var::NativeFunction f, const int timeoutMillis, const std::vector<juce::var>&& args, const bool repeats=false)
+            {
+                static int nextId = 0;
+                timeoutFunctions.emplace(nextId, TimeoutFunction(f, std::move(args), repeats));
```
nick-thompson, line 274: :+1:

blueprint/core/blueprint_EcmascriptEngine.cpp
```diff
@@ -252,8 +252,101 @@ namespace blueprint
...
+        // IsSetter is true for setTimeout / setInterval
+        // and false for clearTimeout / clearInterval
+        template <bool IsSetter = false, bool Repeats = false, typename MethodType>
+        void registerNativeTimerFunction(const char* name, MethodType method)
+        {
+            registerNativeProperty(name, juce::var::NativeFunction([this, name, method] (const juce::var::NativeFunctionArgs& _args) -> juce::var {
+                if constexpr (IsSetter)
+                {
+                    if(_args.numArguments < 2 || !_args.arguments[0].isMethod() || !_args.arguments[1].isDouble())
+                        throw Error(juce::String(name) + " requires a callback and time in milliseconds");
```
nick-thompson, line 317: This gets caught and raised into the js env via the code you have below, right?

blueprint/core/blueprint_EcmascriptEngine.cpp
```diff
@@ -252,8 +252,101 @@ namespace blueprint
...
+                    // build a vector holding all additional arguments
+                    std::vector<juce::var> args(_args.numArguments - 2);
+                    for(auto i=2; i<_args.numArguments; i++)
```
nick-thompson, line 320: Can we use the std vector constructor `std::vector<juce::var> args (_args.arguments + 2, _args.arguments + args.numArguments)`? I don't know if the raw pointer there counts as an iterator type, so maybe not, but that seems like it'd be a tiny cleaner if we can

nick-thompson (commented):
@olidacombe this is really really nice, thanks for refactoring it. I left only a few other tiny nits/questions inline, let me know what you think, otherwise I'm very happy to merge this asap

blueprint/core/blueprint_EcmascriptEngine.cpp
olidacombe: We certainly can, but we've already retrieved an iterator above, which would be re-done by `clearTimeout`.  This is actually what I did first time around, but thought it might be the right side of DRY and performant this way.

blueprint/core/blueprint_EcmascriptEngine.cpp
olidacombe: It does, but the `duk_throw` that happens down there will only result in a red box when called from one of the `ReactApplicationRoot` methods.  If a bad call is made from within a timer callback, we get termination.

e.g.
```
setTimeout('nonsense'); // red box 👍🏾 
setTimeout(()=>setInterval('nonsense'), 1000); // termination 👎🏾 
```

So that's our dilemma, in the catch code inside `LambdaHelper` we have to decide not to re-throw to prevent this kind of termination, or to have a mechanism of pushing the `Error` to some kind of queue that `ReactApplicationRoot` is servicing (and handling with red-boxes), or stick with this possibility of termination.  I reckon somebody around here will have a much better idea than all of the above.

blueprint/core/blueprint_EcmascriptEngine.cpp
olidacombe: Ooh, I'll see if there is a slicker initialization yeah

blueprint/core/blueprint_EcmascriptEngine.cpp
olidacombe: Ah, let me check how the casting shakes out.  If it turns out the same I'm cool with making it explicit anyway, I've yet to see that being a bad idea.

blueprint/core/blueprint_EcmascriptEngine.cpp
nick-thompson: Ah ok, fair enough!

blueprint/core/blueprint_EcmascriptEngine.cpp
nick-thompson: Hmm interesting. Yea, treating this kind of error flow properly I think will be a primary focus after hitting our beta milestone. Let's not worry about it too much here, no need to make 1 issue into 2 :)

blueprint/core/blueprint_EcmascriptEngine.cpp
olidacombe: Sorry I misunderstood, but on further inspection this basically looks right as is.  I'm not a duktape expert but I think a negative `duk_ret_t` like `DUK_ERR_TYPE_ERROR` will be treated by the engine as throwing and never returning.  We are returning `duk_ret_t`, and using the `juce::var` only when not undefined, pushing to the stack and returning 1 to indicate there's a return value.

blueprint/core/blueprint_EcmascriptEngine.cpp
olidacombe: Cool, I will certainly open an issue if this gets merged.

blueprint/core/blueprint_EcmascriptEngine.cpp
olidacombe: Sorry I misunderstood, but on further inspection this basically looks right as is.  I'm not a duktape expert but I think a negative `duk_ret_t` like `DUK_ERR_TYPE_ERROR` will be treated by the engine as throwing and never returning.  We are returning `duk_ret_t`, and using the `juce::var` only when not undefined, pushing to the stack and returning 1 to indicate there's a return value.

blueprint/core/blueprint_EcmascriptEngine.cpp
olidacombe: Yeaaaah, pointers as iterators, that works exactly as you suggested!

nick-thompson:
Ok cool, this is great, thanks @olidacombe!

PR #191, opened 2020-12-22
Prettier

olidacombe:
Have prettier format js, ts, md, y?aml etc in-pipeline.

Initially set to occur on push, but could be set to occur only on PR

Could also just have it reject unprettified code instead of prettifying for you, if this feels a bit heavy-handed.  Folks pushing code without any auto-formatting at-home might be scratching their heads when they have to pull again to push the next commit.


nick-thompson:
Love it! 🔥 🚀 

Another one I think will be quite appreciated by all, cheers @olidacombe 

> Could also just have it reject unprettified code instead of prettifying for you, if this feels a bit heavy-handed. Folks pushing code without any auto-formatting at-home might be scratching their heads when they have to pull again to push the next commit.

Agreed, I'm not a fan of having CI commit to our repo on our behalf. I think we should add an `npm` script to our top-level package.json that users can use. So, we can runny prettier in CI to see if the code submitted matches the expected format. If not, fail the CI run, and the author can then run `npm run pretty` or whatever to run the prettier edit step and fix up the code.

Two other thoughts:
* Can we add a config file to define what prettier styles we want and which ones we don't? (I haven't used prettier much myself yet). I think it'd be a good idea to put that up here and discuss + align before we run it on everything
* In terms of structuring the PR, can we do 1 commit to add the github workflow, the prettier config + ignore files, and the npm script, and then 1 more commit where we group all of the changes that occur from actually running prettier on the codebase? The reason I'm thinking this is that this will be a very broad-strokes update, and any work that needs to rebase against the new edits will have a rough time. If we can isolate such edits to a single commit, any such rebase efforts will be limited to 1 commit

olidacombe:
> Agreed, I'm not a fan of having CI commit to our repo on our behalf. I think we should add an `npm` script to our top-level package.json that users can use. So, we can runny prettier in CI to see if the code submitted matches the expected format. If not, fail the CI run, and the author can then run `npm run pretty` or whatever to run the prettier edit step and fix up the code.
> 

Woops, I didn't even notice we have a top-level `package.json`!  Well in that case we can provide `lint-staged` if we want to validate before commit/push.  That will also be re-usable for `clang-format` I reckon.  In pipeline, I was thinking of making the lint actions allow failure on push, but not allow failure for PRs, so you can see your branch is warning you that a PR will fail checks.  Ultimately, I think people just enable the prettier plugin of their IDE and forget about it, the script and git hooks might see little use outside of CI.

I'm pretty committed now having messed around with this that I don't want the action to have any ability to write to the repo.

> Two other thoughts:
> 
> * Can we add a config file to define what prettier styles we want and which ones we don't? (I haven't used prettier much myself yet). I think it'd be a good idea to put that up here and discuss + align before we run it on everything

Yes, but personally I prefer to provide no customisations, since I see the job of the formatter to be keeping everyone's code and diffs predictably un-quirky, rather than enforcing my preferences.  If you suggest a first override I'll throw in the .prettierrc.  Our `clang-format` pass will or course carry strong opinions out of the gate.

> * In terms of structuring the PR, can we do 1 commit to add the github workflow, the prettier config + ignore files, and the npm script, and then 1 more commit where we group all of the changes that occur from actually running prettier on the codebase? The reason I'm thinking this is that this will be a very broad-strokes update, and any work that needs to rebase against the new edits will have a rough time. If we can isolate such edits to a single commit, any such rebase efforts will be limited to 1 commit

Yeah that's exactly what I had in mind.  This is still in playaround stages.

nick-thompson:
> Well in that case we can provide lint-staged if we want to validate before commit/push.

I'm not sure I follow? What's lint-staged?

> Yes, but personally I prefer to provide no customisations, since I see the job of the formatter to be keeping everyone's code and diffs predictably un-quirky, rather than enforcing my preferences. If you suggest a first override I'll throw in the .prettierrc. Our clang-format pass will or course carry strong opinions out of the gate.

:+1: :+1:

olidacombe:
> > Well in that case we can provide lint-staged if we want to validate before commit/push.
> 
> I'm not sure I follow? What's lint-staged?

https://github.com/okonet/lint-staged sets up some pre-commit hooks for you to run things like prettier, eslint etc.  I've configured it only to run `prettier --write`.  Even this might be too much, I've put it in there, and pretty sure it's "opt-in" (so it won't modify your commits before pushing unless you've run a command to create the hook), so thought it's safe enough.  But I'm happy to ditch it if it's a step too far.  All the cool kids are using it, which might make us wary more than anything else.

olidacombe:
@nick-thompson pending how you feel about lint-staged, currently the "first commit" leaves us in this state with the lint action failing (I turned on pull_request as a trigger for the CMake and Node actions).  The next commit would make that one green.  I don't know what people do about blame in these situations, I risk inheriting half the codebase.

olidacombe:
Hi @nick-thompson , sorry I've been mia for a good while, just revisited this hanging thread.  Not sure what your thoughts are currently on this.  I chucked some spiel in `CONTRIBUTING.md` on the last commit to illustrate the ways in which the tooling can be your friend 🐕

examples/GainPlugin/jsui/src/Meter.js
```diff
@@ -1,11 +1,5 @@
-import React, { Component } from 'react';
-import {
-  EventBridge,
-  Image,
-  Text,
-  View,
-} from 'react-juce';
-
+import React, { Component } from "react";
+import { EventBridge, Image, Text, View } from "react-juce";
```
nick-thompson, line 2: This is super nit-picky of me, but I prefer this type of multiple-import to be separated by newlines so that if you change which things you import here it's very clearly a one-liner change in the git diff, versus having to parse out which part of the long line changed.

Can prettier adapt to that? I don't want to hold this up on account of that, but so far that's my only style preference

package.json
```diff
@@ -10,9 +10,22 @@
...
+    "husky": "^4.3.8",
+    "lint-staged": "^10.5.3",
+    "prettier": "2.2.1"
   },
   "scripts": {
+    "beautify": "prettier --write .",
+    "lint": "prettier --check .",
     "serve-docs": "docsify serve docs"
+  },
+  "husky": {
```
nick-thompson, line 23: Wow, I had no idea about this tool!

nick-thompson (approved):
@olidacombe thanks for following this up, and my apolgoies as well for letting it slip by me!

I just looked through again, and I'm super happy with the shape this has taken. The default prettier format looks quite nice imo, and the tools in the top-level package.json I think will certainly make the CI pipeline and contributing process easier.

I left one quick question inline if you have any idea on that one. Otherwise, I'll go ahead and merge shortly

examples/GainPlugin/jsui/src/Meter.js
nick-thompson: I'm going to go ahead and merge, but if there's a flag we can flip for this one I may prefer that we do that in a follow-up pull request

nick-thompson:
Ah @olidacombe would you mind rebase and fixing the conflicts? Then I'll be happy to get it merged

examples/GainPlugin/jsui/src/Meter.js
olidacombe: I know what you mean, just took a look at prettier config options and nothing jumped out at me.  It will break to multi-line if the list of imports is long enough.  In the single-line case, if the changes aren't massive, I think you'll get pretty meaningful highlighting out of github by default too: https://github.blog/2014-09-04-better-word-highlighting-in-diffs/

nick-thompson:
Merged, cheers @olidacombe! Really happy about this one

olidacombe:
Hey @nick-thompson yeah I just rebased from before beautification and re-ran

JoshMarler:
Yeah this is fantastic @olidacombe. Super happy to see this go in.

PR #192, opened 2020-12-23
Listview

JoshMarler:
@nick-thompson,

This is approaching done I think. Main questions would be the following:

. I'd like to support `scrollToIndex` and `scrollToItem` functionality. Do you have am opinion on how this should look? I've been looking to ReactNative `<FlatList>` for inspiration. 

. Should we support a horizontal `ListView` or leave this as a future addition?

. Any other API functionality you would like to see here?

### DIFF
--- a/blueprint/core/blueprint_ScrollView.h
+++ b/blueprint/core/blueprint_ScrollView.h
@@ -9,8 +9,6 @@
 
 #pragma once
 
-#include <utility>
-
 #include "blueprint_View.h"
 
 namespace blueprint
@@ -22,7 +20,7 @@ namespace blueprint
         methods delegate to a single child juce::Viewport.
      */
     class ScrollView : public View
-                     , public juce::Timer
+                     , private juce::Timer
     {
         //==============================================================================
         class ScrollViewViewport : public juce::Viewport
@@ -31,7 +29,7 @@ namespace blueprint
         public:
             ScrollViewViewport() = default;
 
-            void OnAreaChanged(OnAreaChangedCallback callback)
+            void onAreaChanged(OnAreaChangedCallback callback)
             {
                 onAreaChangedCallback = std::move(callback);
             }
@@ -80,10 +78,11 @@ namespace blueprint
     private:
         //==============================================================================
         void timerCallback() override;
+        void exportNativeMethods();
 
         //==============================================================================
-        ScrollViewViewport                   viewport;
-        ScrollEvent                          lastScrollEvent;
+        ScrollViewViewport viewport;
+        ScrollEvent        lastScrollEvent;
 
         //==============================================================================
         JUCE_DECLARE_NON_COPYABLE_WITH_LEAK_DETECTOR (ScrollView)
--- a/blueprint/core/blueprint_ScrollView.h
+++ b/blueprint/core/blueprint_ScrollView.h
@@ -9,6 +9,8 @@
 
 #pragma once
 
+#include <utility>
+
 #include "blueprint_View.h"
 
 namespace blueprint
@@ -20,7 +22,39 @@ namespace blueprint
         methods delegate to a single child juce::Viewport.
      */
     class ScrollView : public View
+                     , public juce::Timer
     {
+        //==============================================================================
+        class ScrollViewViewport : public juce::Viewport
+        {
+            using OnAreaChangedCallback = std::function<void(const juce::Rectangle<int>&)>;
+        public:
+            ScrollViewViewport() = default;
+
+            void OnAreaChanged(OnAreaChangedCallback callback)
+            {
+                onAreaChangedCallback = std::move(callback);
+            }
+
+            void visibleAreaChanged(const juce::Rectangle<int> &newArea) override
+            {
+                if (onAreaChangedCallback)
+                    onAreaChangedCallback(newArea);
+            }
+
+        private:
+             OnAreaChangedCallback onAreaChangedCallback;
+        };
+
+        //==============================================================================
+        struct ScrollEvent
+        {
+            juce::var event = juce::var();
+            bool      dirty = false;
+        };
+
+        //==============================================================================
+
     public:
         //==============================================================================
         // Props following CSS Scrollbars spec with some omissions/additions.
@@ -31,6 +65,7 @@ namespace blueprint
         static inline juce::Identifier scrollbarThumbColorProp = "scrollbar-thumb-color";
         static inline juce::Identifier scrollbarTrackColorProp = "scrollbar-track-color";
         static inline juce::Identifier scrollOnDragProp        = "scroll-on-drag";
+        static inline juce::Identifier onScrollProp            = "onScroll";
 
         //==============================================================================
         ScrollView();
@@ -44,7 +79,11 @@ namespace blueprint
 
     private:
         //==============================================================================
-        juce::Viewport viewport;
+        void timerCallback() override;
+
+        //==============================================================================
+        ScrollViewViewport                   viewport;
+        ScrollEvent                          lastScrollEvent;
 
         //==============================================================================
         JUCE_DECLARE_NON_COPYABLE_WITH_LEAK_DETECTOR (ScrollView)
--- a/blueprint/core/blueprint_View.cpp
+++ b/blueprint/core/blueprint_View.cpp
@@ -265,4 +265,20 @@ namespace blueprint
             parent->dispatchViewEvent(getViewId(), eventType, e);
     }
 
+    void View::exportMethod(const juce::String &method, juce::var::NativeFunction fn)
+    {
+        JUCE_ASSERT_MESSAGE_THREAD
+        nativeMethods[method] = std::move(fn);
+    }
+
+    juce::var View::invokeMethod(const juce::String &method, const juce::var::NativeFunctionArgs &args)
+    {
+        JUCE_ASSERT_MESSAGE_THREAD
+        auto it = nativeMethods.find(method);
+
+        if (it != nativeMethods.end())
+            return it->second(args);
+
+        throw std::logic_error("Caller attempted to invoke a non-existent View method");
+    }
 }
--- a/blueprint/core/blueprint_ViewManager.cpp
+++ b/blueprint/core/blueprint_ViewManager.cpp
@@ -224,6 +224,14 @@ namespace blueprint
         shadowViewTable[rootId] = std::make_unique<ShadowView>(nh.mapped()->getAssociatedView());
     }
 
+    juce::var ViewManager::invokeViewMethod(ViewId viewId, const juce::String &method, const juce::var::NativeFunctionArgs &args)
+    {
+        if (View *view = getViewHandle(viewId).first)
+            return view->invokeMethod(method, args);
+
+        throw std::logic_error("Caller attempted to invoke method on non-existent View instance");
+    }
+
     std::pair<View*, ShadowView*> ViewManager::getViewHandle (ViewId viewId)
     {
         if (viewId == rootId)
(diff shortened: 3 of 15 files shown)


blueprint/core/blueprint_ScrollView.h
```diff
@@ -9,6 +9,8 @@
 
 #pragma once
 
+#include <utility>
```
JoshMarler, line 12: Don't think this is needed. Might have been IDE. Will clean.

blueprint/core/blueprint_ScrollView.h
```diff
@@ -20,7 +22,39 @@ namespace blueprint
         methods delegate to a single child juce::Viewport.
      */
     class ScrollView : public View
+                     , public juce::Timer
```
JoshMarler, line 25: Should be private. Will fix.

nick-thompson:
Super exciting @JoshMarler thanks for this. Lemme read up on RN's flatList and I'll take a detailed pass tomorrow

nick-thompson:
Ok coming back for a more detailed pass– the code here looks excellent, I'm quite ready to merge once we touch up the few things you've already pointed out. As for your bullets above:

> . I'd like to support scrollToIndex and scrollToItem functionality. Do you have am opinion on how this should look? I've been looking to ReactNative <FlatList> for inspiration.

Yea, let's aim for these. As for how it should look: I think functions makes more sense, like the RN API. The idea of updating a prop to force the scroll feels a little awkward since really the goal is just to mutate some internal state (specifically, the scroll position of the underlying viewport). The tricky bit here is that to implement functions like this, we need to implement something like a `ListViewInstance` in the js side of our renderer that implements `scrollToItem`, bind an appropriate C++ method to that instance (or do some ReactAppRoot top-level think for invoking a named method on some viewTable View based on the given ViewId), and then wire it into our `viewRef` prop that we added for supporting React's refs. It's maybe the last real bit of uncharted territory in the project, so I can't speak much to how difficult it will be, but it may be quite non-trivial!

> Should we support a horizontal ListView or leave this as a future addition?

Let's leave that for the future

> Any other API functionality you would like to see here?

No but an example would be huge. I suspect people will very much like this component, given that we do a lot of preset and sample browsing in our corner of software products!

JoshMarler:
>The tricky bit here is that to implement functions like this, we need to implement something like a ListViewInstance in the js side of our renderer that implements scrollToItem, bind an appropriate C++ method to that instance (or do some ReactAppRoot top-level think for invoking a named method on some viewTable View based on the given ViewId), and then wire it into our viewRef prop that we added for supporting React's refs. It's maybe the last real bit of uncharted territory in the project, so I can't speak much to how difficult it will be, but it may be quite non-trivial!

Right yeah OK. I hadn't given thought to this but I see what you're getting at. I'll digest and take a stab at this today. 

Leaving horizontal ListView for the moment is fine with me. 

>No but an example would be huge. I suspect people will very much like this component, given that we do a lot of preset and sample browsing in our corner of software products!

Yep totally. Once I've got this and the other couple of things I need finished I'll try to pick that "Component Showcase" app back up and fire over to you for a demo. We can decide if we're happy with that forming a basic example of all the stock components.

JoshMarler:
@nick-thompson, 

Noting this here for reference and notes for myself as much as anything. 

So we may be able to just switch on the `elementType` string passed to `BlueprintRenderer.createInstance` and if `"ListView"` call through to `BlueprintBackend.createListViewInstance` where `ListViewInstance` inherits `ViewInstance` and adds the additional functions bound to native?

nick-thompson:
> So we may be able to just switch on the elementType string passed to BlueprintRenderer.createInstance and if "ListView" call through to BlueprintBackend.createListViewInstance where ListViewInstance inherits ViewInstance and adds the additional functions bound to native?
>
> Assumption being we then follow roughly the approach of something like setRawTextValue in which a viewId is passed and we lookup to invoke the relevant native binding to juce::Viewport

Man, you're good. Sometimes it feels like you know this codebase better than I do!

Generally: yes, that's spot on. But there's nuance here: the distinction we have between "Views" and "Text Views" is a distinction we can safely rely on because React's reconciler also deals in this distinction. It can do so because in the Web DOM, text nodes are always truly leaf nodes in the tree, which gives the reconciler a small helpful assumption to rely on. Now, apart from View/Text, the reconciler doesn't care to differentiate any View type from any other View type, so whether we render a `ViewInstance` or a `ListViewInstance` inside of `BlueprintBackend` is of no consequence from the reconciler's point of view. The consequence shows up in the way that we communicate from a `ListViewInstance` through to the actual native `ListView` held in the ViewManager's view table. For TextView, we go right through ReactAppRoot where we can lookup the right view based on its ID, and that feels ok to me because the reconciler is careful to make this distinction between text/view nodes too.

*But* I don't know if I like this as a generic approach for tying View instance methods to methods on the native view instance (and imagine, what if I wrote a custom SpectrogramView for my product and I wanted to add a `reset()` method to it or whatever... as an end user, I don't want to fork blueprint to add my own SpectrogramViewInstance into BlueprintBackend and wire the call through ReactAppRoot). So, I wonder, can we tackle a generic approach here?

My initial idea for that is this.. suppose we wrap every `ViewInstance` returned to a user via the `viewRef` in an ES6 `Proxy` (https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Proxy). Now with the Proxy we can catch every method invocation and ask "Does the `ViewInstance` _itself_ have a method with this name?" If it does, we just invoke it on the view instance, no problem. If it doesn't, perhaps we assume that the user is trying to call a native method on the underlying native view instance. Maybe at that time we ask ReactAppRoot to dynamically invoke a named method on a given view instance with some arbitrary args list?

In detail:
https://github.com/nick-thompson/blueprint/blob/master/packages/react-juce/src/lib/BlueprintBackend.ts#L125
```
    if (propKey === 'viewRef') {
      value.current = new Proxy(this, {
            get: function(target, prop, receiver) {
                if (target.hasOwnProperty(prop))
                    return target[prop];

                // Else the user is looking up something we don't have... let's return a proxy function to native
                return function(...args) {
                    return __BlueprintNative__.invokeViewMethod(target._id, prop, ...args);
                };         
            }
      });
      return;
    }
```

Now in ReactAppRoot, we implement a native "invokeViewMethod" which uses that first argument (viewId) to look up the view, then invokes a method named by `prop` with `...args`.

To do this, I think we probably have to update the idea of `blueprint::View` to "expose" methods to this interface. For example...

```
class ListView : public blueprint::View
{
    ListView()
    {
          export("scrollToIndex", [this](juce::NativeFunctionArgs const& args) {
              scrollToIndex(args[0]); // Or something
          });
    }
}
```

Where `export` really just adds to a `std;;map<juce::String, juce::NativeFunction>` and the `View` class gives ReactAppRoot some way to look stuff up in there and call the funcs.

Whew, that was a lot, let me know if that makes sense!

JoshMarler:
>Man, you're good. Sometimes it feels like you know this codebase better than I do!

Ha! High praise. Thanks but I think I had a cheeky advantage given I wrote the viewRef code!

>But I don't know if I like this as a generic approach for tying View instance methods to methods on the native view instance (and imagine, what if I wrote a custom SpectrogramView for my product and I wanted to add a reset() method to it or whatever... as an end user, I don't want to fork blueprint to add my own SpectrogramViewInstance into BlueprintBackend and wire the call through ReactAppRoot). So, I wonder, can we tackle a generic approach here?

Yeah 100%. This was my one concern as this explodes potentially. 


> If it does, we just invoke it on the view instance, no problem. If it doesn't, perhaps we assume that the user is trying to call a native method on the underlying native view instance. Maybe at that time we ask ReactAppRoot to dynamically invoke a named method on a given view instance with some arbitrary args list?

Yeah awesome. I was thinking about a generic "invoke registered thing" but I would NEVER have known about that proxy object pattern. Gosh. More and more JS foo to learn. That's a really neat trick though. 

OK, awesome @nick-thompson. Thanks for this. As ever direction is much appreciated. I'll get to implementing this today. Seems reasonably straight forward and I much prefer this approach to adding the likes of `ListViewInstance`. 

If we assume the user is trying to invoke a native function when an object property is not present on the underlying `ViewInstance` then hopefully our usual `std::runtime_error` exception propagation is a sensible enough approach for a non-existent "native prop". 

Awesome stuff, opens up as useful pattern for future features.

JoshMarler:
@nick-thompson, OK, I think this is ready for a proper review. I'll try to get this into the showcase/example app when time allows but have put a simple example usage in `ListView`'s doc comment.

blueprint/core/blueprint_View.cpp
```diff
@@ -265,4 +265,24 @@ namespace blueprint
...
+    juce::var View::invokeMethod(const juce::String &method, const juce::var::NativeFunctionArgs &args)
+    {
+        JUCE_ASSERT_MESSAGE_THREAD
+        auto it = nativeMethods.find(method);
+
+        if (it != nativeMethods.end())
+        {
+            return it->second(args);
+        }
+        else
```
nick-thompson, line 283: Nit: we don't need the `else` clause here, because we `return` from the `if`. So can just

```
if (it != nativeMethods.end())
    return it->second(args);

throw std::logic_error....
```

blueprint/core/blueprint_ViewManager.cpp
```diff
@@ -224,6 +224,17 @@ namespace blueprint
...
     }
 
+    juce::var ViewManager::invokeViewMethod(ViewId viewId, const juce::String &method, const juce::var::NativeFunctionArgs &args)
+    {
+        if (View *view = getViewHandle(viewId).first)
+        {
+            return view->invokeMethod(method, args);
+        }
+
+        jassertfalse;
```
nick-thompson, line 234: Can we throw here and catch it in the js runtime?

packages/react-juce/src/components/ListView.tsx
```diff
@@ -0,0 +1,218 @@
...
+        <ScrollView.ContentView {...styles.scroll_view_content} height={positions.innerHeight}>
+          {items}
+        </ScrollView.ContentView>
+      </ScrollView>
+    )
+  }
+}
+
+const styles = {
+  scroll_view_content: {
```
nick-thompson, line 212: Major nitpick: camelCase please. Once we get `prettier` involved we'll catch these things automatically

blueprint/core/blueprint_ScrollView.h
```diff
@@ -20,7 +20,39 @@ namespace blueprint
...
+                     , private juce::Timer
     {
+        //==============================================================================
+        class ScrollViewViewport : public juce::Viewport
+        {
+            using OnAreaChangedCallback = std::function<void(const juce::Rectangle<int>&)>;
+        public:
+            ScrollViewViewport() = default;
+
+            void OnAreaChanged(OnAreaChangedCallback callback)
```
nick-thompson, line 32: Major nitpick: lower case for the method name's first letter please! Clang-format will save us soon

nick-thompson (approved):
@JoshMarler this is super super awesome. Love the abstraction here, and I love the pattern for native view methods. Tiny nitpicks inline, and I'll also ask to squash it down into one or two commits before merge. Otherwise, good to go!

blueprint/core/blueprint_View.cpp
JoshMarler: Yup no worries. think I'd been watching a talk advocating explicit `else` and got overly inspired!

JoshMarler:
@nick-thompson , should be good for final pass and merge. Have published 0.2.7 package.

nick-thompson:
🤘 ! Thank you!

PR #195, opened 2020-12-29
Add juce and blueprint modules in root CMakeLists.txt

stfufane:
As discussed on Discord, I moved the directives that add blueprint and juce modules from the GainPlugin directory to the root directory. 

The main issue addressed was that on Windows the symlinks are not working and it's required to copy the directories inside GainPlugin folder to compile.

With these changes, the CMake configuration is made on the root folder and the GainPlugin example does not need to add juce and blueprint. This way juce will be configured only once and it will make it easier to add new examples.


JoshMarler:
Nice stuff @stfufane, this is a better approach in my opinion. I'm no great fan of build processes which require symlinks generated by scripts etc. etc. @nick-thompson this looks good to me. I can give it a test shortly.

nick-thompson:
Great! Yea, let's do it. I just tested locally myself, all looks good

PR #196, opened 2020-12-29
Remove APVTS dependency from GenericEditor

stfufane:
Like we discussed on Discord, I removed the reference to the APVTS in the BlueprintGenericEditor class.

Instead there's a map that links the parameter IDs to their respective parameter pointers, this allows a lookup in the beforeBundleEvaluated method.

The GainPlugin has been modified accordingly and tested, it works like before, I checked in debug mode to ensure we triggered the methods in the lambda functions when moving the knob.

I tried to follow the coding style, don't hesitate to modify anything if needed :)

Cheers!

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -8,6 +8,7 @@
 
 .vscode/
 .cache/
+/build
 
 examples/TestRunner/Builds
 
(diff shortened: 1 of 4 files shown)


nick-thompson:
Excellent, thank you!

Issue #116, opened 2020-10-17
CMake + Ninja not working?


tomoyanonymous:
Hi. I found that blueprint engine does not work when building with Ninja Generator on CMake though it can be successfully compiled. ~~(Both of my project and GainPlugin Example.)~~ Update: sorry, I confirmed GainPlugin works properly.
When I inspected with a debugger,  `registerNativeRenderingHooks` function was not called on an event loop, which was normally called when built with Unix Makefiles. It finally causes an error because of the missing `window.__BlueprintNative__` variable.

```js
ReferenceError: identifier '__BlueprintNative__' undefined
  at [anon] (../Dependencies/blueprint/blueprint/duktape/src-noline/duktape.c:81729) internal
  at eval (eval:1) preventsyield
```

I could not find a specific reason but just for a bug report...

Environments:

- macOS Catalina 10.15.5
- Xcode 11.4.1
- cmake version 3.18.3(installed via homebrew)
- ninja 1.10.1(installed via homebrew)
- blueprint 5604b941e39a3e6a13633023e5b823fdd8b1269d

tomoyanonymous:
I found a problem that `__FILE__` returns absolute path when using Unix Makefiles but **Ninja returns a relative path to project-root** like `../examples/GainPlugin/PluginProcessor.cpp`.

juce::File's constructor expects an absolute path, so maybe when used with cmake+Ninja, it seems to nice to add a preprocessor definition

```cmake
target_compile_definition(your_project PRIVATE JS_MAINFILE_PATH="path/to/main.js")
```
And in cpp files,
```diff
-   File sourceDir = File(__FILE__).getParentDirectory();
-   File bundle = sourceDir.getChildFile("jsui/build/js/main.js");
+  #ifndef JS_MAINFILE_PATH
+    #define JS_MAINFILE_PATH  File(__FILE__).getParentDirectory().getChildFile("jsui/build/js/main.js");
+  #endif
+   File bundle(JS_MAINFILE_PATH);
```

nick-thompson:
Wow, interesting, thanks @tomoyanonymous! I wonder if we should be using something more standardized than `__FILE__`? I'd prefer to avoid preprocessor defs if we can, but might be a necessary move

tomoyanonymous:
I found a more clear explanation.
https://github.com/approvals/ApprovalTests.cpp/issues/74#issuecomment-572207017

> - On all platforms, the Ninja program:
[...]

PR #197, opened 2020-12-30
(#116) Avoid __FILE__ directive in gain plugin example and docs

nick-thompson:
Per the discussion in #116, I'm proposing here that we avoid `__FILE__` as it seems to have non-standard behavior. Instead, I'm using a preprocessor directive for the GainPlugin example, and I've updated the docs to alert new users that they will need to fill in the correct path to their bundle file.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -8,9 +8,10 @@
 
 .vscode/
 .cache/
-/build
 
+build
 examples/TestRunner/Builds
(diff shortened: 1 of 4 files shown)


JoshMarler:
Looks sensible to me sir!

tomoyanonymous:
Looks reasonable to me too!

PR #198, opened 2021-01-02
First pass with clang-format

nick-thompson:
Big change, but all this really does is use the clang-format file that @JoshMarler mentioned in #24 with juce-like defaults and run it against `blueprint/core/*`

Let me know what you think!


blueprint/core/blueprint_AppHarness.cpp
```diff
@@ -52,22 +56,27 @@ namespace blueprint
...
+            {
+                onAfterAll();
+            }
         });
-
     }
 
     //==============================================================================
-    void AppHarness::watch (const juce::File& f)
+    void AppHarness::watch(const juce::File& f)
```
olidacombe, line 73: Are we trying to stick to the juce style guide?  If so I think non-empty parens would have a preceding space (i.e. `SpaceBeforeParens: NonEmptyParentheses` instead of `SpaceBeforeParens: ControlStatements`)

JoshMarler: Heh. Hopefully we don't need to fully comply .... I hate that leading space.

olidacombe: Haha, such a can of worms this business.  I'm in the not caring so long as a tool is doing it for me camp.

nick-thompson: Hah, I'll admit, I'm also not a fan of that leading space. Generally though, I too am in the "i don't care let's just standardize it" camp.

I think generally if we can reach a small consensus of "yea looks good enough for me" then we should go for it

blueprint/core/blueprint_EcmascriptEngine.cpp
```diff
@@ -282,63 +293,61 @@ namespace blueprint
...
-                    return (this->timeoutsManager.get()->*method)(_args.arguments[0].getNativeFunction(), _args.arguments[1], std::move(args), Repeats);
-                }
-                else
-                {
-                    if(_args.numArguments < 1 || !_args.arguments[0].isDouble())
-                        throw Error(juce::String(name) + " requires an integer ID of the timer to clear");
-                    return (this->timeoutsManager.get()->*method)(_args.arguments[0]);
-                }
-            }));
+            registerNativeProperty(name, juce::var::NativeFunction([this, name, method](const juce::var::NativeFunctionArgs& _args) -> juce::var {
```
nick-thompson, line 327: I'm not a big fan of the way this indents the inline lambda here... is there a way to nudge this back as it was?

blueprint/core/blueprint_EcmascriptEngine.cpp
```diff
@@ -350,17 +359,16 @@ namespace blueprint
 
             // Allocate a new js heap
             dukContext = std::shared_ptr<duk_context>(
-                duk_create_heap (nullptr, nullptr, nullptr, nullptr, detail::fatalErrorHandler),
-                duk_destroy_heap
-            );
+                duk_create_heap(nullptr, nullptr, nullptr, nullptr, detail::fatalErrorHandler),
+                duk_destroy_heap);
```
nick-thompson, line 363: Also not a huge fan of pulling this trailing comma up. But my clang-format knowledge is weak here :)

nick-thompson: Ah man: https://reviews.llvm.org/D33029

Issue #84, opened 2020-08-03
Improve Color Strings Support


JoshMarler:
@nick-thompson, See the below TODO from the GainPlugin example:

```                                                                                                                                                                                                             
  // TODO: Should update the ctx proxy to convert from javascript hex strings, aka                                                                                                                                 
  // #ffffaa to juce's Colour::fromString() API which is ffffffaa.     
```

It would be great to refine Blueprint's color handling behaviours. I think you had a library in mind for this? I'd be happy to investigate if you can point me in the direction of what you wanted to use.

nick-thompson:
Yep! Definitely want this in, and I think it can be quite simple. I intend to do this on the javascript side, where in the `BlueprintBackend` we check "is the user setting a color-related property?" -> if so, transform the property value string from any arbitrar color format to JUCE's format. For example


```
<View {...{'background-color': 'hsla(55, 55, 55, 0.5)'}} />
```

That property update will trickle through to here: https://github.com/nick-thompson/blueprint/blob/master/packages/juce-blueprint/esm/lib/BlueprintBackend.js#L65

And right there we can just check if the prop key is one of a pre-defined list of color-related properties. When it is, we make the transformation.

I had been looking at this lib: https://github.com/brehaut/color-js, but I'm not so sure if that's actually the one we want. In particular, we only really care about being able to transform from one color string format (hsl, rgb, hsla, rgba, hex, etc: https://developer.mozilla.org/en-US/docs/Web/CSS/color) to juce's hex-string-with-the-alpha-chars-first format. These could be a few simple functions that we write ourselves. Now, this library that I mentioned covers that mostly, but it doesn't understand juce's string format so we'd probably have to assemble that ourselves. The other thing that it doesn't do, that I think we should aim to cover here, is supporting gradients (https://developer.mozilla.org/en-US/docs/Web/CSS/CSS_Images/Using_CSS_gradients).

So I imagine writing something like this (hand-waivy, because js doesn't have true pattern matching like this):

```
if (colorProperties.includes(propKey)) {
    __BlueprintNative__.setProperty(propKey, transformColorString(propValue));
}

function transformColorString(s) {
    match (s) {
        `hsla(h, s, l, a)` -> hexFromHSLA(h, s, l, a);
        `rgba(r, g, b, a)` -> hexFromRGBA(r, g, b, a);
        `hsl(h, s, l)` -> hexFromHSL(h, s, l);
[...]

PR #204, opened 2021-01-05
Improved CSS Color String Support

eupston:
Added css styled color string support addressing #84 
It includes:

- Hex Colors
- RGB & RGBA
- HSL & HSLA
- JUCE's named colors
- Basic linear gradients.

linear gradient functionality includes:

- Two colors
- Supporting any variation mix of css colors I.E hex, hsl, rgb, etc
- Directional words. I.E to left, to bottom right etc
- Degrees I.E 45deg, 180deg

I've also updated the Gain plugin with some examples with most of the color variations so users can see how to use the colors.
Anyway let me know if you guys have any questions or thoughts!

### DIFF
--- a/packages/react-juce/package.json
+++ b/packages/react-juce/package.json
@@ -36,6 +36,7 @@
   "dependencies": {
     "@babel/runtime-corejs3": "^7.11.2",
     "camelcase": "^6.2.0",
+    "color-string": "^1.5.4",
     "core-js": "2.6.0",
     "invariant": "^2.2.4",
     "known-css-properties": "^0.20.0",
--- a/packages/react-juce/src/lib/Backend.ts
+++ b/packages/react-juce/src/lib/Backend.ts
@@ -6,6 +6,7 @@ import SyntheticEvents, {
   SyntheticKeyboardEvent,
 } from "./SyntheticEvents";
 import { macroPropertyGetters } from "./MacroProperties";
+import Colors from "./MacroProperties/Colors";
 
 //TODO: Keep this union or introduce a common base class ViewInstanceBase?
 export type Instance = ViewInstance | RawTextViewInstance;
@@ -99,6 +100,15 @@ export class ViewInstance {
     // of a css prop, first convert it to kebab-case
     propKey = cssPropsMap[propKey] || propKey;
 
+    // convert provided color string to alpha-hex code for JUCE
+    let nativeValue;
+    if (Colors.isColorProperty(propKey)) {
+      value = Colors.colorStringToAlphaHex(value);
+      if (value.startsWith("linear-gradient")) {
+        nativeValue = Colors.convertLinearGradientStringToNativeObject(value);
+      }
+    }
+
     this._props = Object.assign({}, this._props, {
       [propKey]: value,
     });
@@ -134,9 +144,12 @@ export class ViewInstance {
         NativeMethods.setViewProperty(this._id, k, v);
       return;
     }
-
     //@ts-ignore
-    return NativeMethods.setViewProperty(this._id, propKey, value);
+    return NativeMethods.setViewProperty(
+      this._id,
+      propKey,
+      nativeValue ? nativeValue : value
+    );
   }
 
   contains(node: Instance): boolean {
--- a/packages/react-juce/src/lib/MacroProperties/Colors.ts
+++ b/packages/react-juce/src/lib/MacroProperties/Colors.ts
@@ -1,0 +1,248 @@
+import ColorString from "color-string";
+import ColorNames from "color-name";
+
+const COLOR_PROPERTIES = ["border-color", "background-color", "color"];
+
+const isColorProperty = (propKey: string): boolean => {
+  return COLOR_PROPERTIES.includes(propKey);
+};
+
+const colorStringToAlphaHex = (colorString: string): string => {
+  //From Hex
+  if (colorString.startsWith("#") && colorString.length === 7) {
+    return `ff${colorString}`.replace("#", "");
+  }
+  //From RGB or RGBA
+  else if (colorString.startsWith("rgb")) {
+    const rgbValues = colorString
+      .split("(")[1]
+      .split(")")[0]
+      .replace(/ /g, "")
+      .split(",");
+    const alphaValue =
+      rgbValues.length === 4
+        ? //@ts-ignore
+          percentToHex(100 * rgbValues.pop())
+        : "ff";
+    const hex = rgbValues.map((color) => {
+      const hex = Number.parseInt(color).toString(16);
+      return hex.length == 1 ? "0" + hex : hex;
+    });
+    return alphaValue + hex.join("");
+  }
+  //From HSL or HSLA
+  else if (colorString.startsWith("hsl")) {
+    const hslValues = colorString
+      .split("(")[1]
+      .split(")")[0]
+      .replace(/ |[%]/g, "")
+      .split(",");
+    const alphaValue =
+      hslValues.length === 4
+        ? //@ts-ignore
+          percentToHex(100 * hslValues.pop())
+        : "ff";
+    const hex = hslToHex(
+      Number.parseInt(hslValues[0]),
+      Number.parseInt(hslValues[1]),
+      Number.parseInt(hslValues[2])
+    );
+    return alphaValue + hex;
+  }
+  //From Linear Gradient
+  else if (colorString.startsWith("linear-gradient")) {
+    const lgValues = colorString
+      .split(/\((.+)/)[1]
+      .split(/\)$/)[0]
+      .split(/,+(?![^\(]*\))/)
+      .map((item) => {
+        return item.split(" ");
+      });
+    //concat any arrays that contain hsl or rba strings
+    lgValues.forEach((itemArr, idx) => {
+      if (itemArr.find((item) => item.includes("(")) !== undefined) {
+        const itemArrClone = itemArr.slice();
+        const startIdx = itemArrClone.findIndex((item) => item.includes("("));
+        const endIdx = itemArrClone.findIndex((item) => item.includes(")"));
+        const colorHSLRGB = itemArrClone.slice(startIdx, endIdx + 1).join("");
+        itemArr.splice(startIdx, endIdx + 1, colorHSLRGB);
+        lgValues[idx] = itemArr;
+      } else {
+        lgValues[idx] = itemArr;
+      }
+    });
+    //flatten array
+    const lgValuesCleaned: Array<String> = [].concat
+      // @ts-ignore
+      .apply([], lgValues)
+      .filter((item) => {
+        return item != "";
+      });
+    //converts any colorStrings in linear gradient to alpha-hex
+    let skipCurrentIdx = -1;
+    const lgValuesConverted = lgValuesCleaned
+      .map((value, idx) => {
+        if (skipCurrentIdx == idx) return;
+        // @ts-ignore
+        let convertedColorString = colorStringToAlphaHex(value);
+        //check if the next element is a percent and attach to current and remove it
+        const nextIdx = idx + 1;
+        if (
+          lgValuesCleaned[nextIdx] !== undefined &&
+          lgValuesCleaned[nextIdx].includes("%") &&
+          !lgValuesCleaned[nextIdx].includes("(")
+        ) {
+          convertedColorString += lgValuesCleaned[nextIdx];
+          skipCurrentIdx = nextIdx;
+        }
+        return convertedColorString;
+      })
+      .filter((item) => {
+        return item !== undefined;
+      });
+    return "linear-gradient(" + lgValuesConverted.join(",") + ")";
+  }
+  //From Named Colours
+  else if (Object.keys(ColorNames).includes(colorString)) {
+    const rgbValues = ColorString.get.rgb(colorString);
+    const hex = ColorString.to.hex(rgbValues);
+    return colorStringToAlphaHex(hex);
+  } else {
+    return colorString;
+  }
+};
+
+const convertLinearGradientStringToNativeObject = (
+  lgColorStringHex: string
+): object => {
+  const linearGradientNativeObject = {};
+  const lgValues = lgColorStringHex
+    .split(/\((.+)/)[1]
+    .split(/\)$/)[0]
+    .replace(/ /g, "")
+    .split(/,+(?![^\(]*\))/);
+  //Check for directional words
+  let angle: number = 0;
+  let rotation: string = lgValues[0];
+  if (rotation == "toleft") angle = 270;
+  if (rotation == "toright") angle = 90;
+  if (rotation == "tobottom") angle = 180;
+  if (rotation == "totop") angle = 0;
+
+  //These are not the exact match to CSS as boundary dimensions are required for calculation
+  if (rotation == "totopleft" || rotation == "tolefttop") angle = 295;
+  if (rotation == "totopright" || rotation == "torighttop") angle = 45;
+  if (rotation == "tobottomright" || rotation == "torightbottom") angle = 115;
+  if (rotation == "tobottomleft" || rotation == "toleftbottom") angle = 205;
+
+  //Check for Turns
+  if (rotation.includes("turn")) {
+    rotation = rotation.replace("turn", "");
+    const turns: number = parseFloat(rotation);
+    angle = 360 * turns;
+  }
+  //Check for Degrees
+  if (rotation.includes("deg")) {
+    rotation = rotation.replace("deg", "");
+    angle = parseInt(rotation);
+  }
+
+  linearGradientNativeObject["angle"] = angle;
+  linearGradientNativeObject["colours"] = [];
+  const colorPositions: Array<String> = lgValues.slice(1);
+
+  colorPositions.forEach((colorPos, idx) => {
+    const hexPosObj = { id: idx };
+    const hex = colorPos.slice(0, 8);
+    //check string is an alpha hex
+    const isValidAlphaHex =
+      // @ts-ignore
+      hex.match(
+        /^[0-9a-fA-F]{8}$|#[0-9a-fA-F]{6}$|#[0-9a-fA-F]{4}$|#[0-9a-fA-F]{3}$/
+      ).length > 0;
+    if (!isValidAlphaHex) return;
+    colorPos = colorPos.replace(hex, "");
+    if (colorPos.includes("%")) {
+      colorPos = colorPos.replace("%", "");
+      // @ts-ignore
+      let colorPercent = parseInt(colorPos) / 100;
+      //check if previous percent is greater or equal to the current
+      if (linearGradientNativeObject["colours"].length > 0) {
+        const previousPos =
+          linearGradientNativeObject["colours"][idx - 1]["position"];
+        if (previousPos >= colorPercent) colorPercent = previousPos + 0.001;
+      }
+      hexPosObj["position"] = colorPercent;
+    } else {
+      if (idx == 0) {
+        hexPosObj["position"] = 0.0;
+      } else if (idx == colorPositions.length - 1) {
+        hexPosObj["position"] = 1.0;
+      }
+      //assign null and iterate over again later
+      else {
+        hexPosObj["position"] = undefined;
+      }
+    }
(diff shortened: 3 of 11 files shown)


JoshMarler:
@eupston, apologies for the delay reviewing. This looks like a reasonably involved change but looks like great stuff. I should be able to take a pass over this properly tomorrow evening but this might be a PR better suited to @nick-thompson's knowledge than mine.

nick-thompson:
Hey @eupston apologies from me as well for the delay! Looking over it right now

blueprint/core/blueprint_CanvasView.h
```diff
@@ -123,8 +123,11 @@ namespace blueprint
                         //TODO: Implement fillStyle pattern
                         //TODO: Implement fillStyle gradient
 
-                        const auto colourHex = args.arguments[0].toString();
-                        properties.fillStyle.setColour(juce::Colour::fromString(colourHex));
+                        const auto colorString = args.arguments[0].toString();
+                        juce::Colour colour = juce::Colour::fromString(colorString);
+                        colour = juce::Colours::findColourForName(colorString, colour);
```
nick-thompson, line 128: I'm :+1: for named color lookup, but I think we should do it on the javascript side.

My general philosophy is that we should write absolutely minimal native C++ when we can, because if we can ship helpful change without asking end users to recompile, I think that's a big win (especially for something like SOUL, where it would require that SOUL update and recompile their binaries _and_ for all SOUL end users to go get that update).

So, via that philosphy, I think our color approach should generally be: drastically simplify the language that our native API speaks (maybe we only ever take color strings of the form `ffaabbcc`, and then it's javascript business for getting from various color types `hsl`, `rgba`, `#hex` etc to the unified native color type).

Maybe it's worth me explaining that philosophy in the contributors guide? What do you think?

blueprint/core/blueprint_View.cpp
```diff
@@ -202,10 +206,123 @@ namespace blueprint
 
         if (props.contains(backgroundColorProp))
         {
-            juce::Colour c = juce::Colour::fromString(props[backgroundColorProp].toString());
-
-            if (!c.isTransparent())
-                g.fillAll(c);
+            auto colorString = props[backgroundColorProp].toString();
+            if(colorString.contains("linear-gradient")) {
```
nick-thompson, line 210: Similarly here, my hunch is that we should address linear-gradient as much as possible in javascript. I think perhaps we should define a minimal native API for dealing in color gradients, and then tackle `linear-gradient` as a function which maps to this minimal API.

For example, maybe our native API is this:

If the property value is a color type, we expect
* If the value type is a string, then it is of the form `ffaabbcc` and we can construct a `juce::Colour` directly
* If the value type is a dynamic object (juce::var::isDynamicObject()), then it is of the form
```
{
  type: 'linear',
  startSide: 'left',
  angle: 0.139, // perhaps we define that this number is always radians
  colourStops: [
    'ffaabbcc', // These are simply strings of the expected juce hex format
    'ffbbaacc',
  ]
}
```

Then native's only business is "is the value type string or object?" and "how do we go from string -> color (easy), and how do we go from object->colourGradient (harder, but you've done all the hard work here already)"

packages/react-juce/src/lib/BlueprintBackend.ts
```diff
@@ -123,6 +168,21 @@ export class ViewInstance {
     // of a css prop, first convert it to kebab-case
     propKey = cssPropsMap[propKey] || propKey;
 
+    // convert provided color string to alpha-hex code for JUCE
+    if(propKey.includes("color")){
```
nick-thompson, line 172: Yep, right idea exactly.

Instead of searching for `color` though I think we should maintain a list of known color properties. For example,

```
border-color
background-color
color
```

and do a quick `if (isColorProperty(propKey))` here. That way if the end user writes a custom native component with some custom property named `tshirtColor` or something we don't trip this check accidentally

packages/react-juce/src/lib/BlueprintBackend.ts
```diff
@@ -123,6 +168,21 @@ export class ViewInstance {
...
+      //converts any colorStrings in linear gradient to alpha-hex
+      if(value.includes("linear-gradient")) {
+        const lgValues = value.split(/\((.+)/)[1].split(/\)$/)[0].replace(/ /g, "").split(/,+(?![^\(]*\))/);
+        const lgValuesConverted = lgValues.map(value => {
+          return colorStringToAlphaHex(value);
+        })
+        value = "linear-gradient(" + lgValuesConverted.join(",") + ")"
```
nick-thompson, line 179: Yea exactly, you're on the right path here. My suggestion from above is basically just to take it one step further here before we pass it over to the native API, doing our start side + angle calculations in here.

nick-thompson (changes requested):
Yup @eupston this is excellent work, very much the right idea.

Left a couple of thoughts inline, let me know what you think!

blueprint/core/blueprint_CanvasView.h
eupston: awesome thanks for feedback @nick-thompson

Ahh yup that's smart then easy as `npm update react-juce` for updates. Yeah sure quick note in contributors guide couldn't hurt.

So for the named colors should we maintain our own map of JUCE names to hex value in JS land then? If so any particular place you'd like to keep them? Seems like we may need to create our own Colors class in JS to manage the conversion logic and mappings etc.

JoshMarler: Yeah totally agreed. Advising this as the preferred approach in the contributors guide sounds great. I too tried to do too much in C++ in my first contributions.

blueprint/core/blueprint_View.cpp
eupston: ahh I got you. Yes I like the cleanliness of dynamic objects. The JUCE ColourGradient is relatively simple class, provide two colors and a pair of coordinates for each, so maybe we leave the native API as simple as that? As far as 3+ colors we can use the  `addColour()` method which only takes a relative position along the gradient line. So maybe the object we pass to native is:

```
{
  colour1: {
    color: ffaabbcc
    x: 0
    y: 0
  },
  color2: {
    color: ffaabbcc
    x: 100
    y: 100
  }
  additionalColors: [
    {
      color: ffaabbcc
      position: 0.2
    },
    {
      color: ffaabbcc
      position: 0.8
    }
  ]
}
```
We can do all logic in JS then. The only tricky thing is getting the bounds of the component in JS for the calculations before rendering. Do we have a way to do this?

packages/react-juce/src/lib/BlueprintBackend.ts
eupston: ah sweet makes sense will make this change shortly

blueprint/core/blueprint_CanvasView.h
nick-thompson: > Ahh yup that's smart then easy as npm update react-juce for updates.

Exactly :)

> So for the named colors should we maintain our own map of JUCE names to hex value in JS land then? If so any particular place you'd like to keep them? Seems like we may need to create our own Colors class in JS to manage the conversion logic and mappings etc.

Yea, should be easy enough. Actually I just took a quick peek at npm to see if anything existed already, and check this out: https://www.npmjs.com/package/color-string. I wasn't aware of this before, but I think we could simplify much of our javascript side of things with this guy as well. That should solve taking named colors as well as hsl/a, rgb/a, etc and then we can map the resulting numeric array back into our own JUCE-hex style:

```js
var numbers = colorString.get.rgb('blue'); // [0, 0, 255, 1]
var hex = numbers.map(numericToHex);
var juceHex = `${hex[3]}${hex[0]}${hex[1]}${hex[2]}`;
```

blueprint/core/blueprint_View.cpp
nick-thompson: Yea good question. So, I think generally for a linear gradient we need to know two things: the angle of the gradient axis, and where the color stops are. I wonder... can we arrive at something like this for the native API?

```js
{
    angle: 1.923, // radians
    colours: [
      { hex: 'ffaabbcc', position: 0.0 },
      { hex: 'ffaabbdd', position: 0.6 },
      { hex: 'ffaaccdd', position: 1.0 },
      // ...etc
    ]
}
```

Here's my thinking: `angle` and `startSide` in the web API generally tell us how to orient this axis. But, interestingly, `startSide: left, angle: 0` is equivalent to `startSide: right, angle: 3.14`. So, I wonder, can we map _all_ notions of the gradient axis onto a single line rotated some angle from the origin?

I think we can probably map this whole thing onto something of a "unit line" where we let 0.0 denote the start of the gradient axis and 1.0 denote the end of the gradient axis, then use the angle to spin that around as needed. Now, all `position` values on each color are relative points along that unit line: 0.25 is 25% along the line, 0.8 is 80%, etc. This way, once we get into native land we can simply do (pseudocode to follow)

* Let x0 = getWidth();
* Let y0 = getHeight() / 2;
* Let x1 = 0;
* Let y1 = y0;
* Let axisStartPoint = rotate({x0, y0}, angle);
* Let axisEndPoinit = rotate({x1, y1}, angle);
* Let gradient = juce::ColourGradient(colors[0]['hex'], x0, y0, colors[1]['hex'], x1, y1));
* For each color in {colors[2]...colors[n]} gradient.addColour(colors[i]['hex'], colors[i]['position']['x;] * (x1 - x0), colors[i]['position']['x;] * (y1 - y0))

That's very crude pseudocode, but hopefully it gets the point across :D

What do you think? This let's us minimize the native API while also keeping everything abstract/relative until we get into native, at which point measuring the component is trivial

blueprint/core/blueprint_CanvasView.h
eupston: > Yea, should be easy enough. Actually I just took a quick peek at npm to see if anything existed already, and check this out: https://www.npmjs.com/package/color-string. I wasn't aware of this before, but I think we could simplify much of our javascript side of things with this guy as well. That should solve taking named colors as well as hsl/a, rgb/a, etc and then we can map the resulting numeric array back into our own JUCE-hex style:
> 
> ```js
> var numbers = colorString.get.rgb('blue'); // [0, 0, 255, 1]
> var hex = numbers.map(numericToHex);
> var juceHex = `${hex[3]}${hex[0]}${hex[1]}${hex[2]}`;
> ```
Nice find! Shame I didn't find it early haha. Yup looks like this will do everything we need and named colors seems to match JUCE 1 for 1 with the [color-name](https://github.com/colorjs/color-name/blob/master/index.js) package. 
I'll look into adding that tomorrow.

blueprint/core/blueprint_View.cpp
eupston: > I think we can probably map this whole thing onto something of a "unit line" where we let 0.0 denote the start of the gradient axis and 1.0 denote the end of the gradient axis, then use the angle to spin that around as needed. Now, all `position` values on each color are relative points along that unit line: 0.25 is 25% along the line, 0.8 is 80%, etc. This way, once we get into native land we can simply do (pseudocode to follow)
> 

Yeah I think you're right, we could make it that simple. 
I know for sure that'll get us at least the basic linear gradient combinations (angles & directional words).

> * Let x0 = getWidth();
> * Let y0 = getHeight() / 2;
> * Let x1 = 0;
> * Let y1 = y0;
> * Let axisStartPoint = rotate({x0, y0}, angle);
> * Let axisEndPoinit = rotate({x1, y1}, angle);
> * Let gradient = juce::ColourGradient(colors[0]['hex'], x0, y0, colors[1]['hex'], x1, y1));
> * For each color in {colors[2]...colors[n]} gradient.addColour(colors[i]['hex'], colors[i]['position']['x;] * (x1 - x0), colors[i]['position']['x;] * (y1 - y0))
> 
Ah yup think I see your point (no pun intended).  Only change would be gradient.addColour only takes a percentage along the gradient line not coordinates. So JUCE should actually do calculation work for us there. I'm assuming the first color/point added will always be considered the zero position, but will have a play to confirm this. 

Anyway sounds like a plan, i'll start looking into it tomorrow. I think this design should actually simplify the calculation logic too.

nick-thompson: > Only change would be gradient.addColour only takes a percentage along the gradient line not coordinates. So JUCE should actually do calculation work for us there. I'm assuming the first color/point added will always be considered the zero position, but will have a play to confirm this.

Ah, right, yea that's convenient actually, since then we can basically take the relative position verbatim and just scale it up to percentage.

Yea, great! Thanks for tackling it :D

eupston:
You're making me do math again @nick-thompson! :P

I've updated with all discussion points mentioned above.
In short we can now do the following:

- Hex Colors
- RGB & RGBA
- HSL & HSLA
- JUCE/HTML named colors

linear gradient functionality includes:

- Multiple color support
- Supporting any variation mix of css colors I.E hex, hsl, rgb, etc
- Directional words. I.E to left, to bottom right etc
- Degrees I.E 45deg, 180deg
- Turns I.E 0.5turn 1.25turn
- Percent along the gradient I.E red 40% blue 60%

So we can do any of these linear-gradient combinations:

- `linear-gradient(45deg, red, blue)`
- `linear-gradient(0.25turn, red 25%, blue 75%)`
- `linear-gradient(to left top, red -25%, blue 125%)`
- `linear-gradient(to top left, red 25%, hsla(225, 15%, 11%, 0.3),  rgba(102, 253, 207, 1) 125%)`
- `linear-gradient(to left, #17191f,  #626262 70%,  #17191f)`

I also found this medium [article](https://medium.com/@patrickbrosset/do-you-really-understand-css-linear-gradients-631d9a895caf
) very helpful for explaining how linear-gradients are calculated, and I have now updated the linear gradient line to the true coordinates.

anyway let me know whatcha think!

nick-thompson:
Really awesome @eupston thanks a ton. Will review in detail shortly!

packages/react-juce/package.json
```diff
@@ -17,7 +17,6 @@
     "@babel/cli": "^7.10.0",
     "@babel/core": "^7.11.1",
     "@babel/plugin-transform-runtime": "^7.11.0",
-    "@babel/plugin-proposal-class-properties": "^7.11.0",
```
nick-thompson, line 20: Seems like maybe an unrelated change? Although if our CI is passing then we don't really need this dependency, so, I'm cool with keeping the change

packages/react-juce/src/lib/Backend.ts
```diff
@@ -99,6 +100,15 @@ export class ViewInstance {
     // of a css prop, first convert it to kebab-case
     propKey = cssPropsMap[propKey] || propKey;
 
+    // convert provided color string to alpha-hex code for JUCE
+    const colors = new Colors();
+    if (colors.isColorProperty(propKey)) {
+      value = colors.colorStringToAlphaHex(value);
+      if (value.startsWith("linear-gradient")) {
+        value = colors.convertLinearGradientStringToNativeObject(value);
```
nick-thompson, line 108: One thought here: might we be interested in holding on to the `value` given by the user when we write to `this._props`? We only need this computed value when we speak to native, I could imagine it making sense to otherwise continue to deal in "non-computed" property values at the javascript level.

In which case, I would just say let's do something like `let nativeValue = colors.convertLinearGradientStringToNativeObject(value);` and then pass `nativeValue` below in the NativeMethods call, and pass `value` to the `Object.assign()` going onto `this._props`.

packages/react-juce/src/lib/MacroProperties/Colors.ts
```diff
@@ -0,0 +1,251 @@
+import ColorString from "color-string";
+import ColorNames from "color-name";
+
+export default class Colors {
```
nick-thompson, line 4: A bit of a nitpick, but I don't think we need this to be a class, and in fact when we use this class with `let colors = new Colors()` we're making an object and then throwing it away almost immediately because we don't need the notion of a persistent colors object here.

What I think I'd prefer is to keep a const list of color properties at module scope:
```js
import ColorString from 'color-string';
import ColorNames from 'color-name';

const COLOR_PROPERTIES = [
    'border-color',
    'background-color',
    ...etc
];
```

Then we can just export a set of pure functions:

```
function isColorProperty(propKey: string) {
    return COLOR_PROPERTIES.includes(propKey);
}

...etc...

export default {
  isColorProperty,
  colorStringToAlphaHex,
  ...
}
```

Then when using it:

```
import Colors from 'lib/MacroProperties/Colors';

// No need to construct an object here
if (Colors.isColorProperty(blah)) {
    ...etc
}
```

react_juce/core/View.cpp
```diff
@@ -204,10 +206,155 @@ namespace reactjuce
 
         if (props.contains(backgroundColorProp))
         {
-            juce::Colour c = juce::Colour::fromString(props[backgroundColorProp].toString());
-
-            if (!c.isTransparent())
-                g.fillAll(c);
+            auto colorProp = props[backgroundColorProp];
+            //If object we assume it's a linear gradient
+            if(colorProp.isObject()) {
```
nick-thompson, line 211: Wowwww yes. Love this.

Thinking we might want to reuse this in many places... maybe we pull it out into a standalone function in our Utils.h/cpp file? The function could return a `std::variant<juce::Colour, juce::ColourGradient>` depending on what it learned from the object passed in.

nick-thompson (commented):
@eupston this is massive, great work. I love it!

I left a few nit-picky things inline, let me know what you think. I think with those last few tiny changes I'm very excited to merge

packages/react-juce/src/lib/Backend.ts
eupston: roger that. can see reusing `this._props` being useful for say presenting the user with an error when passing in an invalid prop value. Sure there's lot of other useful reasons we might want to access props besides just during rendering too.

packages/react-juce/src/lib/MacroProperties/Colors.ts
eupston: yup yup good call

packages/react-juce/package.json
eupston: removed it locally and seems fine, so gonna take it out.

react_juce/core/View.cpp
eupston: sweeet yup did have in the back of my mind that this logic will probably need to be reusable for a few other things we paint so good call. 
 
Have moved color variant logic over to utils, let me know if it's along line of what you were thinking.

eupston:
Alrighty @nick-thompson think that should be all discussions resolved. 
let me know if you see anything else. cheers

react_juce/core/Utils.cpp
```diff
@@ -22,5 +22,154 @@ namespace reactjuce
...
+                    if(color->getProperty("type") == "additional"){
+                        juce::String hexString = color->getProperty("hex");
+                        juce::Colour additionalColor = juce::Colour::fromString(hexString);
+                        double originalPercent = color->getProperty("position");
+                        double gradientLineDistanceExtension = gradientLineDistance * (1 + (negativeExtensionPercent + positiveExtensionPercent));
+                        double offsetPercent = ((gradientLineDistance * originalPercent) + (gradientLineDistance * negativeExtensionPercent)) / gradientLineDistanceExtension;
+                        gradient.addColour(offsetPercent, additionalColor);
+                    }
+                }
+                colorVar = gradient;
```
olidacombe, line 166: Maybe @nick-thompson or @JoshMarler can confirm, I think in the to-be written style guide returning inside the if would be preferred to adding an else.

i.e.
```
// no declaration of colorVar
if(colorVariant.isObject()) {
  // ...
  juce::ColourGradient gradient = juce::ColourGradient (minColor, x1, y1, maxColor, x2, y2,false);
  // ...
  return gradient
}

return juce::Colour::fromString(colorVariant.toString());
```

Almost not worth mentioning, but the style guide needs a little love.  Also apologies if I've got it completely the wrong way around.

eupston: O yup sweet as think I see a few other instances of this style. I'll change that

eupston: O yup sweet as think I see a few other instances of this style. I'll change that

nick-thompson:
Finally getting back around to this, this is awesome @eupston thanks so much for the effort.

Going ahead with the merge! 🚀

nick-thompson (approved):

Issue #205, opened 2021-01-06
MouseEnter / MouseLeave


nick-thompson:
This task is for supporting `onMouseEnter` and `onMouseLeave` in `blueprint::View`, just as we have with `onMouseUp` and `onMouseDown`

https://developer.mozilla.org/en-US/docs/Web/API/Element/mouseenter_event
https://developer.mozilla.org/en-US/docs/Web/API/Element/mouseleave_event

stfufane:
Hey! I could not find how to assign myself to this issue but I gave it a try anyway :) Cheers

nick-thompson:
Excellent, thanks @stfufane! Will check it out shortly, and I've assigned you here

nick-thompson:
Closed with #206!

PR #206, opened 2021-01-09
#205 - Manage mousenter/mouseleave events

stfufane:
Hi there,

I know it was not the most urgent issue, but I wanted to give it a try to learn a bit more about the react side. 

I modified the GainPlugin example to have the button change its border color on hover, it works but I'll let you check if it's done correctly, I may have misunderstood the right way to do it.

Cheers!

### DIFF
--- a/packages/react-juce/src/components/Button.tsx
+++ b/packages/react-juce/src/components/Button.tsx
@@ -86,26 +86,14 @@ export class Button extends Component<ButtonProps, ButtonState> {
     }
   }
 
-  handleEnter = (e: SyntheticMouseEvent): void => {
-    if (typeof this.props.onMouseEnter === 'function')
-      this.props.onMouseEnter.call(null, e);
-  }
-
-  handleLeave = (e: SyntheticMouseEvent): void => {
-    if (typeof this.props.onMouseLeave === 'function')
-      this.props.onMouseLeave.call(null, e);
-  }
-
   render = () => {
-    const { onMouseDown, onMouseUp, onClick, onMouseEnter, onMouseLeave, ...other } = this.props;
+    const { onMouseDown, onMouseUp, onClick, ...other } = this.props;
     const opacity = this.state.down ? 0.8 : 1.0;
 
     return (
       <View
         onMouseDown={this.handleDown}
         onMouseUp={this.handleUp}
-        onMouseEnter={this.handleEnter}
-        onMouseLeave={this.handleLeave}
         opacity={opacity}
         viewRef={this._ref}
         {...other}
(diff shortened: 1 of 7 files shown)


JoshMarler:
@stfufane, nailed it. This looks great. 

I'll pull down when I'm back at my desk to test but no issues from me. Anything to add @nick-thompson or shall I go ahead and merge ? Will do the package bumping necessaries.

packages/react-juce/src/components/Button.tsx
```diff
@@ -84,14 +86,26 @@ export class Button extends Component<ButtonProps, ButtonState> {
     }
   }
 
+  handleEnter = (e: SyntheticMouseEvent): void => {
```
nick-thompson, line 89: For cases like this one where we only wrap the supplied `props.onMouseEnter` in a function which checks that the thing is callable, we can just omit this code.

MouseDown, MouseUp, and onClick are special cases where we have to do some juggling for the end user before we call their provided callbacks. MouseEnter and MouseLeave are straightforward enough that we can just let them through to the underlying `View` if they're present.

So that means that where we write:
```
const { onMouseDown, onMouseUp, onClick, ...other } = this.props;
```
onMouseLeave and OnMouseEnter will get wrapped up into `...other` if they're present, and then get unwrapped via `{...other}` onto the View on line 111. If these props are not available, they won't get wrapped/unwrapped and passed on, which is just as we want.

So basically you can omit all the changes to this file except for 2 line change to the typescript interface above

nick-thompson (approved):
Agreed this is essentially perfect. Great work @stfufane thank you.

I left one comment in line where we can remove some code. After that, I think we're quite ready to merge!

nick-thompson:
And yea @JoshMarler, after the quick edit to remove that bit of code above, if you get around to the merge before I do, yes please + thank you!

packages/react-juce/src/components/Button.tsx
JoshMarler: Ahh yeah good shout. Skimmed too quickly.

stfufane: Oh okay I see, thanks for the review :) I noticed the more complex behaviour with the other events but didn't realize it wasn't needed to explicitly declare these ones, but the explanation is clear! 

I'll clean this and commit the changes!

stfufane: It's done! I'll let you check by principle but I tested and it still works without the useless code :)

nick-thompson:
Excellent, thanks @stfufane! I'll go ahead and rebase + merge this, then bump the npm package.

nick-thompson:
Merged in a9f1f53cafc1ba4c982ef53d78416acbd99b8f30 and e12e99ec485c4593afdfac86f3a6f1494886a0b8 (commit hashes changed due to the rebase). I'll go ahead and close this accordingly

Issue #178, opened 2020-12-15
V8 EcmascriptEngine Pimpl


nick-thompson:
As of #145 we now have a cleaner EcmascriptEngine interface which is satisfied via a Duktape implementation Pimpl by default. The task here is to provide another EcmascriptEngine implementation via V8 (https://v8.dev/docs) with a new Pimpl.

Quick details:
* Let's move the existing `EcmascriptEngine.cpp` to `EcmascriptEngine_Duktape.cpp`
* Let's add a new `EcmascriptEngine_V8.cpp`
* Let's add a preprocessor definition to allow the user to toggle which engine they want to build with
    * e.g. `-DBLUEPRINT_USE_V8=1`
* Let's _not_ include the v8 source with the project, as we do with Duktape.
    * With Duktape, it's cool because the source is really quite small and provides an easy default
    * With v8 things are slightly more complicated, so we'll ask that users bring their own V8 directory somewhere else and configure their project with the correct header includes `-Ipath/to/your/v8`

nick-thompson:
Adding a bit more detail here as I think v8 integration will become a big priority after beta.

Generally I think the details of integration here fall into two bits. The first is just getting things to compile, and the second is fleshing out the full `EcmascriptEngine` interface.

To the first point, I envision something like

```cpp
// EcmascriptEngine_V8.cpp
#include <libplatform/libplatform.h>
#include <v8.h>

struct EcmascriptEngine::Pimpl
{
    Pimpl()
    {
        // This kinda stuff, probably need to hold some of it as member vars
        std::unique_ptr<v8::Platform> platform = v8::platform::NewDefaultPlatform();
        v8::V8::InitializePlatform(platform.get());
        v8::V8::Initialize();
    }
};
```
Then, for the user to compile successfully, they'll need to `-DREACTJUCE_USE_V8=1` which we'll use inside of `blueprint.cpp` to decide whether we `#include EcmascriptEngine_Duktape.cpp` or `#include EcmascriptEngine_V8.cpp`, they'll need to add an include path for their v8 headers: `-Ipath/to/v8/include`, and link against V8 `-lv8_monolith`. It's my intention for the v8 source directory to be something that blueprint pays no mind to, the user can put it wherever and do whatever they want with it, just make sure they specify the correct include/link flags in their CMake config or whatever.

[...]

PR #208, opened 2021-01-10
First pass v8 integration

nick-thompson:
A'ight @JoshMarler this one's for you!

This work is _not_ a branch I intend to merge, at least not in its current state. This is more of a proof-of-concept branch for kicking off the v8 integration behind the EcmascriptEngine interface (#178).

The work here addresses only this:
* Linking the GainPlugin against v8 somewhere external to the project. You can see that I have v8 in `~/Dev/v8/v8`, and have built a macos x64 v8_monolith myself for linking.
* Adding a `REACTJUCE_USE_V8` compiler definition for toggling whether to load the Duktape implementation or the V8 implementation, and enabling that def in GainPlugin
* Filling in junk to satisfy the EcmascriptEngine interface in most cases. This essentially has the GainPlugin totally broken, but my goal was to (1) get something compiling and (2) get the first `evaluate` pass to run successfully.

This is almost there. There is a little more work to satisfy some global shims. The first, which I'll add here shortly, is to shim out more of the __BlueprintBackend__ interface, for which I have a clean approach in mind. The second, is adding polyfills for timers to satisfy the updated React-Reconciler dependency.

Then, of course, there's work to actually persist the v8 isolate and context on the `EcmascriptEngine::Pimpl` instance (right now I just make the and throw them away inside of evaluate), and start fleshing out the remainder of the interface. My hunch right now is that this will generally be an easier integration than Duktape, which is quite nice!

### DIFF
--- a/examples/GainPlugin/CMakeLists.txt
+++ b/examples/GainPlugin/CMakeLists.txt
@@ -1,6 +1,9 @@
 cmake_minimum_required(VERSION 3.15)
 project(GainPlugin VERSION 0.1.0)
 
+# Apparently this has to come before any `add_executable` calls
+link_directories(~/Dev/v8/v8/out.gn/x64.release.sample/obj/)
+
 # `juce_add_plugin` adds a static library target with the name passed as the first argument.
 # This target is a normal CMake target, but has a lot of extra properties set
 # up by default. As well as this shared code static library, this function adds targets for each of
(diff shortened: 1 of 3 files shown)


nick-thompson:
@JoshMarler I just snuck a little commit onto master: c4274df910bd8195175b305b61dd3b699ee14643. This nudges the way that our `react-juce` package relies on the global `__BlueprintNative__` object, and shims a little Proxy in there when we know that `process.env.NODE_ENV !== 'production'`. This means that if you do the following:

```
cd packages/react-juce
npm run watch
cd ../../examples/GainPlugin/jsui
npm install ../../packages/react-juce
npm start
```

You'll be running GainPlugin's jsui built against react-juce _development_, where `process.env.NODE_ENV === 'development'`, so the proxy will be shimmed in there. This trick makes the bundle quite portable, so you could drop it in Chrome and run it there or drop it into v8 without injecting all the required native methods.

With that, all that's really left here to get the initial bundle really running is shimming the timers (can just shove some bs code in there for now to get it to run)

nick-thompson:
@JoshMarler with all of the work going into Hermes and Quickjs lately, I'm going to close this PR out. We can revisit v8 at a later time

JoshMarler:
Yeah sounds fair enough. Hopefully will have a clean approach to follow now if we come back to it.

PR #211, opened 2021-01-13
Support of image web URL

stfufane:
I made the changes to have the simple version without DownloadTask, it's much better ;)

Just one thing, I didn't manage to compile when trying to call native methods on props with `props[onloadprop]()` and could not find examples of how to do so, I guess I'll let you manage this part, or you can show me later for the sake of learning, it's up to you, I'll look at the code anyway if you do it :) 

It works like this anyway so it can be merged in this version if you're okay and maybe the onLoad/onError callbacks can be added in an other PR.

Cheers!


stfufane:
Actually I can maybe do an extra commit to replace the cute kittens that now replace the logo when loading an image URL, sorry about that 😄

react_juce/core/ImageView.cpp
```diff
@@ -0,0 +1,228 @@
...
+        {
+            downloadImageAsync(props[sourceProp].toString());
+        }
+    }
+
+    //==============================================================================
+    void ImageView::downloadImageAsync(const juce::String& source)
+    {
+        // Allow only one download at the same time for a given ImageView.
+        if (downloading)
```
nick-thompson, line 108: This seems like probably a premature optimization, and so I think we should probably not do this.  Basically a user might change their image source property quite quickly, and we should not cancel the most recent request if the older one is still in-flight. We'll end up with stale state that way.

Let's go ahead without this idea, and if we find that its' really a performance problem we'll come back and revisit it then.

react_juce/core/ImageView.cpp
```diff
@@ -0,0 +1,228 @@
...
+        {
+            appRoot->getThreadPool().addJob([this, source] ()
+            {
+                downloading = true;
+                juce::MemoryBlock mb;
+
+                if (!juce::URL(source).readEntireBinaryStream(mb)) {
+                    downloading = false;
+                    shouldDownloadImage = true;
+                    throw std::runtime_error("Failed to fetch the image!");
```
nick-thompson, line 121: Instead of throwing here, let's enqueue a different call to `juce::MessageManager::callAsync` that basically just invokes `props["onError"]` if the user has supplied such a prop. Else maybe we just log to the console

react_juce/core/ImageView.cpp
```diff
@@ -0,0 +1,228 @@
...
+                }
+
+                auto image = juce::ImageFileFormat::loadFrom(mb.getData(), mb.getSize());
+                if (!image.isNull()) {
+                    juce::MessageManager::callAsync([this, image]()
+                    {
+                        auto drawableImg = std::make_unique<juce::DrawableImage>();
```
nick-thompson, line 128: Yup! :+1:

nick-thompson (commented):
Awesome, thanks @stfufane! I left a couple quick comments inline, but generally this is very close to ready to merge

react_juce/core/ImageView.cpp
stfufane: Yeah blocking is probably not the right way. But what you're saying made me think of something else. 
If the source is changing quickly, we can't be sure that the last request will finish after previous ones, so there could still be issues. 
And on another level, I think some caching would be interesting to have to avoid new web requests if the same URL is required several times and the source is toggled a lot. It would require some specification but I think it could be useful.

react_juce/core/ImageView.cpp
stfufane: Good one! So just to be sure, it works like that?
```cpp
if (props.contains(onErrorProp) && props[onErrorProp].isMethod())
    std::invoke( props[onErrorProp].getNativeFunction() );
```

nick-thompson: Yup, here's an example: https://github.com/nick-thompson/blueprint/blob/master/react_juce/core/CanvasView.h#L718

react_juce/core/ImageView.cpp
nick-thompson: > If the source is changing quickly, we can't be sure that the last request will finish after previous ones, so there could still be issues.

Ah yea great point! I didn't think of that

How about this: https://docs.juce.com/master/classImageCache.html :) And this: https://docs.juce.com/master/structDefaultHashFunctions.html#ab3ad9697d3bbe3a07b64f49eddb259f7

So what we can do here is this... during `setProperty`:
* We construct a hash of the source property string using the method I linked above
* If the existing hash already exists in the ImageCache, we can just set the drawable, call repaint(), and return early
* If not, we run the existing code flow.
* Then when we finish downloading an image from the web, we call `juce::MessageManager::callAsync` but in there we do things slightly differently:
    * Insert the image that we just downloaded into the cache
    * Set the `drawable` by reading an Image from the cache based on the hash of the current source prop
    * Repaint

This way, even if requests come in out of order, we will always make sure that the drawable we draw is the correct one for the current source prop. That way too, we can skip subsequent fetches for the same resource and return early.

What do you think?

stfufane: Sounds great to me! I'll push the changes when I'm done :)

stfufane:
Here are the last changes, it starts to look like something I think :) 

I refactored a few lines in ImageView.cpp and added the callbacks, I finally understood how it works, hurray! 

I tested both callbacks and it works well, I'll let you check if it looks good to you, cheers!

nick-thompson:
@stfufane just looking again now, this is great, thank you! Two small things I want to do:

* I think we can probably start using that ImageCache for all image sources, not just web ones (no need to go to disk twice if we already went once for that image!)
* Now that we have the onError callback we should change those `throw` lines to instead invoke the error. Or maybe try/catch at the top of setProperty and when catching call onError.

These are little things though; not terribly important and can easily be part of a follow-up PR, so I'm going to go ahead and merge :)

Issue #133, opened 2020-11-17
Optimise Canvas implementation


JoshMarler:
We've yet to spend anytime optimising our `CanvasView` implementation is this is almost certainly going to be a performance hot spot, particularly for dynamic components like sliders and meters. 

A likely hotspot here is going to be GC churn with all the `CanvasContext` objects passed between Native/JS. Are there some wins to be had here or ways to reduce GC churn? 

@nick-thompson, if you have any thoughts on this maybe we can add to the comments here and then tackle.

nick-thompson:
Ok so here's an idea I came up with the other day that completely eliminates the canvas context native binding and I think should help with the GC churn. Below I'll list some scratch code, but the idea is essentially to shim a `CanvasRenderingContext` in such a way that we just build up a list of "draw commands," where each command is basically a tuple containing the command name and the arguments for that command. When it's time to draw, we reset our internal `drawCommands` array, call the user-supplied `onDraw`, build up a list of commands, and at the end we return that list over to native. The native side then runs a big `for` loop unpacking each tuple and affecting the appropriate write to the `Graphics` context.

```
class CanvasRenderingContext {
  constructor() {
    this.drawCommands = [];
  }

  reset() {
    this.drawCommands.length = 0;
  }

  set fillStyle(value) {
    this.drawCommands.push(["setFillStyle", value]);
  }

  set lineWidth(value) {
    this.drawCommands.push(["setLineWidth", value]);
  }

  fillRect(x, y, width, height) {
    this.drawCommands.push(["fillRect", x, y, width, height]);
  }

  ...etcetera
}

export class Canvas extends Component<CanvasProps, CanvasState> {
  constructor(props: CanvasProps) {
    super(props);

    this._context = new CanvasRenderingContext();
    this._onMeasure = this._onMeasure.bind(this);
    this._onDraw = this._onDraw.bind(this);

    this.state = {
      width: 0,
      height: 0
    };
  }

    _onDraw() {
    if (typeof this.props.onDraw === 'function') {
      this._context.reset();
      this.props.onDraw(this._context);
      return this._context.drawCommands;
    }
    return [];
  }

  render()...etc
}
```

And over on native we do (https://github.com/nick-thompson/blueprint/blob/master/blueprint/core/blueprint_CanvasView.h#L722):

```
auto commands = std::invoke(props["onDraw"].getNativeFunction(), nfArgs);

for (auto& cmd : commands->getArray()) {
    auto name = cmd[0].toString();
[...]

PR #212, opened 2021-01-18
Canvas optimisation

JoshMarler:
@nick-thompson, 

Tackles https://github.com/nick-thompson/blueprint/issues/133

Opening up a PR just so you can see the work in progress. This all needs a bit of tidying. Main issue is that we would need to return JS function return values from `EcmascriptEngine::readVarFromDukStack` if we decide to keep this approach. 

I think the overall approach should be quite neat though. Will move all the draw commands into cpp and separate `Canvas.h` into cpp/h pair etc.


react_juce/core/EcmascriptEngine.cpp
```diff
@@ -744,10 +744,13 @@ namespace blueprint
                                 }
 
                                 // Clean the result and the stash off the top of the stack
+                                //TODO: Might not be exactly the right clean up code but GainPlugin
+                                //      working with this change for canvas drawCommands return.
+                                //      Double check and fixup.
+                                juce::var result = readVarFromDukStack(sharedContext, -1);
```
nick-thompson, line 750: I think this is good :+1:

react_juce/core/CanvasView.h
```diff
@@ -57,613 +57,495 @@ namespace blueprint
...
-                        juce::ignoreUnused(args);
+            g.setFillType(ctx.properties.fillStyle);
+            g.fillRoundedRectangle(x, y, width, height, cornerSize);
+        }
 
-                        const float x = args.arguments[0];
-                        const float y = args.arguments[1];
+        void clearRect(CanvasContextState &ctx, juce::Graphics &g, const juce::var::NativeFunctionArgs &args)
+        {
+            //TODO: How are we implementing this without an image?
```
nick-thompson, line 281: We could still do this via image if needed, right? instead of passing the actual `juce::Graphics` instance that we get in `paint` we could make a `juce::Image`, construct a graphics instance pointing to that image, run all the context/draw operations, then paint the image to the real graphics context?

I don't know how necessary that is, but seems like an option

nick-thompson (commented):
Excellent, this is just what I had in mind.

One thought inline about using an Image if necessary to support some of the underlying implementation details, but generally I think this is great.

I'd love to attach some before/after profile numbers here if you happen to have them?

Also, what are you thinking? What's left to move this away from WIP status?

react_juce/core/CanvasView.h
JoshMarler: Yeah we could do that indeed. And from memory I think there may be actually functional reasons for doing this related to the old `autoClear` prop. Removing it does nicely solve the scalling issue on retina :-). 

I'm quite happy to render to an Image still and then fix the scalling stuff separately. There was a state based reason for rendering to an `Image` from what I remember? Something about scrolling waveforms etc.

nick-thompson: Ah yea, for a scrolling waveform you could, for example, snapshot the current state, paint it back to the canvas but shifted 1px to the left, and then explicitly paint only the rightmost 1px region of the canvas.

So... yea... might want the image there :D. The retina scaling issue is annoying though... I wonder if we could solve it by making that `juce::Image` 2x as big as the actual component and render all these draw commands onto that but automatically scale the positions by 2? Then when painting the image to the screen we just scale it then? That would address the retina @2x part right?

JoshMarler: Yeah totally. I did attempt something along those lines a while back and got most of it working I think. OK cool. I'll continue rendering to an Image and reinstate that `autoclear` behaviour. I think we should setup a default prop for that set to `true` maybe?

JoshMarler:
> Also, what are you thinking? What's left to move this away from WIP status?

Yep, will totally get some A/B profile data on here before merge. 

To move from WIP I'm thinking probably moving all those `DrawCommand` functions to `CanvasView.cpp`. Sort the `juce::Image` render and general cleanup. Then I think this can go in. Following that I can tackle the remaining Canvas API functions from our beta list.

nick-thompson:
> Yep, will totally get some A/B profile data on here before merge.
> 
> To move from WIP I'm thinking probably moving all those `DrawCommand` functions to `CanvasView.cpp`. Sort the `juce::Image` render and general cleanup. Then I think this can go in. Following that I can tackle the remaining Canvas API functions from our beta list.

Perfect! :+1:

react_juce/core/CanvasView.h
nick-thompson: :+1: sounds good. I'm tempted to default `autoclear` to `false`, it feels like a "favor explicit instead of implicit" type of argument. That is, if we autoclear by default, then we have this implicit behavior that occurs in between two subsequent calls to your onDraw callback. If we don't autoclear by default, then nothing happens between two subsequent calls, there's no surprise there, and if you want clearing behavior you can explicitly request it with a 1 prop change

JoshMarler: OK cool sounds reasonable to me. Will fix this up today and finish things off.

JoshMarler: Actually thinking about this it's really useful to have a C++ `CanvasContext` equivalent exposed in the `CanvasView` header. 

With the old implementation I used this to allow me to push look and feel paint overrides into JS on some C++ components. i.e. a hybrid between a C++ and React component. This is useful for very specific components which blueprint doesn't provide equivalent functionality for right now but which we still want to "skin" from React. 

How would you feel about keeping a public function available similar to the following:

`DrawCanvasCommands(CanvasContext &ctx, const juce::var &drawCommands)`

In the above `CanvasContext` will probably just be a struct or similar to the original implementation (i.e. holds image and graphics context) and `drawCommands` is the JS command array passed into native.

JoshMarler: Thinking about it. Might make more sense from API point of view to have a 

`CanvasContext::processDrawCommands(const juce::var &drawCommands)` ?

nick-thompson: Hmm interesting, so you're thinking something like the following?

```cpp
class CanvasView : public blueprint::View
{
        struct CanvasContext
        {
            int                                width  = 1;
            int                                height = 1;
            juce::Path                         path;
            CanvasContextProperties            properties;
            std::vector<juce::AffineTransform> transformStack;
            juce::Image internalImage;
            
            void processDrawCommands(const juce::var &drawCommands);
        };

        // Regular canvasView stuff here etc etc

private:
       CanvasContext ctx;
};
```

And with this, CanvasView's implementation largely composes over the behavior of the internal `CanvasContext` struct, which is made public to accommodate situations like the one you have?

I can get into this idea, seems like a clean way API for the CanvasView itself anyways

JoshMarler: Yep exactly. Something along these lines. I think we'd still have an anonymous std::map of command -> function but it would look more like:

```
void setFillStyle(CanvasContextState &ctx, const juce::var::NativeFunctionArgs &args)
{
    //DO necessary work with ctx. Possibly just ctx.setFillStyle()?
}

std::unordered_map<juce::String, DrawCommand> DrawCommands
{
    {"setFillStyle", setFillStyle}
}

void CanvasContext::processDrawCommands(const juce::var &drawCommands)
{
    jassert(drawCommands.isArray());                                                                                                                                                                 
                                                                                                                                                                                                                   
    for (juce::var &command : *drawCommands.getArray())                                                                                                                                              
    {                                                                                                                                                                                                
        jassert(command.isArray());                                                                                                                                                                  
        jassert(command.getArray()->size() >= 1);                                                                                                                                                    
                                                                                                                                                                                                                   
        const auto args = command.getArray();                                                                                                                                                        
        const auto name = (*args)[0];                                                                                                                                                                
                                                                                                                                                                                                                   
        if (auto it = DrawCommands.find(name); it != DrawCommands.end())                                                                                                                             
        {                          
            // CanvasContext passes self to the existing static functions in our DrawCommands map
            // Could take a different approach but this is probably lighter than each CanvasContext object
            // holding a map of strings to functions                                                                                                                                                                  
            it->second(this, g, juce::var::NativeFunctionArgs(juce::var(), args->getRawDataPointer() + 1, args->size() - 1));                                                                         
        }                                                                                                                                                                                            
    } 
}
```

nick-thompson: Yup, sounds good. Last thought:

```
void setFillStyle(CanvasContextState &ctx, const juce::var::NativeFunctionArgs &args)
{
    //DO necessary work with ctx. Possibly just ctx.setFillStyle()?
}
```

I think I'd prefer to have these static functions just be kind of pure functions that operate over the CanvasContext and make the necessary mutations. It seems like it'd end up being a lot of duplication to write a static `setFillStyle` that just unwraps the args and then calls _back_ to `ctx.setFillStyle`

JoshMarler: Yeah cool. That was a choice to make. In which case these functions will probably need to just directly access the `CanvasContext`'s `juce::Graphics` instance? That graphics instance will be setup to render to the `juce::Image` in `CanvasContext`'s constructor I guess. 

Cool with that general approach?

nick-thompson: Ah yea, yup, that sounds ok to me!

JoshMarler: Sweeeeet. Let's finish this bad boy.

react_juce/core/ShadowView.cpp
```diff
@@ -127,7 +127,7 @@ namespace blueprint
         {juce::String("padding-") + YGEdgeToString(YGEdgeHorizontal), getYogaNodeDimensionSetter(BP_SPREAD_SETTER_PERCENT(YGNodeStyleSetPadding), YGEdgeHorizontal)},
         {juce::String("padding-") + YGEdgeToString(YGEdgeVertical), getYogaNodeDimensionSetter(BP_SPREAD_SETTER_PERCENT(YGNodeStyleSetPadding), YGEdgeLeft)},
         {juce::String("margin-") + YGEdgeToString(YGEdgeAll), getYogaNodeDimensionAutoSetter(BP_SPREAD_SETTER_AUTO(YGNodeStyleSetMargin), YGEdgeAll)},
-        
+
```
JoshMarler, line 130: This shouldn't be here. Will remove

packages/react-juce/src/components/Slider.tsx
```diff
@@ -1,9 +1,9 @@
-import React, {Component, PropsWithChildren} from 'react';
+import React, { Component, PropsWithChildren } from 'react';
 import invariant from 'invariant';
 
-import {Canvas} from './Canvas';
-import {View} from './View';
-import {SyntheticMouseEvent} from '../lib/SyntheticEvents';
+import { Canvas, CanvasRenderingContext } from './Canvas';
```
JoshMarler, line 4: Will remove these indent changes pre merge.

nick-thompson:
This looks excellent @JoshMarler, really like the shape it's taking. Let me know when you're ready to mark this non-WIP and I'll take another review pass!

JoshMarler:
@nick-thompson , Should be good for final pass. Functionality is as before. I've not added any additional API calls or fixed the Retina issue in this PR. Once this goes in I'll follow up with those two as separate tasks. 

When you're happy with this I'll do the usual npm bump dance. 

Will profile on Mac now and get A/B comparison.

JoshMarler:
So, 

Some performance data. 

. With this change watching the memory usage monitor in Xcode shows memory sat at a constant (around 56 MB) whilst 
  standalone plugin idle. Prior to this change memory usage steadily climbs from 56Mb whilst idle in standalone plugin. CPU usage similar with both approaches.  
  
. With this change moving slider in standalone plugin shows memory usage fluctuate between 56 -> 80 MB. So still a decent amount of heap churn. Prios to this change 

Instruments actually shows big allocation hotspots (in terms of frequency) around JUCE `drawImageAt` inside `CanvasView::paint`. It's possible we could investigate whether there is anything we can do to improve that. I expect not. 

Heaviest stack trace out of `juce::CoreGraphicsContext::drawImage`. I do wonder if using the `paint` functions `juce::Graphics` object would provide a significant performance boost but we've suggested we want to keep this image rendering behaviour for things like scrolling waveforms.

JoshMarler:
@nick-thompson , I think I actually saw the biggest performance gain in my earlier implementation which didn't render to a juce::Image. 

If I remove that image rendering I say 20% CPU reduction in xcode and much less time spent in paint call under Instruments.

JoshMarler:
@nick-thompson , OK, following our Discord chat and discussions here this should be ready to review now with the `stateful` prop added as opt-in behaviour to render to a persistent `juce::Image` instance.

packages/react-juce/src/components/Canvas.ts
```diff
@@ -1,66 +1,164 @@
...
+  }
+
+  resetTransform(): void {
+    this._drawCommands.push(['resetTransform']);
+  }
+
+  //================================================================================
+  // Image functions
+  //
+  //TODO: Add support for other drawImage overloads.
```
nick-thompson, line 132: I think maybe we can match the `source` prop API on `ImageView` here to support various image types

JoshMarler: Yeah awesome. This was actually just a comment pulled over from the original code. Could we take that as separate issue/PR ? I've deliberately avoided implementing any of the existing TODO's or additional API routines in this PR. Hoping to follow up with those next.

nick-thompson: Yep of course

packages/react-juce/src/components/Canvas.ts
```diff
@@ -1,66 +1,164 @@
...
+  //      Currently only support SVG string. What is correct
+  //      type to use here?
+  drawImage(image: string, dx: number, dy: number): void {
+      this._drawCommands.push(['drawImage', image, dx, dy]);
+  }
+
+  //================================================================================
+  // Text functions
+  strokeText(text: string, x: number, y: number, maxWidth?: number): void {
+    if (maxWidth === undefined)
```
nick-thompson, line 142: Might be better here to go with
```
if (typeof maxWidth === 'number')
  // pass it along
else
  // dont pass it
```

Seems more appropriately explicit as a condition for when we want to pass it along.

I don't want to hold up the merge on account of this though, so if you're down for this change let's come back for it in the follow up PR

react_juce/core/CanvasView.h
```diff
@@ -9,739 +9,74 @@
 
 #pragma once
 
-#include "View.h"
+#include"View.h"
```
nick-thompson, line 12: Weird space? Same thing, let's not block the merge for it. Clang-format will save us!

nick-thompson (approved):
This looks perfect. I'm going to go ahead and merge, thanks @JoshMarler!

Issue #34, opened 2019-12-18
Rename project?


nick-thompson:
Someone sent me a link to this recently: https://blueprintjs.com/

Turns out "React" + "Blueprint" is already a thing... might be time to come up with a new name for this project!

teotigraphix:
`Blueprint++`  ;-)

nick-thompson:
Thinking about this again the past few days. When I think about the React ecosystem, there's a pretty consistent naming pattern for rendering backends: `react-dom`, `react-native`, `react-art`, `react-native-desktop`, `react-native-windows`, `proton-native`, etc.

In truth, Blueprint is just another rendering backend, it just happens to target native C++/JUCE to provide a react-based approach to writing audio apps/plugins. So perhaps we should follow the pattern? 

Going down that road, I think the primary features of this particular rendering backend are:
* Cross-platform
* Native C++
* Audio-focused

The obvious name is `react-juce`, but the way blueprint is written there's nothing _really_ tying us to JUCE itself. For example, we could very likely support targeting iPlug2 or JUCE. For that reason, I hesitate on `react-juce` and instead think we should lean towards the primary feature list up there.

A couple other name ideas:
* React XPN, `react-xpn` – React for Cross(X)-Platform Native
* React XPA, `react-xpa` – React for Cross(X)-Platform Audio
* React C++, `react-cpp` – React for C++
    * I don't love this one because it's not exactly generic C++ that we're targeting...

I'm curious to hear anyone's input!

JoshMarler:
Heh. Yeah, tricky one. 

Is there a real desire to support targets other than JUCE? I really like `react-juce` and that would have been my (pretty obvious) suggestion also. 

Depends on where you want to draw the line with this framework I guess. Will Blueprint (or whatever it becomes) ever want to go beyond JUCE? If the purpose of the framework is something along the lines of "React based UI's for Cross Platform Audio Applications" is there a compelling reason for it not to exclusively target JUCE ? 

I understand the reasons one might wish to target other cross-platform audio frameworks but do we reeeeaaaaly want to?

Could be a brave new world of:

```
  createViewInstance(viewType, props, parentInstance) {                                                                                                                                                            
    const id = __ReactJuceNative__.createViewInstance(viewType);                                                                                                                                                   
    const instance = new ViewInstance(id, viewType, props);                                                                                                                                                        
                                                                                                                                                                                                                   
    __viewRegistry[id] = instance;                                                                                                                                                                                 
[...]

PR #216, opened 2021-01-23
(#34) Rename top-level namespace to reactjuce::

nick-thompson:
w00t!

After this, most of what's left is updating references to "Blueprint" in the documentation and internal names for coordinating js<>C++ access (e.g. BlueprintBackend)


nick-thompson:
Just tacked on a second commit to cover what I _think_ is the rest of the real renaming effort. Probably best to look at this commit by commit now.

After this, all that's left is really:
* Updating the documentation to replace instances of "Blueprint"
* Addressing the TestRunner code

Before merging this I'll bump react-juce to 0.2.10 on npm and update the package jsons

JoshMarler:
@nick-thompson, Nice! Looking solid. Pulled down on linux machine and all looking good. 

Ran a grep under react-juce/core directory and still a few "blueprint" references. Are we leaving for the moment? Things like: 

`#ifdef BLUEPRINT_H_INCLUDED`

`funcId(juce::String("blueprintCallback")`

Also index.js in GainPlugin example. 

I'm not sure if these are outside the planned scope of this PR?

nick-thompson:
Ok, covered another few and bumped 0.2.10 on npm.

`#ifdef BLUEPRINT_H_INCLUDED` I will come back for in a later PR while also addressing the test runner.

Issue #21, opened 2019-08-22
TextInput native view


nick-thompson:
See https://docs.juce.com/master/classTextEditor.html and https://facebook.github.io/react-native/docs/textinput.html

JoshMarler:
@nick-thompson, quick question on this one. Do you envisage this basically being a wrapper around `juce::TextEditor` our would you imagine some composition of `View` and `Text` which handles keypress events and triggers an `onChange` prop/event?

nick-thompson:
Yea, good question. I think we can go ahead with `juce::TextEditor`, and in general I want to mimic the behavior of a web's `<input type="text">` with things like support for a `placeholder` prop and a `minlength`, `maxlength`, etc: https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input/text.

We'll want both an `onChange` and an `onInput` event, and we'll want to follow React's controlled components model: https://reactjs.org/docs/uncontrolled-components.html. For example, if the user supplies a `value` prop, then we never render anything into that text editor other than the string supplied by that value prop.

We'll support `readonly` which is just an alias for `editable` basically, and yea `multiline` makes sense although I think I'd prefer to support it by allowing it on the native interface, but _not_ allowing it at the JS level: if a user wants a multiline texteditor they should use `<TextArea>`, which as far as implementation considered is equivalent to `<TextInput multiline>` but making this distinction at the user-facing API level will make it easier to support web.

Lastly, `font`, yea, I think the `<TextInput>` should more or less support the same text-style props as the `<Text>` element itself, if we can.

nick-thompson:
Done in #217 by @utokusa!

PR #217, opened 2021-01-24
Add TextInput component

utokusa:
This PR is related to #21 .

## Summary
This PR allows you to use `TextInput` component which has similar behavior of `<input type="text">`.

screenshot

## Features
- React's controlled components model (when `value` prop is provided)
- `OnInput` and `OnChange` callback
- Some basic properties. `fontSize`, `placeholder`, `maxlength`, etc.
- `juce::TextEditor`'s properties. e.g. 'highlight-color' which can be set by 
  `juce::Component::setColour()` in original JUCE

### DIFF
--- a/packages/react-juce/src/components/TextInput.tsx
+++ b/packages/react-juce/src/components/TextInput.tsx
@@ -1,63 +1,59 @@
-import React, { PropsWithChildren } from 'react';
+import React, { PropsWithChildren } from "react";
 
 export interface InputEvent {
-  value: string
+  value: string;
 }
 
 export interface ChangeEvent {
-  value: string
+  value: string;
 }
 
 export interface TextInputProps {
-  value?: string,
+  value?: string;
 
-  color?: string,
-  fontSize?: number,
-  fontStyle?: number,
-  fontFamily?: string,
-  justification?: number,
-  'kerning-factor'?: number,
+  color?: string;
+  fontSize?: number;
+  fontStyle?: number;
+  fontFamily?: string;
+  justification?: number;
+  "kerning-factor"?: number;
 
-  placeholder?: string,
-  'placeholder-color'?: string,
-  maxlength?: number,
-  readonly?: boolean,
-  'outline-color'?: string,
-  'focused-outline-color'?: string,
-  'highlighted-text-color'?: string,
-  'highlight-color'?: string,
-  'caret-color'?: string,
+  placeholder?: string;
+  "placeholder-color"?: string;
+  maxlength?: number;
+  readonly?: boolean;
+  "outline-color"?: string;
+  "focused-outline-color"?: string;
+  "highlighted-text-color"?: string;
+  "highlight-color"?: string;
+  "caret-color"?: string;
 
-  onChange?: (e: ChangeEvent) => void,
-  onInput?: (e: InputEvent) => void,
+  onChange?: (e: ChangeEvent) => void;
+  onInput?: (e: InputEvent) => void;
 }
 
 /**
-* TextInput component imitates web's `<input type="text">`.
-* https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input/text
-* So it supports props like `placeholder`, `maxlength`, `onInput`, etc...
-* 
-* TextInput also supports React's controlled components model:
-* https://reactjs.org/docs/uncontrolled-components.html.
-* If the user supplies a `value` prop,
-* then it never renders anything into that text editor other than
-* the string supplied by that `value` prop.
-*/
+ * TextInput component imitates web's `<input type="text">`.
+ * https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input/text
+ * So it supports props like `placeholder`, `maxlength`, `onInput`, etc...
+ *
+ * TextInput also supports React's controlled components model:
+ * https://reactjs.org/docs/uncontrolled-components.html.
+ * If the user supplies a `value` prop,
+ * then it never renders anything into that text editor other than
+ * the string supplied by that `value` prop.
+ */
 export function TextInput(props: PropsWithChildren<TextInputProps | any>) {
-  return React.createElement('TextInput', props, props.children);
+  return React.createElement("TextInput", props, props.children);
 }
 
 /*
 // ================================================
 // Example code
 
-import React, { Component } from 'react';
+import React, { Component } from "react";
 
-import {
-  Text,
-  TextInput,
-  View,
-} from 'react-juce';
+import { Text, TextInput, View } from "react-juce";
 
 class App extends Component {
   constructor(props) {
@@ -65,24 +61,24 @@ class App extends Component {
     this._onInput = this._onInput.bind(this);
 
     this.state = {
-      textValue: '',
-    }
+      textValue: "",
+    };
   }
 
   _onInput(event) {
     console.log(`onInput: ${event.value}`);
-    this.setState({textValue: event.value});
+    this.setState({ textValue: event.value });
   }
 
   render() {
     return (
       <View>
         <TextInput
-            value={this.state.textValue}
-            placeholder="init message"
-            onInput={this._onInput}
-            {...styles.text_input}
-          />
+          value={this.state.textValue}
+          placeholder="init message"
+          onInput={this._onInput}
+          {...styles.text_input}
+        />
       </View>
     );
   }
@@ -90,15 +86,15 @@ class App extends Component {
 
 const styles = {
   text_input: {
-    backgroundColor: 'ff303030',
-    color: 'ff66FDCF',
+    backgroundColor: "ff303030",
+    color: "ff66FDCF",
     fontSize: 15.0,
-    fontFamily: 'Menlo',
+    fontFamily: "Menlo",
     fontStyle: Text.FontStyleFlags.bold,
-    'placeholder-color': 'ffaaaaaa',
+    "placeholder-color": "ffAAAAAA",
     height: 30,
     width: 200,
-  }
+  },
 };
 
 export default App;
--- a/packages/react-juce/src/components/TextInput.tsx
+++ b/packages/react-juce/src/components/TextInput.tsx
@@ -1,0 +1,107 @@
+import React, { PropsWithChildren } from 'react';
+
+export interface InputEvent {
+  value: string
+}
+
+export interface ChangeEvent {
+  value: string
+}
+
+export interface TextInputProps {
+  value?: string,
+
+  color?: string,
+  fontSize?: number,
+  fontStyle?: number,
+  fontFamily?: string,
+  justification?: number,
+  'kerning-factor'?: number,
+
+  placeholder?: string,
+  'placeholder-color'?: string,
+  maxlength?: number,
+  readonly?: boolean,
+  'outline-color'?: string,
+  'focused-outline-color'?: string,
+  'highlighted-text-color'?: string,
+  'highlight-color'?: string,
+  'caret-color'?: string,
+
+  onChange?: (e: ChangeEvent) => void,
+  onInput?: (e: InputEvent) => void,
+}
+
+/**
+* TextInput component imitates web's `<input type="text">`.
+* https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input/text
+* So it supports props like `placeholder`, `maxlength`, `onInput`, etc...
+* 
+* TextInput also supports React's controlled components model:
+* https://reactjs.org/docs/uncontrolled-components.html.
+* If the user supplies a `value` prop,
+* then it never renders anything into that text editor other than
+* the string supplied by that `value` prop.
+*/
+export function TextInput(props: PropsWithChildren<TextInputProps | any>) {
+  return React.createElement('TextInput', props, props.children);
+}
+
+/*
+// ================================================
+// Example code
+
+import React, { Component } from 'react';
+
+import {
+  Text,
+  TextInput,
+  View,
+} from 'react-juce';
+
+class App extends Component {
+  constructor(props) {
+    super(props);
+    this._onInput = this._onInput.bind(this);
+
+    this.state = {
+      textValue: '',
+    }
+  }
+
+  _onInput(event) {
+    console.log(`onInput: ${event.value}`);
+    this.setState({textValue: event.value});
+  }
+
+  render() {
+    return (
+      <View>
+        <TextInput
+            value={this.state.textValue}
+            placeholder="init message"
+            onInput={this._onInput}
+            {...styles.text_input}
+          />
+      </View>
+    );
+  }
+}
+
+const styles = {
+  text_input: {
+    backgroundColor: 'ff303030',
+    color: 'ff66FDCF',
+    fontSize: 15.0,
+    fontFamily: 'Menlo',
+    fontStyle: Text.FontStyleFlags.bold,
+    'placeholder-color': 'ffaaaaaa',
+    height: 30,
+    width: 200,
+  }
+};
+
+export default App;
+
+// ================================================
+*/
--- a/react_juce/core/TextInputView.cpp
+++ b/react_juce/core/TextInputView.cpp
@@ -29,31 +29,31 @@ namespace blueprint
         }
     }
 
-    void TextInput::setControlledValue(const juce::String &value)
+    void TextInputView::TextInput::setControlledValue(const juce::String &value)
     {
         insertedAsControlledValue = true;
         setText(value);
     }
 
-    void TextInput::setMaxLength(int maxLen)
+    void TextInputView::TextInput::setMaxLength(int maxLen)
     {
         maxLength = maxLen;
         setInputRestrictions(maxLen);
     }
 
-    void TextInput::setPlaceholderText(const juce::String &text)
+    void TextInputView::TextInput::setPlaceholderText(const juce::String &text)
     {
         placeholderText = text;
         setTextToShowWhenEmpty(placeholderText, placeholderColour);
     }
 
-    void TextInput::setPlaceholderColour(const juce::Colour &colourToUse)
+    void TextInputView::TextInput::setPlaceholderColour(const juce::Colour &colourToUse)
     {
         placeholderColour = colourToUse;
         setTextToShowWhenEmpty(placeholderText, placeholderColour);
     }
 
-    void TextInput::insertTextAtCaret(const juce::String &textToInsert)
+    void TextInputView::TextInput::insertTextAtCaret(const juce::String &textToInsert)
     {
         const juce::String currentValue = getText();
         juce::TextEditor::insertTextAtCaret(textToInsert);
@@ -82,17 +82,17 @@ namespace blueprint
 
     //==============================================================================
 
-    void TextInput::textEditorReturnKeyPressed(juce::TextEditor &)
+    void TextInputView::TextInput::textEditorReturnKeyPressed(juce::TextEditor &)
     {
         invokeChangeEventIfNeeded();
     }
 
-    void TextInput::textEditorFocusLost(juce::TextEditor &)
+    void TextInputView::TextInput::textEditorFocusLost(juce::TextEditor &)
     {
         invokeChangeEventIfNeeded();
     }
 
-    void TextInput::invokeChangeEventIfNeeded()
+    void TextInputView::TextInput::invokeChangeEventIfNeeded()
     {
         if (dirty)
         {
--- a/react_juce/core/TextInputView.cpp
+++ b/react_juce/core/TextInputView.cpp
@@ -64,11 +64,11 @@ namespace blueprint
         }
 
         // Invoke JavaScript's `input` event.
-        if ((*props).contains(TextInputView::onInputProp) && (*props)[TextInputView::onInputProp].isMethod())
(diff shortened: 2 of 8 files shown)


nick-thompson:
Awesome work @utokusa thank you. I've looked briefly and the approach generally looks totally right. I want to review in closer detail before merge, will try to get to it in the morning!

react_juce/core/TextInputView.h
```diff
@@ -0,0 +1,118 @@
...
+     *   only the string value given by `setControlledValue()`
+     *   is shown in the text editor.
+     *
+     * - Allow to set 'placeholder' text and 'placeholder-color' separately.
+     */
+    class TextInput : public juce::TextEditor, public juce::TextEditor::Listener
+    {
+    public:
+        //==============================================================================
+        explicit TextInput(const juce::NamedValueSet *_props)
```
JoshMarler, line 39: Is it possible for this to be made a reference? I might be missing something obvious!

react_juce/core/TextInputView.cpp
```diff
@@ -0,0 +1,225 @@
...
+        textInput.applyColourToAllText(colour);
+
+        int just = props.getWithDefault(justificationProp, 1);
+        textInput.setJustification(just);
+
+        juce::String hexBackgroundColor = props.getWithDefault(backgroundColorProp, "00000000");
+        juce::Colour backgroundColour = juce::Colour::fromString(hexBackgroundColor);
+        textInput.setColour(juce::TextEditor::ColourIds::backgroundColourId, backgroundColour);
+
+        if (name == outlineColorProp)
```
JoshMarler, line 168: It feels like there might be a nice little refactor here to avoid some of this color setter duplication. 

I'm thinking maybe some sort of map of propID => setter which does the following code block a little more generically:

```
juce::String hexFocusedOutlineColor = value;
            juce::Colour focusedOutlineColor = juce::Colour::fromString(hexFocusedOutlineColor);
            textInput.setColour(juce::TextEditor::ColourIds::focusedOutlineColourId, focusedOutlineColor);
```

react_juce/core/TextInputView.cpp
```diff
@@ -0,0 +1,225 @@
...
+            textInput.setReadOnly(value);
+        }
+
+        textInput.applyFontToAllText(getFont());
+
+        juce::String hexColor = props.getWithDefault(colorProp, "ff000000");
+        juce::Colour colour = juce::Colour::fromString(hexColor);
+        textInput.applyColourToAllText(colour);
+
+        int just = props.getWithDefault(justificationProp, 1);
```
JoshMarler, line 161: Super nitpicky but could make this `const int just`

react_juce/core/TextInputView.cpp
```diff
@@ -0,0 +1,225 @@
...
+                throw std::invalid_argument("Invalid prop value. Prop \'caret-color\' must be a color string.");
+            juce::String hexCaretColor = value;
+            juce::Colour caretColor = juce::Colour::fromString(hexCaretColor);
+            textInput.setColour(juce::CaretComponent::ColourIds::caretColourId, caretColor);
+        }
+    }
+
+    void TextInputView::paint(juce::Graphics &g)
+    {
+        View::paint(g);
```
JoshMarler, line 212: Think this `paint` override implementation can be removed if we're just calling through to the base class?

react_juce/core/TextInputView.cpp
```diff
@@ -0,0 +1,225 @@
...
+    }
+
+    void TextInputView::resized()
+    {
+        View::resized();
+        textInput.setBounds(0, 0, getWidth(), getHeight());
```
JoshMarler, line 218: Nitpicky but, might be a bit nicer to use `textInput.setBounds(getLocalBounds())`

react_juce/core/TextInputView.h
```diff
@@ -0,0 +1,118 @@
...
+    /** The TextInputView class is a core view for text editor
+     * within Blueprint's layout system.
+     */
+    class TextInputView : public View
+    {
+    public:
+        //==============================================================================
+        static inline juce::Identifier valueProp = "value";
+
+        static const inline juce::Identifier colorProp = "color";
```
JoshMarler, line 84: Think all these `juce::Identifier` instances should be `static const inline` ?

react_juce/core/TextView.h
```diff
@@ -38,21 +38,27 @@ namespace blueprint
         TextView() = default;
 
         //==============================================================================
-        /** Assembles a Font from the current node properties. */
-        juce::Font getFont()
+        /** Assembles a Font from properties. */
+        static juce::Font getFont(const juce::NamedValueSet &_props)
```
JoshMarler, line 42: Could we remove the underscore here?

react_juce/core/TextView.h
```diff
@@ -38,21 +38,27 @@ namespace blueprint
...
         {
-            float fontHeight = props.getWithDefault(fontSizeProp, 12.0f);
-            int textStyleFlags = props.getWithDefault(fontStyleProp, 0);
+            float fontHeight = _props.getWithDefault(fontSizeProp, 12.0f);
```
JoshMarler, line 44: Super picky but `fontHeight` and `textStyleFlags` could be const

JoshMarler (commented):
LIke the overall shape. Some skin deep comments here. Will take a deeper dive and compare against HTML input API shortly. Nice stuff!

packages/react-juce/src/components/TextInput.tsx
```diff
@@ -0,0 +1,107 @@
...
+
+export interface TextInputProps {
+  value?: string,
+
+  color?: string,
+  fontSize?: number,
+  fontStyle?: number,
+  fontFamily?: string,
+  justification?: number,
+  'kerning-factor'?: number,
```
olidacombe, line 19: Probably not a concern for this PR, but I think we should include some mechanism in https://github.com/nick-thompson/blueprint/blob/master/packages/react-juce/src/lib/Backend.ts for taking any relevant keys from interfaces under components/* for auto camelCase translation.  We only really do css properties at the moment.  That way props can have a consistent case.  I'll raise something for that soon.

react_juce/core/TextInputView.h
utokusa: Thanks @JoshMarler for reviewing! I think it's possible😁
Do you want me to change the member `const juce::NamedValueSet *props` to `const juce::NamedValueSet &props` too?

react_juce/core/TextInputView.h
utokusa: Definitely yes! Sorry for making such a easy mistakes

react_juce/core/TextView.h
utokusa: I think we can, but is it really OK?
If we change `_props` to `props`, it will shadow the member inherited from type `View` and we will get some Clang-Tidy warnings.

JoshMarler: Good spot. `fontProps` maybe?

react_juce/core/TextInputView.h
JoshMarler: No need to apologise! I make them left right and center!

react_juce/core/TextInputView.h
JoshMarler: Yep perfect

utokusa: I fixed it in c724958.

react_juce/core/TextInputView.cpp
utokusa: I made a change to remove duplication (d7c2299).
Also I stopped providing default value to 'backgound-color' for simplicity.
Now I think if we really want default values, it is better to add default values to all props in `textEditorColourIdsByProp`.
What do you think @JoshMarler?

react_juce/core/TextInputView.cpp
utokusa: I fixed it in c724958.

react_juce/core/TextInputView.cpp
utokusa: I fixed it in c724958.

react_juce/core/TextInputView.cpp
utokusa: I fixed it in c724958.

react_juce/core/TextInputView.h
utokusa: I fixed it in c724958.

react_juce/core/TextView.h
utokusa: Good idea! Simple and easy to understand.
I fixed it in c724958.

react_juce/core/TextView.h
utokusa: I fixed it in c724958.

react_juce/core/TextInputView.h
```diff
@@ -0,0 +1,120 @@
...
+     *   only the string value given by `setControlledValue()`
+     *   is shown in the text editor.
+     *
+     * - Allow to set 'placeholder' text and 'placeholder-color' separately.
+     */
+    class TextInput : public juce::TextEditor, public juce::TextEditor::Listener
+    {
+    public:
+        //==============================================================================
+        explicit TextInput(const juce::NamedValueSet &_props)
```
JoshMarler, line 39: I think I'd move definition of constructor into the cpp here personally.

react_juce/core/TextView.h
```diff
@@ -38,21 +38,27 @@ namespace blueprint
...
+            if (fontProps.contains(fontFamilyProp))
+                f = juce::Font (fontProps[fontFamilyProp], fontHeight, textStyleFlags);
 
-            f.setExtraKerningFactor(props.getWithDefault(kerningFactorProp, 0.0));
+            f.setExtraKerningFactor(fontProps.getWithDefault(kerningFactorProp, 0.0));
             return f;
         }
 
+        /** Assembles a Font from the current node properties. */
+        juce::Font getFont()
```
JoshMarler, line 57: This should be const if possible `juce::Front getFont() const`

react_juce/core/TextInputView.cpp
```diff
@@ -0,0 +1,207 @@
...
+
+    void TextInput::setPlaceholderColour(const juce::Colour &colourToUse)
+    {
+        placeholderColour = colourToUse;
+        setTextToShowWhenEmpty(placeholderText, placeholderColour);
+    }
+
+    void TextInput::insertTextAtCaret(const juce::String &textToInsert)
+    {
+        juce::String currentValue = getText();
```
JoshMarler, line 58: `currentValue` and `newValue` could be `const`

react_juce/core/TextInputView.cpp
```diff
@@ -0,0 +1,207 @@
...
+        {
+            if (!value.isString())
+                throw std::invalid_argument("Invalid prop value. Prop \'placeholder\' must be a string.");
+            textInput.setPlaceholderText(value);
+        }
+        if (name == placeholderColorProp)
+        {
+            if (!value.isString())
+                throw std::invalid_argument("Invalid prop value. Prop \'placeholder-color\' must be a color string.");
+            juce::String hexPlaceholderColor = value;
```
JoshMarler, line 140: `const` vars here.

react_juce/core/TextInputView.cpp
```diff
@@ -0,0 +1,207 @@
...
+            textInput.setMaxLength(value);
+        }
+        if (name == readonly)
+        {
+            textInput.setReadOnly(value);
+        }
+
+        textInput.applyFontToAllText(getFont());
+
+        juce::String hexColor = props.getWithDefault(colorProp, "ff000000");
```
JoshMarler, line 157: Bit more `const` ere.

react_juce/core/TextInputView.cpp
```diff
@@ -0,0 +1,207 @@
...
+    juce::Font TextInputView::getFont()
+    {
+        return TextView::getFont(props);
+    }
+
+    bool TextInputView::isTextEditorColorProp(const juce::Identifier &textEditorColorProp)
+    {
+        auto it = textEditorColourIdsByProp.find(textEditorColorProp);
+        bool found = (it != textEditorColourIdsByProp.end());
+        return found;
```
JoshMarler, line 183: Could tidy this up a little

```
{
    const auto it = textEditorColourIdsByProp.find(textEditorColorProp);
    return it != textEditorColourIdsByProp.end();
}
```

react_juce/core/TextInputView.cpp
```diff
@@ -0,0 +1,207 @@
...
+    }
+
+    void TextInputView::setTextEditorColorProp(const juce::Identifier &textEditorColorProp, const juce::var &value)
+    {
+        if (!value.isString())
+        {
+            std::string propName = textEditorColorProp.toString().toStdString();
```
JoshMarler, line 190: Most of these inline vars in `setTextEditorColorProp` can be `const`

JoshMarler: This is much nicer though! Good stuff.

react_juce/core/TextInputView.cpp
```diff
@@ -0,0 +1,207 @@
...
+        : textInput(props)
+    {
+        addAndMakeVisible(textInput);
+        textInput.addListener(&textInput);
+        textInput.setPopupMenuEnabled(false);
+    }
+
+    void TextInputView::setProperty(const juce::Identifier &name, const juce::var &value)
+    {
+        View::setProperty(name, value);
```
JoshMarler, line 122: Could we maybe get some newlines between these if statements? Feels a bit more readable that way. Personally I'd also have a newline below the `throw` statements.

react_juce/core/TextInputView.cpp
```diff
@@ -0,0 +1,207 @@
...
+    }
+
+    void TextInput::textEditorFocusLost(juce::TextEditor &)
+    {
+        invokeChangeEventIfNeeded();
+    }
+
+    void TextInput::invokeChangeEventIfNeeded()
+    {
+        if (dirty)
```
JoshMarler, line 97: Good catch/idea with the `dirty` flag!

react_juce/core/TextInputView.h
```diff
@@ -0,0 +1,120 @@
...
+     *
+     * - Support React's controlled component model.
+     *   https://reactjs.org/docs/uncontrolled-components.html
+     *   If field `controlled` is true,
+     *   only the string value given by `setControlledValue()`
+     *   is shown in the text editor.
+     *
+     * - Allow to set 'placeholder' text and 'placeholder-color' separately.
+     */
+    class TextInput : public juce::TextEditor, public juce::TextEditor::Listener
```
JoshMarler, line 35: I'd actually be tempted to make this a nested class in `TextView`. What do you think @nick-thompson ?

react_juce/core/TextInputView.cpp
JoshMarler: Yeah. I'd even be tempted to make such things default prop values in React so users of the package can clearly see what the default color/control values will be. What do you think @nick-thompson ?

react_juce/core/TextInputView.h
utokusa: Agree.
  It seems that all existing constructor only with few lines are put in header.
  I also changed `TextInput` class's constructor's coding style to match other constructors!
973a63c

react_juce/core/TextView.h
utokusa: OK!
Maybe I can use handle better next time :)
973a63c

react_juce/core/TextInputView.cpp
utokusa: I fixed it in 973a63c.

react_juce/core/TextInputView.cpp
utokusa: I fixed it in 973a63c.

react_juce/core/TextInputView.cpp
utokusa: I fixed it in 973a63c.

react_juce/core/TextInputView.cpp
utokusa: I fixed it in 973a63c.

react_juce/core/TextInputView.cpp
utokusa: Thanks!
I added const in 973a63c.

react_juce/core/TextInputView.cpp
utokusa: I fixed it in 973a63c.

react_juce/core/TextInputView.cpp
utokusa: Thank you @JoshMarler!

JoshMarler:
Excellent work @utokusa. This looks pretty much ready to merge now. We'll need to publish a new version of the `react-juce` npm package for this and bump the example project etc. Very happy to pull down your branch to test and push the package changes etc. 

@nick-thompson did you want to take a quick pass over this before merge?

react_juce/core/TextInputView.cpp
```diff
@@ -0,0 +1,207 @@
...
+        if (props.contains(TextInputView::onInputProp) && props[TextInputView::onInputProp].isMethod())
+        {
+            std::array<juce::var, 1> args{{detail::makeInputEventObject(newValue)}};
+            juce::var::NativeFunctionArgs nfArgs(juce::var(), args.data(), static_cast<int>(args.size()));
+            std::invoke(props[TextInputView::onInputProp].getNativeFunction(), nfArgs);
+        }
+
+        dirty = true;
+
+        if (controlled && !insertedAsControlledValue)
```
nick-thompson, line 76: Wondering if we can move this check to the top of the function? Seems that we might be able to just return early and no-op if we know that we're a controlled component? It seems odd that we will have just invoked `onInput` event too if we know we're going to undo it immediately after, but maybe that matches ReactDOM's behavior anyway?

react_juce/core/TextInputView.h
nick-thompson: On account of namespacing? Or actually, marking it as a private nested class makes sense to me, as this class interface is really a private implementation detail, it's not a user-facing API that we anticipate maintaining.

Yea, I'm :+1: for it

nick-thompson (approved):
@utokusa @JoshMarler sorry I've been slow on this one!

I just took a detailed pass, this is near perfect, @utokusa thanks so much. I would be happy to merge this as is, but I left two quick comments inline for your discussion.

Let me know what you think, and we'll get this merged ASAP

react_juce/core/TextInputView.cpp
utokusa: Thanks @nick-thompson for detailed review!
And I'm OK for making it better!

I want make sure what you're saying.
Do you want me to change the code like this?

```cpp
    void TextInput::insertTextAtCaret(const juce::String &textToInsert)
    {
        if (controlled && !insertedAsControlledValue)
            return;

        // same as before

        undo();
        insertedAsControlledValue = false;
    }
```

If so, I don't agree with the idea.
If we adopt it, input event will never be invoked for controlled components and there will be no way to know characters entered in the controlled text editor.

I might misunderstand what you say.
If so, please let me know!

react_juce/core/TextInputView.h
utokusa: I also think it's much better!
I will fix it.

react_juce/core/TextInputView.cpp
nick-thompson: Ohh of course, of course. I misunderstood, you're right!

Let's keep it as is, I'll go ahead and merge :)

Thank you again!

react_juce/core/TextInputView.cpp
nick-thompson: Following up on default prop values in React components: yea, I agree that we should generally apply default properties at the React layer. For example

```js
function TextInput(props) {
    const defaultProps = {
        ...our defaults here...
    };
    return React.createElement('TextInput', Object.assign({}, defaultProps, props), props.children);
}
```

That way we can set defaults and change defaults re-using all the same code paths that we're used to.

I want to say too, though, that we should _still_ use `props.getWithDefault` when we're trying to read property values inside our `setProperty` overrides. The reason for that is that `setViewProperty` gets called many times during a React update batch, and so our `setProperty` overrides get called many times as well. For any given invocation of `setProperty`, we have no guarantees that _all_ of our properties have been set already. So even if we're relying on defaults at the React layer, we have to be prepared for the fact that we might not have all of our properties during any given invocation of a `setProperty` override, hence my suggestion that we should be using `props.getWithDefault` there generally.

This behavior can change slightly once we introduce the StyleSheet API, which I'll open an issue for now.

That all said, I think we're good to merge this PR!

nick-thompson:
@utokusa looks like we have a merge conflict here, would you mind rebasing your branch against the latest master and then force pushing? And then I'll go ahead and merge!

utokusa:
@nick-thompson I have just rebased and force pushed.
Now the commit log is bit strange and my non-latest commits fails to build because of the namespace change.
So I think it might be better to squash them.
What do you think?
Either way, it would be helpful if you could check for the latest commit.

utokusa:
I noticed that the GitHub Action was failed.
Please wait and I will fix it.

utokusa:
I have just fixed it.
The CI really helped me :)

nick-thompson:
Excellent, thanks again @utokusa! 😃 🚀

PR #218, opened 2021-01-24
Improve image cache usage and error handling

stfufane:
Hey there,

As discussed [here](https://github.com/nick-thompson/blueprint/pull/211#issuecomment-766103859) I added some better cache and error management, let me know what you think :)

Cheers!

### DIFF
--- a/react_juce/core/ImageView.cpp
+++ b/react_juce/core/ImageView.cpp
@@ -35,56 +35,34 @@ namespace reactjuce
         {
             const juce::String source = value.toString();
             const int sourceHash      = juce::DefaultHashFunctions::generateHash(source, 1024);
-            const juce::URL sourceURL = source;
 
             // No need to recalculate the image if it's already in the ImageCache.
             const juce::Image& cachedImage = juce::ImageCache::getFromHashCode(sourceHash);
             if (cachedImage.isValid())
+                return setDrawableImage(cachedImage, sourceHash);
+
+            try
             {
-                setDrawableImage(cachedImage, sourceHash);
+                const juce::URL sourceURL = source;
+
+                // Web images are downloaded in a separate thread to avoid blocking.
+                if (isWellFormedURL(sourceURL) && sourceURL.isProbablyAWebsiteURL(sourceURL.toString(false)))
+                    return downloadImageAsync(source); 
+                    
+                if (sourceURL.isLocalFile())
+                    return setDrawableImage(loadImageFromFileURL(sourceURL), sourceHash);
+                    
+                if (source.startsWith("data:image/")) // juce::URL does not currently handle Data URLs
+                    return setDrawableImage(loadImageFromDataURL(source), sourceHash);
+
+                // Last case left, possibly SVG/Image data.
+                return setDrawableData(source);
             }
-            else
+            catch (const std::exception& l)
             {
-                try
-                {
-                    if (isWellFormedURL(sourceURL))
-                    {
-                        // Web images are downloaded in a separate thread to avoid blocking.
-                        if (sourceURL.isProbablyAWebsiteURL(sourceURL.toString(false)))
-                        {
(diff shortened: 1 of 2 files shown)


react_juce/core/ImageView.cpp
```diff
@@ -34,47 +34,58 @@ namespace reactjuce
...
         {
             const juce::String source = value.toString();
-            const juce::URL    sourceURL = source;
+            const int sourceHash      = juce::DefaultHashFunctions::generateHash(source, 1024);
+            const juce::URL sourceURL = source;
 
-            if (isWellFormedURL(sourceURL))
+            // No need to recalculate the image if it's already in the ImageCache.
+            const juce::Image& cachedImage = juce::ImageCache::getFromHashCode(sourceHash);
+            if (cachedImage.isValid())
```
nick-thompson, line 42: Yep, perfect.

One comment here which is mostly a style comment: in this case if we can find the image in the cache, we don't even need to run the rest of the function. In these cases, I prefer an early return:

```
if (cachedImage.isValid())
    return setDrawableImage(cachedImage, sourceHash);

// No longer need the "else" block because control flow will only reach down here if
// we need to run the rest of the function
```

Basically let's us keep the indenting easier to read

react_juce/core/ImageView.cpp
```diff
@@ -98,6 +109,39 @@ namespace reactjuce
...
+    void ImageView::setDrawableImage(const juce::Image& image, const int sourceHash)
+    {
+        // Add the freshly retrieved image to the cache.
+        juce::ImageCache::addImageToCache(image, sourceHash);
+
+        auto drawableImg = std::make_unique<juce::DrawableImage>();
+        drawableImg->setImage(image);
+        drawable = std::move(drawableImg);
+
+        sendOnLoadCallback();
```
nick-thompson, line 142: I think we should `repaint()` here too, right?

react_juce/core/ImageView.cpp
```diff
@@ -120,7 +164,9 @@ namespace reactjuce
                 if (!juce::URL(source).readEntireBinaryStream(mb)) 
                 {
                     shouldDownloadImage = true;
```
nick-thompson, line 166: Hm, realizing this, shouldn't we not need this here? This would essentially mean "keep trying to download it" if I understand correctly? We could in theory end up in an endless loop trying to reach the same server and constantly failing, no?

react_juce/core/ImageView.cpp
```diff
@@ -34,47 +34,58 @@ namespace reactjuce
...
+                setDrawableImage(cachedImage, sourceHash);
+            }
+            else
+            {
+                try
                 {
-                    const auto sourceHash = juce::DefaultHashFunctions::generateHash(source, 1024);
-                    const auto& cachedImage = juce::ImageCache::getFromHashCode(sourceHash);
-                    if (cachedImage.isValid())
+                    if (isWellFormedURL(sourceURL))
```
nick-thompson, line 50: Same here I think we can collapse some of this with early returns.

```
if (sourceURL.isLocalFile())
    return setDrawableImage(loadImageFromFileURL(sourceURL));
    
if (sourceURL.isProbablyAWebsiteURL(source) && isWellFormedURL(sourceURL))
    return downloadImageAsync(source);

if (sourceURL.startsWith("data:image/")
    return setDrawableImage(....);
    
throw std::logic_error("Unsupported image URL:" + source);
```

That feels a bit neater to me. What do you think?

nick-thompson (commented):
This looks great @stfufane. I left a few style/nitpicky comments inline, let me know what you think

stfufane:
Good point on the remarks :) Should be better now!

nick-thompson (approved):
Perfect @stfufane thank you!

nick-thompson:
I'm going to squash these two commits into one as I merge in

PR #219, opened 2021-01-25
update package.json to publish types to npm module

tobq:


nick-thompson:
Thanks @tobq, I appreciate it!

Merging now, I'm going to squash on the way in, as this makes sense as a single commit to me

PR #220, opened 2021-01-25
Implement dispatch event to pass gain level values to Meter component

ilionic:
Fixes Meter in GainPlugin example.

* Read `gainPeakValue` in processBlock.

* Initialise 100ms (arbitrary number) Timer to call `dispatchEvent` passing gain values to `Meter` component

* Add public `getReactAppRoot ` getter to `GenericEditor` in order to call `dispatchEvent` from `appRoot` instance created by it as suggested by Nick.

### DIFF
--- a/examples/GainPlugin/PluginProcessor.cpp
+++ b/examples/GainPlugin/PluginProcessor.cpp
@@ -166,6 +166,9 @@ void GainPluginAudioProcessor::processBlock (AudioBuffer<float>& buffer, MidiBuf
         if (muted)
             buffer.applyGain(0.0f);
     }
+    
+    // Read current block gain peak value
+    gainPeakValue = buffer.getMagnitude (0, buffer.getNumSamples());
 }
 
 //==============================================================================
--- a/examples/GainPlugin/PluginProcessor.cpp
+++ b/examples/GainPlugin/PluginProcessor.cpp
@@ -47,6 +47,11 @@ GainPluginAudioProcessor::GainPluginAudioProcessor()
 {
 }
 
+GainPluginAudioProcessor::~GainPluginAudioProcessor()
+{
+    Timer::stopTimer();
+}
+
 //==============================================================================
 const String GainPluginAudioProcessor::getName() const
 {
@@ -186,8 +191,23 @@ AudioProcessorEditor* GainPluginAudioProcessor::createEditor()
     editor->setResizeLimits(400, 240, 400 * 2, 240 * 2);
     editor->getConstrainer()->setFixedAspectRatio(400.0 / 240.0);
     editor->setSize (400, 240);
+    
+    // Get appRoot reference created by GenericEditor in order to dispatch events
+    appRoot = &editor->getReactAppRoot();
+    // Start timer to dispatch gainPeakValues event to update Meter values
+    Timer::startTimer(100);
 
     return editor;
+}
+
+void GainPluginReactAudioProcessor::timerCallback()
+{
+    // Dispatch gainPeakValues event used by Meter React component
+    appRoot->dispatchEvent(
+        "gainPeakValues",
+        static_cast<float>(gainPeakValue),
+        static_cast<float>(gainPeakValue)
+    );
 }
 
 //==============================================================================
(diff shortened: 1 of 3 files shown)


examples/GainPlugin/PluginProcessor.cpp
```diff
@@ -186,10 +194,25 @@ AudioProcessorEditor* GainPluginAudioProcessor::createEditor()
     editor->setResizeLimits(400, 240, 400 * 2, 240 * 2);
     editor->getConstrainer()->setFixedAspectRatio(400.0 / 240.0);
     editor->setSize (400, 240);
+    
+    // Get appRoot reference created by GenericEditor in order to dispatch events
+    appRoot = &editor->getReactAppRoot();
```
nick-thompson, line 199: This is a bit dangerous here as once we hand over the pointer to our GenericEditor to the caller, we have no control over the lifetime of that object. The host (e.g. a DAW) can close the editor at will and totally invalidate this pointer, which we've saved here and call in the timer callback.

The safer move would be this:
* Don't hang onto the `appRoot` pointer here.
* In the timer callback, we use `getActiveEditor` to see if we still have an editor alive: https://docs.juce.com/master/classAudioProcessor.html#a0f1386123ab008b42a82c7d121b2ff13

Then `timerCallback` can look more like this:

```
if (auto* editor = dynamic_cast<reactjuce::GenericEditor*>(getActiveEditor())
{
    // Here we know we have a safe handle to the editor
    editor->getReactAppRoot().dispatchEvent(
        "gainPeakValues",
        static_cast<float>(gainPeakValue),
        static_cast<float>(gainPeakValue)
    );
}

**Review by nick-thompson — CHANGES_REQUESTED** [review]
Excellent @ilionic, thank you!

One quick comment inline about safely handling access to the editor

File: examples/GainPlugin/PluginProcessor.cpp (hunk 1 above)
Comment (ilionic) [reply examples/GainPlugin/PluginProcessor.cpp]: Thanks! That was challenging bit I wasn't sure about. Now clear!

File: examples/GainPlugin/PluginProcessor.h [hunk 2]
```diff
@@ -15,7 +15,7 @@
 //==============================================================================
 /**
 */
-class GainPluginAudioProcessor  : public AudioProcessor
+class GainPluginAudioProcessor  : public AudioProcessor, public Timer
```
JoshMarler, line 18: Nice stuff @ilionic !

Super minor comment from me here. Would inherit privately from `juce::Timer` here.

ilionic: Ah ok, updated. I actually had it all namespaced since I am using Juce 6 and removed it to adapt to Juce 5. 
Btw, perhaps worth considering updating example to Juce6?

JoshMarler: Ahh sorry, I meant private inheritance rather than namespacing!

So like:

```
class GainPluginAudioProcessor  : public AudioProcessor, private Timer
```

JoshMarler: But yeah we should probably update the code accordingly here!

ilionic: @JoshMarler do you want me to update to Juce6 part of this PR or separate? There are not many changes, few places need to be prefixed with juce namespaces.

JoshMarler: Ahhh no don't worry about that right now. Lets keep this commit as a separate thing as you've kept it really nice and clean. Small concise commits FTW

ilionic: Ok updated to private Timer. Removed juce namespacing prefix for later when updating to Juce 6 to keep things consistent.

JoshMarler:
@nick-thompson , happy for this to be merged now?

nick-thompson:
Yep this is excellent! Thanks very much @ilionic 🙏 , I'm going to squash these commits into one on the way in

nick-thompson (approved):

ilionic:
Noticed build failed due to Juce 6 namespacing left overs, sorry about that :)

nick-thompson:
No worries! I'm surprised I didn't notice it before hitting the merge button 🤷‍♂️ . Anyways, all is good now

ilionic:
Actually might be good idea to run checks on PR's as well, not only upon master merge?

nick-thompson:
Definitely!

PR #221, opened 2021-01-27
Optimise CanvasView implementation

JoshMarler:
This change helps to reduce the amount of JS memory churn caused by
    CanvasView when painting.


nick-thompson:
I think we can close this @JoshMarler? We merged the canvas changes earlier and it looks like we've already got `Bump react-juce@0.2.11` on `master` now?

I'll go ahead and close, please reopen if I've misunderstood

PR #227, opened 2021-02-09
Add hermes js engine

JoshMarler:
@nick-thompson, here is a first pass at Hermes JS integration. 

There are a couple of TODOs in the code that might warrant discussion. I'm going to stick a couple of questions in the comments for us to run over.

I'll rebase and squash this all down once we're ready. Will follow up with a PR to implement Hermes Debugger hooks after this goes in.


CMakeLists.txt
```diff
@@ -5,11 +5,52 @@ project(blueprint VERSION 0.1.0)
...
+set(REACTJUCE_JS_LIBRARY HERMES CACHE STRING "The JS Engine to use: either HERMES or DUKTAPE")
+
 add_subdirectory(ext/juce)
 
 # Adding any custom modules you might have:
 juce_add_module(react_juce)
 
+# Setup the JS engine/interpreter to use.
+if (REACTJUCE_JS_LIBRARY STREQUAL "HERMES")
+    #TODO: Review Question:
```
JoshMarler, line 17: @nick-thompson, TODO here worth discussing.

react_juce/core/View.cpp
```diff
@@ -9,6 +9,8 @@
 
 #include "View.h"
 
+#include <climits>
```
JoshMarler, line 12: @nick-thompson, build fix I required on Linux. I think you added this to your quickjs branch too

react_juce/core/EcmascriptEngine_Hermes.cpp
```diff
@@ -0,0 +1,522 @@
...
+
+                    return juce::var(memBlock);
+                }
+            }
+
+            return {};
+        }
+
+        //==============================================================================
+        jsi::Object varObjectToJSIObject(juce::DynamicObject *v, jsi::Runtime &runtime)
```
JoshMarler, line 138: @nick-thompson , so this is where I have an implementation question. 

At the moment this conversion assumes that the `juce::DynamicObject` instance passed in will live to match the lifetime of the equivalent `jsi` object type. We make this same assumption in the duktape implementation. The same concern applies to the other side of the conversion `jsiObjectToVarObject`. Do we think this is fine? The other approach I had for this to ensure lifetime safety was as below:

```
+        class JSIDynamicObject : public jsi::HostObject
+        {
+        public:
+            explicit JSIDynamicObject(juce::DynamicObject *obj)
+                : objPtr(obj)
+            {
+                jassert(objPtr);
+            }
+
+            jsi::Value get(jsi::Runtime& rt, const jsi::PropNameID& name) override
+            {
+                return varToJSIValue(objPtr->getProperty(name.utf8(rt).c_str()), rt);
+            }
+
+            void set(jsi::Runtime& rt, const jsi::PropNameID& name, const jsi::Value& value)
+            {
+                objPtr->setProperty(name.utf8(rt).c_str(), jsiValueToVar(value, rt));
+            }
+
+            std::vector<jsi::PropNameID> getPropertyNames(jsi::Runtime& rt)
+            {
+                std::vector<jsi::PropNameID> props;
+                for (auto &p : objPtr->getProperties())
+                {
+                    props.push_back(jsi::PropNameID::forAscii(rt, p.name.getCharPointer()));
+                }
+
+                return props;
+            }
+
+        private:
+            juce::DynamicObject::Ptr objPtr;
+        };
+

         jsi::Object varObjectToJSIObject(juce::DynamicObject *v, jsi::Runtime &runtime)
         {
-            jassert(v);
-
-            //TODO: Should we be using jsi::HostObject here and binding to DynamicObject? Probably
-            //      in a full implementation.
-            //auto ho = std::make_shared<jsi::HostObject>();
-            jsi::Object jsiObj(runtime);
-
-            for (auto &prop : v->getProperties())
-            {
-                jsiObj.setProperty(runtime, prop.name.getCharPointer(), varToJSIValue(prop.value, runtime));
-            }
-
-            return jsiObj;
-            //return jsi::Object::createFromHostObject(runtime, ho);
+            auto obj = std::make_shared<JSIDynamicObject>(v);
+            return jsi::Object::createFromHostObject(runtime, obj);
         }
```

This approach doesn't work quite as nicely for the `jsiObjectToVarObject` side. I'm not sure whether guarding against lifetime issues is something we need to protect against here or if the current implementation is OK. What do you think? This is mainly important when users push some custom DynamicObject instance through to JS which invokes native methods etc. etc.

react_juce/core/EcmascriptEngine_Hermes.cpp
```diff
@@ -0,0 +1,522 @@
...
+                auto nf = v.getNativeFunction();
+                return varMethodToJSIMethod(nf, runtime);
+            }
+
+            if (v.isObject())
+                return varObjectToJSIObject(v.getDynamicObject(), runtime);
+
+            if (v.isBinaryData())
+            {
+                //TODO: Currently appears no way to create a jsi::ArrayBuffer from C++.
```
JoshMarler, line 220: This is annoying. I'll speak to the Hermes maintainers about this once I've fired other contributions back but in principle I don't think it's likely to be an issue for 99% of users right now.

react_juce/core/EcmascriptEngine_Hermes.cpp
```diff
@@ -0,0 +1,522 @@
...
+
+        void registerNativeProperty(const juce::String &target, const juce::String &name, const juce::var &value)
+        {
+            auto obj = runtime->global().getPropertyAsObject(*runtime, target.toRawUTF8());
+            obj.setProperty(*runtime, name.toRawUTF8(), varToJSIValue(value, *runtime));
+        }
+
+        //==============================================================================
+        //TODO: We can probably improve on this by refining the invoke API
+        juce::var invoke(const juce::String &name, const std::vector<juce::var> &vargs)
```
JoshMarler, line 385: @nick-thompson, I think this `invoke` approach is fine for the moment using dot separated accessor strings but it feels a bit cumbersome. Perhaps not essential for this v1 merge but I wonder if we can do something more sophisticated with the `invoke` API down the line. I'm not exactly sure what that would be though ...

react_juce/core/EcmascriptEngine_Hermes.cpp
```diff
@@ -0,0 +1,522 @@
...
+        void reset()
+        {
+            if (timeoutsManager)
+                timeoutsManager->clear();
+
+            runtime         = facebook::hermes::makeHermesRuntime();
+            timeoutsManager = std::make_unique<TimeoutFunctionManager>(*runtime);
+
+            // Quick and dirty console object provide. Could be improved upon.
+            jsi::Function logFunction =
```
JoshMarler, line 472: @nick-thompson , Curious to hear if you think this is a sufficient enough implementation for `console.log()` bindings for now?

CMakeLists.txt
JoshMarler: Also, this is the new approach I imagined for making the JS library/engine use easily switchable by the user. Doing this all I need to do is change one variable in the cmake cache for `REACTJUCE_JS_LIBRARY` and everything "just works"

.gitmodules
```diff
@@ -4,3 +4,6 @@
 [submodule "yoga"]
 	path = react_juce/yoga
 	url = ../../facebook/yoga.git
+[submodule "react_juce/hermes"]
+	path = react_juce/hermes
+	url = https://github.com/facebook/hermes.git
```
nick-thompson, line 9: Mind editing this gitmodules file to use a relative submodule URL like above? This way if a user clones the project with HTTP, the submodules will follow suit, and likewise for SSH.

Maybe not a big deal since all these submodules are open source anyway, but that's a practice I try to maintain for submods

CMakeLists.txt
nick-thompson: Yep, love this approach, I think the only other thing would be to add another var up top that specifies the path to hermes, allows a user override, and defaults to react_juce/hermes, right?

examples/GainPlugin/CMakeLists.txt
```diff
@@ -59,7 +59,7 @@ target_compile_definitions(GainPlugin
 # dependencies are resolved automatically, so `juce_core`, `juce_events` and so on will also be
 # linked automatically. If we'd generated a binary data target above, we would need to link to it
 # here too. This is a standard CMake command.
-target_link_libraries(GainPlugin PRIVATE
+target_link_libraries(GainPlugin PUBLIC
```
nick-thompson, line 62: I'm really a cmake newb still, what's the difference here?

react_juce/core/EcmascriptEngine.h
```diff
@@ -67,6 +67,17 @@ namespace reactjuce
...
+        /**
+         * An evaluate overload for engines which support loading precompiled
+         * JS bytecode.
+         * @param code The precompiled bytecode filed to evaluate.
+         * @return juce::var result of the evaluation
+         * @throws EcmascriptEngine::Error in the event of an evaluation error
+         *         or when called on an engine instance which does not support
+         *         loading of precompiled bytecode.
+         */
+        juce::var evaluateBytecode(const juce::File &code);
```
nick-thompson, line 79: :+1:

react_juce/core/View.cpp
nick-thompson: Yep :+1:. I'll address mine when I rebase the qjs branch onto this

react_juce/core/EcmascriptEngine_Hermes.cpp
```diff
@@ -0,0 +1,522 @@
+#include "EcmascriptEngine.h"
+
+#include <hermes/hermes.h>
```
nick-thompson, line 3: No pragma warning pushes?? 😱 

Good work hermes ;)

react_juce/core/EcmascriptEngine_Hermes.cpp
```diff
@@ -0,0 +1,522 @@
...
+            }
+
+            return varArray;
+        }
+
+        juce::var::NativeFunction jsiMethodToVarMethod(jsi::Function v, jsi::Runtime &runtime)
+        {
+            auto fPtr = std::make_shared<CopyableJSIMethodWrapper>(std::move(v));
+
+            return [fPtr = fPtr, &rt = runtime] (const juce::var::NativeFunctionArgs &args)
```
nick-thompson, line 64: Do we only need the `CopyableJSIMethodWrapper` helper so that we can capture here? If so, I think we can clean it up by moving in the capture group:

```
return [fPtr = std::move(v), &rt = runtime](...blah...)
```

If that works, let's go for it and drop the CopyableJSIMethodWrapper code

react_juce/core/EcmascriptEngine_Hermes.cpp
```diff
@@ -0,0 +1,522 @@
...
+
+                if (obj.isHostObject(runtime))
+                {
+                    auto hostObjectPtr = obj.getHostObject(runtime);
+                    return jsiObjectToVarObject(*hostObjectPtr, runtime);
+                }
+
+                if (obj.isArrayBuffer(runtime))
+                {
+                    jsi::ArrayBuffer arrayBuffer = obj.getArrayBuffer(runtime);
```
nick-thompson, line 126: IIInnnnteresting. Are we using these array buffers anywhere? Feel like there's probably some cool stuff we could do with array buffers

react_juce/core/EcmascriptEngine_Hermes.cpp
```diff
@@ -0,0 +1,522 @@
...
+                    juce::MemoryBlock memBlock(static_cast<const void*>(arrayBuffer.data(runtime)),
+                                                                        arrayBuffer.size(runtime));
+
+                    return juce::var(memBlock);
+                }
+            }
+
+            return {};
```
nick-thompson, line 134: Maybe `jassertfalse` here since we expect not to hit this? Or at least to alert us when we do hit it

react_juce/core/EcmascriptEngine_Hermes.cpp
nick-thompson: Good question. So, in the js->native direction, most of the work that goes into marshaling an object to a juce::DynamicObject will copy primitive values (even if deeply nested). I think functions are really the only place to be careful, and our duktape approach gets by with this by sticking a global reference to the function in one of the private stashes so that the function won't disappear.

Here too we might be ok for functions, under one condition... Above, when we `std::move` the jsi::Function into the capture group of the juce::var::NativeFunction, that capture group will stay alive as long as somebody's holding a reference to the NativeFunction. So, if the `jsi::Function` itself implements the reference counting for the js func in the runtime, we won't drop that last count until the last reference to the NativeFunction in cpp drops. In that case I think we should be fine. Worth checking that detail of jsi::Function though.

Finally, for the native->js direction, yes you're super right, but I think this is not necessarily our responsibility. Let's make sure that wherever we push native->js for internal use, we are careful to protect lifetimes. But we don't need to jump through hoops to protect the end user from that case, I think the end user needs to be aware of their own lifetime management as they push values from native->js.

So all in all, I think the current approach is totally fine, and we should maybe just check that detail of jsi::Function

react_juce/core/EcmascriptEngine_Hermes.cpp
```diff
@@ -0,0 +1,522 @@
...
+            }
+
+            return jsiObj;
+        }
+
+        jsi::Function varMethodToJSIMethod(juce::var::NativeFunction &v, jsi::Runtime &runtime)
+        {
+            return jsi::Function::createFromHostFunction(
+                runtime,
+                jsi::PropNameID::forAscii(runtime, ""),
```
nick-thompson, line 155: What is this?

react_juce/core/EcmascriptEngine_Hermes.cpp
```diff
@@ -0,0 +1,522 @@
...
+                0,
+                [v = v](jsi::Runtime& rt, const jsi::Value& thisVal, const jsi::Value* args, size_t count)
```
nick-thompson, line 157: Love how easy that is. Believe we could also `v = std::move(v)` here if we want

react_juce/core/EcmascriptEngine_Hermes.cpp
nick-thompson: Yea, no worries. I don't think anyone will use this yet

react_juce/core/EcmascriptEngine_Hermes.cpp
```diff
@@ -0,0 +1,522 @@
...
+            jsi::Value newTimeout(jsi::Function f, const int timeoutMillis, std::vector<jsi::Value> args, const bool repeats=false)
+            {
+                static int nextId = 0;
+                timeoutFunctions.emplace(nextId, TimeoutFunction(std::move(f), std::move(args), repeats));
+                startTimer(nextId, timeoutMillis);
+
+                return nextId++;
+            }
+
+            void timerCallback(int id) override
```
nick-thompson, line 276: Yup, great. Need to adopt this in my quickjs branch

react_juce/core/EcmascriptEngine_Hermes.cpp
nick-thompson: Yep, I think so. I think in the long run we will want to make the `console.log` configurable at the EcmascriptEngine interface so people can slice in whatever logging infrastructure they want, but for now this is fine

react_juce/core/EcmascriptEngine_Hermes.cpp
nick-thompson: Agreed. I was thinking initially that the implementation here would basically just call `evaluateInliine` with the string argument, receive a function object back, and then invoke it with the args. But, that's a little cumbersome. 

>  I'm not exactly sure what that would be though ...

Agreed :) Let's keep it as is for now and iterate on this as we have a more clear idea of what the API should be

nick-thompson (commented):
Boom!!! This is a huge win for the project, thanks a million @JoshMarler.

Left a bunch of comments inline. Overall shape is excellent, really just a few minor thoughts/comments/questions

CMakeLists.txt
JoshMarler: Totally. I'm gonna experiment with this just to make sure we don't annihilate include paths via `juce_add_module` etc. if we follow suit with Duktape path.

.gitmodules
JoshMarler: Ah yeah my bad. Does that mean you need to push up a hermes fork to your user account?

examples/GainPlugin/CMakeLists.txt
JoshMarler: Ah I think that might be a left over error! Will fix.

react_juce/core/EcmascriptEngine_Hermes.cpp
JoshMarler: Hahah I know right.

react_juce/core/EcmascriptEngine_Hermes.cpp
JoshMarler: It's actually a result of returning the lambda and associated capture block from this function. This is the only way I could get around it after much stack overlow. Tried moving the jsi::Function and didn't work. See here: https://stackoverflow.com/questions/24375342/returning-a-lambda-function-with-a-move-capture

This is a result of `juce::var::NativeFunction` being typed as `std::function` object.

react_juce/core/EcmascriptEngine_Hermes.cpp
JoshMarler: We're not at the moment but yeah I was thinking the same. Could be a nice way to pack data for something expensive you want to send over the bridge etc. Thought I'd implement as it's there. Just need to speak to hermes so we can do the same inverse operation.

react_juce/core/EcmascriptEngine_Hermes.cpp
JoshMarler: Ahh I had this and believe I hit it for a non-convertible type like jsi::Pointer or something. Will check.

react_juce/core/EcmascriptEngine_Hermes.cpp
JoshMarler: Tis this: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Function/name

JoshMarler: I think it's OK to leave empty otherwise the implementation needs to be a bit more painful ... ?

react_juce/core/EcmascriptEngine_Hermes.cpp
JoshMarler: Yeah move here should be good. Will check.

.gitmodules
nick-thompson: Nope, it will just make a relative URL against github.com

react_juce/core/EcmascriptEngine_Hermes.cpp
nick-thompson: In React Native I believe you'll see something like "[Native Method]" for the name here

JoshMarler: Right yeah. Do we want to modify to match?

react_juce/core/EcmascriptEngine_Hermes.cpp
JoshMarler: Yep agreed. It was the end user case I was concerning myself over. i.e. the user pushes some DynamicObject instance through to JS which has functions bound to some C++ object. This would be an issue if the user hasn't subclassed `juce::DynamicObject` and kept the object alive for the duration if it's usage in JS, i.e. JS stores the object for usage later. 

I think I agree though. This is probably the user's responsibility as in our case we aren't storing these objects and basically work solely on the UI thread. So in the very rare cases where we push a DynamicObject instances with methods (I don't thnk we have any of these now) we're usually invoking the JS function immediately and the DynamicObject lifetime dies right after. 

I think leave implementation as is for now and we have this comment to refer to later should needs arise.

react_juce/core/EcmascriptEngine_Hermes.cpp
JoshMarler: Thinking about it, is move actually right here? What if user has pushed a `DynamicObject` through to JS? We could be attempting to move an objects method prop? i.e. we have to do the copy somewhere. Just happensere based on the code that it's here.

JoshMarler: See here: 

```
        jsi::Object varObjectToJSIObject(juce::DynamicObject *v, jsi::Runtime &runtime)
        {
            jassert(v);

            jsi::Object jsiObj(runtime);
            for (auto &prop : v->getProperties())
            {
                jsiObj.setProperty(runtime, prop.name.getCharPointer(), varToJSIValue(prop.value, runtime));
            }

            return jsiObj;
        }
```

`varToJSIValue` in the above block will call through to `varMethodToJSIMethod` for the object props.

react_juce/core/AppHarness.cpp
```diff
@@ -18,61 +18,43 @@ namespace reactjuce
         : appRoot(_appRoot)
     {
         JUCE_ASSERT_MESSAGE_THREAD
-
-        fileWatcher = std::make_unique<FileWatcher>([this]() {
```
JoshMarler, line 22: @nick-thompson , I think using this single implementation of `handleFilesChanged` is OK? Let me know if I've aggressively blown over some edge case here which I've forgotten or not thought about. This is nice if we can keep it I think. Removes some duplication and keeps the bytecode facility fairl well encapsulated. I prefer this to a dynamic_cast.

react_juce/core/EcmascriptEngine_Hermes.cpp
JoshMarler: Fixed with latest change.

react_juce/core/EcmascriptEngine_Hermes.cpp
JoshMarler: @nick-thompson, just in case this one gets lost in the weeds. Did you want to update the function name property here to `ReactJuceNative` or something?

react_juce/core/EcmascriptEngine_Hermes.cpp
JoshMarler: Again, in case lost in the weeds. What dya reckon to this one based on above explanation @nick-thompson ?

CMakeLists.txt
JoshMarler: @nick-thompson , before I close this. Do we think a custom hermes include path is actually necessary ? I think this is only a thing we really require for Duktape. If user's are using some custom hermes checkout that's fairly non-standard and in which case I'd argue it would make more sense they had a blueprint fork and modified submodule.

JoshMarler:
@nick-thompson, Think this is probably ready for a final pass. Only thing remaining is to pragma a few warnings off which appear on Mac builds. I'll follow up tomorrow with PR for some sort of "Use Hermes" doc.

CMakeLists.txt
nick-thompson: Well, I feel like if I were in the situation trying to use blueprint but needing a custom hermes (or qjs, or duktape) and _needed_ to fork to achieve that, it would be annoying. But, AFAIK we have 0 users with that use case, so yea, good point, let's hold back any real efforts on this until it becomes apparent people want/need it

react_juce/core/EcmascriptEngine_Hermes.cpp
nick-thompson: Ahhh, right. Ok, let's avoid the move

CMakeLists.txt
JoshMarler: Yeah. OK I'll follow this PR up with another for "custom library includes". I guess my reservation is that we submodule to be locked against a particular version for API compatibility etc. etc. The same could be argued with customised Duktape though. I'd be tempted to get Hermes and QuickJS in and then add a step for all three libs where the user can (as much as possible) take the same basic configure approaches for each

react_juce/core/EcmascriptEngine_Hermes.cpp
nick-thompson: I think so, only because it will show up in stack traces and console logs if the user needs

nick-thompson (approved):
💥 🚀 

Let's do it!

react_juce/core/EcmascriptEngine_Hermes.cpp
JoshMarler: Yeah cool. Will Add. "[ReactJuce Native Method]".

nick-thompson:
Boom!

PR #233, opened 2021-02-12
Ensure node CI step is run on pull requests

JoshMarler:
@nick-thompson , Looks like I missed enabling node/bundle checks on pull requests/


nick-thompson:
🙏 thanks

PR #234, opened 2021-02-12
Fix CanvasView assertion when stateful and component has 0 width/height.

JoshMarler:
@nick-thompson, This fixes the assert people have been mentioning on Discord. 

Think this simple check is enough. I've tested this by switching the `stateful` prop on the animated Canvas element in the example plugin. Resizing of the canvas image appears to work as expected including when width/height drops to 0.


nick-thompson:
Yep, awesome, thanks

PR #235, opened 2021-02-12
Fix ts package

JoshMarler:


nick-thompson:
@JoshMarler confused about this PR, I think both of these commits are already on master, no?

JoshMarler:
Oops. Accident!

Issue #230, opened 2021-02-10
Make react-juce onMouseEnter and onMouseLeave events behave like their web counterparts


androidStern:
### The Problem
react-juce mouseEnter and mouseLeave events don't behave like their web counterparts

### Background
Juce's mouseEnter and mouseExit events differ from their browser counterparts in a few ways (detailed below). Because react-juce simply forwards Juce mouse events onto the javascript engine, the react environment has inherited Juce's behavior which is unintuitive for react developers. So, the goal here is to make the behavior in react-juce components conform to the web behaviors as much as possible.

### Behavior Differences
1. When moving from a parent to a child element
    - Juce _(and by extension, react-juce)_ will fire a mouseLeave on the Parent and a mouseEnter on the Child
    - Browsers will only fire a mouseEnter on the Child
2. When moving from a child to it's parent
    - Juce will fire a mouse leave on the Child and a mouseEnter on it's Parent
    - Browsers will only fire a mouseLeave on the child
3. When entering an exiting elements while holding the mouse button down
    - Juce will not fire any enter or exit events
    - Browsers will fire both events as described above

> TLDR for points 1 & 2: From the browser's perspective, the mouse is still inside an element if it is inside any of its children and mouse buttons have no effect.

### Extra Stuff
To get a feel for how these events behave in the browser you can check out [this page](https://www.w3schools.com/jsref/tryit.asp?filename=tryjsref_onmouseenter_html).

See [this discord thread](https://discord.com/channels/382895736356077570/763508420300374037/808348426362159115) for additional context

Finally, I haven't researched what the expected behavior is when the DOM hierarchy changes while the mouse is moving but that's something to consider as well.

nick-thompson:
Excellent report @androidStern thank you.

I'm thinking that 1 & 2 we can do our own small check [right here](https://github.com/nick-thompson/blueprint/blob/master/react_juce/core/View.cpp#L254) to say "if we're leaving but the place we're leaving to is a child component, don't fire the event" and similar for the other direction.

I think (3) will probably be a game of just finding and implementing other details of the juce::Component spec.

And (4), great catch, we can filter which types of events get bubbled [right here](https://github.com/nick-thompson/blueprint/blob/master/packages/react-juce/src/lib/Backend.ts#L279).

I'm going to leave this one here and mark it as a good first issue for a new contributor, this is a great way to dip your toes into the project

androidStern:
[...]

PR #238, opened 2021-02-16
sq

androidStern:
Draft for https://github.com/nick-thompson/blueprint/issues/230

## Things this does:
1. changes mouseEnter and mouseExit to act like browsers
2. responds to drag events at a top level component and turns them into mouseEnter and mouseExit events when necessary

## Things this doesn't do:
1. handle multitouch

BIG DISCLAIMER: I'm not a C++ guy. I'm sure the final PR will have much better code. Theres probably bugs. This is just to facilitate a design discussion.

### DIFF
--- a/react_juce/core/ReactApplicationRoot.h
+++ b/react_juce/core/ReactApplicationRoot.h
@@ -63,6 +63,29 @@ namespace reactjuce
         bool keyPressed(const juce::KeyPress& key) override;
 #endif
 
+        juce::Component* componentUnderMouse;
+
+        void mouseDrag(const juce::MouseEvent &e) override {
+            auto* newComponentUnderMouse = getComponentAt(e.x, e.y);
+
+            if (componentUnderMouse != newComponentUnderMouse){
+                // The old target may need a mouseLeave event now
+                if (componentUnderMouse != nullptr) {
+                    if (auto v = dynamic_cast<View*>(componentUnderMouse)){
+                        v->checkMouseEnter(e.getEventRelativeTo(v));
+                    }
+                }
+
+                // The new target may need a mouseEnter event now
+                if (auto v = dynamic_cast<View*>(newComponentUnderMouse)){
+                    v->checkMouseEnter(e.getEventRelativeTo(v));
+                }
+
+            }
+
+            componentUnderMouse = newComponentUnderMouse;
+        }
+
         //==============================================================================
         /** Evaluates a javascript bundle file in the EcmascriptEngine.
          *
--- a/react_juce/core/View.cpp
+++ b/react_juce/core/View.cpp
@@ -244,22 +244,59 @@ namespace reactjuce
     void View::mouseUp (const juce::MouseEvent& e)
     {
         dispatchViewEvent("onMouseUp", detail::makeViewEventObject(e, *this));
+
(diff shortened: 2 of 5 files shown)


react_juce/core/View.h
```diff
@@ -103,6 +103,7 @@ namespace reactjuce
         /** Invokes an "exported" native method on the View instance */
         juce::var invokeMethod(const juce::String &method, const juce::var::NativeFunctionArgs &args);
 
+        void checkMouseEnter(const juce::MouseEvent& e);
```
androidStern, line 106: This should really be called something like checkMouseEnterAndExit.

react_juce/core/View.cpp
```diff
@@ -248,18 +248,52 @@ namespace reactjuce
...
+            dispatchViewEvent("onMouseEnter", detail::makeViewEventObject(e, *this));
+        }
+
+        wasMouseOver = true;
     }
 
     void View::mouseExit (const juce::MouseEvent& e)
     {
-        dispatchViewEvent("onMouseLeave", detail::makeViewEventObject(e, *this));
+        if (!reallyContains(e.getPosition(), true))
```
androidStern, line 261: reallyContains returns true even if the mouse is over one of this components children.

react_juce/core/ReactApplicationRoot.h
```diff
@@ -63,6 +63,29 @@ namespace reactjuce
         bool keyPressed(const juce::KeyPress& key) override;
 #endif
 
+        juce::Component* componentUnderMouse;
+
+        void mouseDrag(const juce::MouseEvent &e) override {
```
androidStern, line 68: Used to translate mouseDrags into enter and exit events

nick-thompson: Hmm interesting, won't ReactAppRoot not get this callback if the mouse is dragged over a child component? https://docs.juce.com/master/classComponent.html#aff01e4d339bda474a66984b709044b63

Or is AppRoot getting this callback at all times?

androidStern: see View::mouseDrag I forward child drag events onto ReactAppRoot::mouseDrag.

but yes, its pretty ugly. Another way to handle it would be making ReactAppRoot a mouse listener for all its children I suppose.

nick-thompson: Got it, thanks! Finally coming back around to this, sorry for the delay.

So, reading this through again with a better understanding, I actually think this is quite like the way juce implements its own drag and drop patterns, except here we're sort of re-implementing it with ReactAppRoot == DragAndDropContainer and View == DragAndDropTarget. (See https://github.com/juce-framework/JUCE/blob/b8206e3604ebaca64779bf19f1613c373b9adf4f/modules/juce_gui_basics/mouse/juce_DragAndDropContainer.cpp)

This has me thinking then, especially in the context of the approach we laid out in the drag and drop issue (see https://github.com/nick-thompson/react-juce/issues/60#issuecomment-755713591), that perhaps we should just merge this goal into that work. The approach laid out in that issue is basically 1 step away from solving this problem here, and we can nudge the dragstart/dragmove/dragend callbacks that each View will receive (because its a DragAndDropTarget) to correctly send mouseenter/mouseleave callbacks as well.

What do you think?

PR #242, opened 2021-02-17
ShadowView Pimpl pattern.

JoshMarler:
There's a few nasty things to sort out here.
    
   @nick-thompson , Opening this up slightly ahead of time as might be a few discussion points here. Couple elements of controversy!

### DIFF
--- a/react_juce/core/ShadowView.h
+++ b/react_juce/core/ShadowView.h
@@ -1,27 +1,18 @@
-/*
-  ==============================================================================
-
-    ShadowView.h
-    Created: 17 Apr 2019 8:38:37am
-
-  ==============================================================================
-*/
-
 #pragma once
 
 #include "View.h"
 
 
-#define BP_SPREAD_SETTER_PERCENT(setter) setter, setter##Percent
-#define BP_SPREAD_SETTER_AUTO(setter) BP_SPREAD_SETTER_PERCENT(setter), setter##Auto
-
 namespace reactjuce
 {
 
-    struct BoundsAnimator : public juce::Timer {
+    //==============================================================================
+    struct BoundsAnimator : public juce::Timer
+    {
         using StepCallback = std::function<void(juce::Rectangle<float>)>;
 
-        enum class EasingType {
+        enum class EasingType
+        {
             Linear,
             QuadraticIn,
             QuadraticOut,
@@ -36,8 +27,11 @@ namespace reactjuce
         int frameRate = 45;
         EasingType easingType = EasingType::Linear;
 
-        BoundsAnimator(double durationMs, int frameRateToUse, EasingType et,
-                       juce::Rectangle<float> startRect, juce::Rectangle<float> destRect,
+        BoundsAnimator(double durationMs,
+                       int frameRateToUse,
+                       EasingType et,
+                       juce::Rectangle<float> startRect,
+                       juce::Rectangle<float> destRect,
                        StepCallback cb)
             : start(startRect)
             , dest(destRect)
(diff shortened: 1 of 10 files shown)


react_juce/core/ShadowView.h
```diff
@@ -160,147 +106,56 @@ namespace reactjuce
...
-        std::vector<ShadowView*> children;
+        //==============================================================================
 
     private:
+        class Pimpl;
+        std::unique_ptr<Pimpl> mPimpl;
+
+        //TODO: Damn. Is there a better way to do what is required here? Is this the end of the world?
+        //            Probably not in this very particular case.
+        friend Pimpl;
```
JoshMarler, line 157: @nick-thompson , This isn't wonderful but I think it's OK as completely hidden and doesn't really break encapsulation. It means that the ShadowView yoga pimpl can use the likes of `childView->mPimpl->yogaNode` in `ShadowView_Yoga.cpp`

react_juce/core/TextShadowView.h
```diff
@@ -31,44 +15,26 @@ namespace reactjuce
...
-        {
-            YGNodeMarkDirty(yogaNode);
-        }
+        void markDirty();
+
+        //==============================================================================
 
     private:
+        //==============================================================================
+        class Pimpl;
```
JoshMarler, line 35: Here's where things get a bit ugly. Shadowing here. Easy enough to fix with a rename which I'm tempted to do but can we do better?

react_juce/core/TextShadowView_Yoga.cpp
```diff
@@ -0,0 +1,49 @@
...
+        return { tl.getWidth(), tl.getHeight() };
+    }
+
+    //==============================================================================
+    class TextShadowView::Pimpl
+    {
+    public:
+        Pimpl()
+        {
+            YGNodeSetContext(yogaNode, this);
```
JoshMarler, line 37: @nick-thompson , here is a thing to be decided. How to get access to the relevant `yogaNode` instance inside the pimpl. We can pass the pimpl's owning `TextShadowView` instance through to the `Pimpl` constructor and then have this `Pimpl` again as a friend to `ShadowView/TextShadowView`. Perhaps there is a better design to use here though? 

I'm actually not totally against `friend` in these particular cases even though I think it should almost always be considered evil. What do you think?

JoshMarler:
@nick-thompson , think this is ready for a final pass after our Discord discussions. I think there's probably some opportunity to refactor around `TextShadowView` even more to remove some ugliness but not immediately apparent. It think this might be "OK for now" but require revisiting at some point. 

As discussed I've hit a need for this due to facebook namespace clash with folly function due to `YGLayout.h` having a `using namespace facebook::yoga`

react_juce/core/ShadowView.h
```diff
@@ -160,147 +106,54 @@ namespace reactjuce
...
     protected:
         //==============================================================================
-        YGNodeRef yogaNode;
-        View* view = nullptr;
-        juce::NamedValueSet props;
+        std::vector<ShadowView*>& getChildren();
 
-        std::unique_ptr<BoundsAnimator> animator;
-        std::vector<ShadowView*> children;
+        class ShadowViewPimpl;
```
nick-thompson, line 149: I think these things can be private, right? Because the only references are within the ShadowView_Yoga.cpp implementation? Or does the TextView need to reference them as well?

nick-thompson (approved):
Awesome @JoshMarler LGTM.

Left one question inline; if it's something we can't or shouldn't change then feel free to ignore. Either way, :+1: from me to go ahead with a merge

react_juce/core/ShadowView.h
JoshMarler: Yeah this is the one bit I'm not particularly happy with right now. `TextShadowView` needs these at the moment. I think a better thing might emerge but I'm not sure what that is right now with the current shape of `TextShadowView`.

PR #249, opened 2021-03-17
Optimize the gain plugin example

szk2s:
I did some refactoring and optimization on the react side.
Because my purpose was a performance test mainly, this is not a thorough refactor. 
I left some components with the older class component, which causes some inconsistency in the coding style. 

- rewrite some components with functional component using Hooks
- extract knob components because the update of mute state causes irrelevant rerender of ParameterSlider.

### DIFF
--- a/examples/GainPlugin/jsui/src/ParameterSlider.js
+++ b/examples/GainPlugin/jsui/src/ParameterSlider.js
@@ -1,25 +1,30 @@
 import React, { memo, useCallback } from "react";
 import { Slider } from "react-juce";
+import {
+  beginParameterChangeGesture,
+  endParameterChangeGesture,
+  setParameterValueNotifyingHost,
+} from "./nativeMethods";
 
 const ParameterSlider = ({ value, paramId, children, ...props }) => {
   console.log("Slider rendered");
   const onMouseDown = useCallback(
     (e) => {
-      global.beginParameterChangeGesture(paramId);
+      beginParameterChangeGesture(paramId);
     },
     [paramId]
   );
 
   const onMouseUp = useCallback(
     (e) => {
-      global.endParameterChangeGesture(paramId);
+      endParameterChangeGesture(paramId);
     },
     [paramId]
   );
 
   const onSliderValueChange = useCallback(
     (value) => {
-      global.setParameterValueNotifyingHost(paramId, value);
+      setParameterValueNotifyingHost(paramId, value);
     },
     [paramId]
   );
--- a/examples/GainPlugin/jsui/src/ParameterSlider.js
+++ b/examples/GainPlugin/jsui/src/ParameterSlider.js
@@ -7,26 +7,17 @@ import {
 } from "./nativeMethods";
 
 const ParameterSlider = ({ value, paramId, children, ...props }) => {
-  const onMouseDown = useCallback(
-    (e) => {
-      beginParameterChangeGesture(paramId);
-    },
-    [paramId]
-  );
+  const onMouseDown = (e) => {
+    beginParameterChangeGesture(paramId);
+  };
 
-  const onMouseUp = useCallback(
-    (e) => {
-      endParameterChangeGesture(paramId);
-    },
-    [paramId]
-  );
+  const onMouseUp = (e) => {
+    endParameterChangeGesture(paramId);
+  };
 
-  const onSliderValueChange = useCallback(
-    (value) => {
-      setParameterValueNotifyingHost(paramId, value);
-    },
-    [paramId]
-  );
+  const onSliderValueChange = (value) => {
+    setParameterValueNotifyingHost(paramId, value);
+  };
 
   return (
     <Slider
--- a/examples/GainPlugin/jsui/src/ParameterSlider.js
+++ b/examples/GainPlugin/jsui/src/ParameterSlider.js
@@ -7,7 +7,6 @@ import {
 } from "./nativeMethods";
 
 const ParameterSlider = ({ value, paramId, children, ...props }) => {
-  console.log("Slider rendered");
   const onMouseDown = useCallback(
     (e) => {
       beginParameterChangeGesture(paramId);
--- a/examples/GainPlugin/jsui/src/ParameterSlider.js
+++ b/examples/GainPlugin/jsui/src/ParameterSlider.js
@@ -59,5 +59,6 @@ const ParameterSlider = ({ paramId, children, ...props }) => {
 
 // TODO: PropTypes Validation
 // paramId should be required and has type of string
+// https://www.npmjs.com/package/prop-types
 
 export default memo(ParameterSlider);
--- a/examples/GainPlugin/jsui/src/ParameterSlider.js
+++ b/examples/GainPlugin/jsui/src/ParameterSlider.js
@@ -1,28 +1,8 @@
-import React, { memo, useEffect, useCallback, useState } from "react";
-import { EventBridge, Slider } from "react-juce";
+import React, { memo, useCallback } from "react";
+import { Slider } from "react-juce";
 
-const ParameterSlider = ({ paramId, children, ...props }) => {
-  const [value, setValue] = useState(0.0);
-
-  const onParameterValueChange = useCallback(
-    (index, changedParamId, defaultValue, currentValue) => {
-      if (changedParamId === paramId) {
-        setValue(currentValue);
-      }
-    },
-    [setValue]
-  );
-
-  useEffect(() => {
-    EventBridge.addListener("parameterValueChange", onParameterValueChange);
-    return () => {
-      EventBridge.removeListener(
-        "parameterValueChange",
-        onParameterValueChange
-      );
-    };
-  }, [onParameterValueChange]);
-
+const ParameterSlider = ({ value, paramId, children, ...props }) => {
+  console.log("Slider rendered");
   const onMouseDown = useCallback(
     (e) => {
       global.beginParameterChangeGesture(paramId);
--- a/examples/GainPlugin/jsui/src/ParameterSlider.js
+++ b/examples/GainPlugin/jsui/src/ParameterSlider.js
@@ -1,90 +1,63 @@
-import React, { Component } from "react";
-import { Slider } from "react-juce";
+import React, { memo, useEffect, useCallback, useState } from "react";
+import { EventBridge, Slider } from "react-juce";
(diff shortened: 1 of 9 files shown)


szk2s:
As [Nick's suggestion](https://github.com/nick-thompson/react-juce/discussions/250#discussioncomment-495182),  there was a performance issue where each ParameterSlider listens to the event and then `setState`.
So I did some optimizations with Context API (see `e20d5b8`), and did some tests with 50 knob copies. 
As a result, knobs move more smoothly as we can see in the GIF. 
But it didn't show up in the profiler: the CPU usage is always 100%.

- knob-test-1
[Image from Gyazo](https://gyazo.com/513d261b3924e995c3b0e056a1e298db)

- knob-test-2
[Image from Gyazo](https://gyazo.com/a41e3c3f08c2032e50b6d17631d84dc6)

szk2s:
For curious person, the test codes can be found at:
https://github.com/magnetica-studio/react-juce/tree/knob-test-1
https://github.com/magnetica-studio/react-juce/tree/knob-test-2

And the log file just in case. [knob-test.trace.zip](https://github.com/nick-thompson/react-juce/files/6161575/knob-test.trace.zip)

examples/GainPlugin/jsui/src/ParameterSlider.js
```diff
@@ -1,90 +1,39 @@
...
+      {children}
+    </Slider>
+  );
+};
+
+// TODO: PropTypes Validation
+// paramId should be required and has type of string
+// https://www.npmjs.com/package/prop-types
+
+export default memo(ParameterSlider);
```
olidacombe, line 39: 👍🏾

JoshMarler:
@szk2s looks like some solid wins and practices to follow here. FYI I've also just patched a performance disaster when laying out canvas components: 

https://github.com/nick-thompson/react-juce/pull/251

A minor thing I had missed. We tested whacking 10 sliders in the gain plugin example and then resizing the app window and performance fell off of a cliff. The above PR resolves. Combined with your PR we should see sizeable improvement.

nick-thompson:
💥 🚀 

Sorry @szk2s for the delay, but yea @JoshMarler found a huge win in here earlier today. I've got your branch now rebased against his fix and I have 10 knobs rendering super smoothly with ~50% CPU during heavy drag, and 50 knobs rendering smoothly with ~85% CPU during heavy drag.

On my profiles it seems that the heavy lift comes during the dispatching of param values into javascript, which is good news because that means our canvas rendering is quite good ;). Now, to me this actually makes a fair bit of sense because I believe with the context provider at the root of the app, we're doing a lot of re-render in response to new values. See here: https://reactjs.org/docs/context.html#contextprovider

> All consumers that are descendants of a Provider will re-render whenever the Provider’s value prop changes. The propagation from Provider to its descendant consumers (including .contextType and useContext) is not subject to the shouldComponentUpdate method, so the consumer is updated even when an ancestor component skips an update.

So I think with context we'll want to focus the provider's scope to only include the knob components, and then we'll want to get very particular with how the knobs handle a state update to avoid reconciliation as much as possible. I've got some ideas here, and will try to look into it some more tomorrow!

szk2s:
Thanks, @JoshMarler and @nick-thompson.
When the parameter change, I think reconciliation is triggered only once because the state change is only in the provider. But I don't know how to make sure it easily as it is hidden in the React inner implementation. Any ideas?

nick-thompson:
Ok, getting back to this today and, wow, we have some big wins to make here :)

> When the parameter change, I think reconciliation is triggered only once because the state change is only in the provider. But I don't know how to make sure it easily as it is hidden in the React inner implementation. Any ideas?

Yes, you're totally right. So, by putting the context up at the top we do guarantee batching, which is helpful, but it comes at the cost of forcing a full-tree reconciliation, even skipping any shouldComponentUpdate operations. I could imagine this being a helpful pattern for some cases, but for something like a parameter which is changing so frequently, a full-tree reconciliation at that rate might be expensive.

I'm exploring an alternative pattern where, at the beginning of a parameter change, I set a value in the ParameterValueStore to say "this value is in a change gesture." All knobs which aim to render that parameter see the state update and mark their respective canvas child into `animate={true}` mode, which means that without any reactjs intervention, the knob canvases will get repainted at 45Hz. This is neat, because then I don't actually have to involve any react state or tree diffing in this frequent change loop. With this pattern, I've got 50 sliders rendering smoothly with 50% CPU during heavy drag, and hermes doesn't even show up in the profile ;) At that point the work that remains is optimizing exactly when and how we paint, and optimizing our juce paint internals.

Now, there's more here: the tree renders are actually way more expensive than we want them to be, and in part it's because we're doing some naive things in our view implementation. In particular, not throttling our repaints on setProperty calls. I have a change locally for marking a view dirty after a setProperty call instead of calling `repaint()`. Then during the `commitUpdate` phase we do a tree traversal and call `repaint(area)` for any dirty child. I think there's more to do here, but at least this means that we only paint a child once for each batch of property changes.

They're also expensive because we do things like run the react loop, batch our layout into the commit update phase, then dispatch `measure` events back into js where we then re-run the react loop because we're not yet using `shouldComponentUpdate` everywhere, and then repeat.

So, long story short, I think you've actually cracked open a great can of worms @szk2s and I very much appreciate it! Showing me we've got quite a bit of low hanging fruit here that I think can drastically (drastically!) improve our performance.

If you don't mind, I'm going to hold off on merging here a little bit longer because I'm not yet convinced that using Context is our best option. I want to dig a little deeper on it still while I pick off some of the unrelated low hanging fruit. Besides that, I quite like the way you've refactored much of this code so I do want to merge at least several parts of it.

szk2s:
Awesome! Adding `animate=true` sounds a big step to specialize React in audio development. In this field, dragging is much more used than in the web dev, and more performance is required.

Passing props from parent to child can be cumbersome when they are required by many components within an application. Context API only solves this problem. I think this gain example can be written without Context API to keep the state at the top and pass the data as props. This is not so cumbersome in this small example. Context API is powerful in general, but not a minimal solution in this case. I can add some further refactors on top of this PR if this plugin is to show the minimal solution instead of an opinionated solution that might be helpful for the library users.

nick-thompson:
> Awesome! Adding `animate=true` sounds a big step to specialize React in audio development. In this field, dragging is much more used than in the web dev, and more performance is required.

Totally :) `animate=true` is actually already supported on `<Canvas>`, it's just not well-documented yet. (So much to do, so little time!)

> Passing props from parent to child can be cumbersome when they are required by many components within an application. Context API only solves this problem. I think this gain example can be written without Context API to keep the state at the top and pass the data as props.

Yea, I agree 100%. I think the Context API is quite elegant for those scenarios too. The only reason I'm suspicious of it in this case is because during a parameter value change we'll post new param values into js at 30Hz (via the GenericEditor's event dispatching), and that means doing a _full_ tree reconciliation at 30Hz during the drag gesture. I could imagine here using Context for values which change less frequently and then using a simple store strictly for the paramter value. I'm going to keep digging in here before drawing any conclusions though, I think we're uncovering a solid set of performance gains!

szk2s:
OK! So, I will wait for your conclusion. Let's hold merging this a little bit longer. 🙂

nick-thompson:
Ok, back with some more updates:
* I think there are some internal optimizations to make, as I mentioned above, but I don't want to hold this PR up on that because I think clearly there are some wins in this PR that make sense to merge as soon as we can.
* In my measurements it looks clear that using the Context API helps React batch the whole reconcile into one pass, and after adding some additional `shouldComponentUpdate`-style fixes to our internal Slider and Canvas I've got 50 sliders rendering smoothly with ~70-75% CPU.
* I got maybe 5% CPU improvement by moving the `<ParameterValueProvider>` to surround just the knobs instead of the whole app, which prevented unnecessary reconciliation
* Of that ~70-75% CPU I measure running javascript as ~50% of the work, and painting the juce components ~35%
* If I change this instead to not use canvas, and to not pass parameter values as props, but rather set each slider canvas to animate and during `onDraw` fetch the latest value from the ParameterValueStore, I've got 50 sliders rendering smoothly with ~60-65% CPU, of which 82% is painting the juce components, and 2.8% is running javascript.
* For me, this shows just how quickly running the react reconcile step can add up in terms of performance, which confirms my hesitation about Context for quickly-changing values.

So, what I'd like to do here is this: if you're both cool with it @szk2s @JoshMarler I'd like to go ahead and merge this to start, because as-is this does bring performance improvements to the gain example and I think the code here is just much nicer than what I had written :). Then I think we should follow up with some experimental PRs comparing alternative ways to squeeze out more performance improvements (I'll start by opening a PR showing my animate trick), while at the same time tackling some of the internal optimizations that we discovered here (in particular, optimizing our paint strategy and, if we can, our canvas paint procedures)

What do you think?

JoshMarler:
Super interesting @nick-thompson. This all sounds great. I wonder if the shift in the canvas draw commands approach had anything to do with push of work into JS https://github.com/nick-thompson/react-juce/commit/9103187dd3b0a91326aaae337ae064de84be2ad1. Definitely keen to look into optimising canvas. Would be happy to take a look into some of this.

szk2s:
OK, sounds cool.
If you move the `<ParameterValueProvider>` to surround just the knobs, I think you won't need Context API. `useState` should be enough. 
It can be discussed in another PR. Go ahead!😄

nick-thompson:
> If you move the <ParameterValueProvider> to surround just the knobs, I think you won't need Context API. useState should be enough.

Ah, of course :D I totally missed that.

> This all sounds great. I wonder if the shift in the canvas draw commands approach had anything to do with push of work into JS 9103187. Definitely keen to look into optimising canvas. Would be happy to take a look into some of this.

It was actually pretty awesome to see– painting those 50 sliders at 30Hz barely cost anything in terms of javascript time. Assembling those draw commands arrays seems quite cheap. It's actually processing them and painting them on the juce side that gets expensive. Much to dig into there!

Ok awesome, let's go 🚀 . @szk2s do you mind if I squash this into 1 commit on the way in?

JoshMarler:
> actually

Ah OK nice. I might profile the Canvas View tonight and see if I can pull out any cpp gems.

szk2s:
> Ok awesome, let's go 🚀 . @szk2s do you mind if I squash this into 1 commit on the way in?

Sure! No problem.

nick-thompson:
Thank you @szk2s!

PR #251, opened 2021-03-24
Fix a bug with canvas resize performance in the standard non-stateful

JoshMarler:
case.

    We should probably investigate our "stateful" canvas component after
    this to ensure we don't have any giant performance holes. Worth
    noting that juce::Image::rescaled is a pretty slow thing should we
    rely on it again.

### DIFF
--- a/react_juce/core/CanvasView.cpp
+++ b/react_juce/core/CanvasView.cpp
@@ -611,6 +611,11 @@ namespace reactjuce
             if (!shouldAnimate && isTimerRunning())
                 stopTimer();
         }
+        
+        if (name == statefulProp && props[statefulProp])
+        {
+            resized();
+        }
     }
 
     //==============================================================================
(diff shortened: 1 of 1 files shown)


react_juce/core/CanvasView.cpp
```diff
@@ -611,6 +611,11 @@ namespace reactjuce
             if (!shouldAnimate && isTimerRunning())
                 stopTimer();
         }
+        
+        if (name == statefulProp)
```
nick-thompson, line 615: Kind of nit-picky but I think we might want `if (name == statefulProp && (bool) value)` It's possible, though unlikely, that someone would write `<Canvas stateful={false}>` in their js

react_juce/core/CanvasView.cpp
```diff
@@ -652,15 +657,18 @@ namespace reactjuce
     {
         View::resized();
 
-        //TODO: Fix image scalling for retina displays.
-        //      May require passing an optional scaleFactor
-        //      arg through to draw commands.
-        const auto bounds = getLocalBounds();
-        if (bounds.getWidth() > 0 && bounds.getHeight() > 0)
+        if (props.contains(statefulProp))
```
nick-thompson, line 660: Similar deal here, `if (props.contains(statefulProp) && (bool) props[statefulProp])`

nick-thompson (approved):
This is a huuuuuuge win, wow. I'm surprised we missed it earlier 😄 

Thanks so much for the find. I'm ready to merge asap, just two kind of nitpicks inline

JoshMarler:
@nick-thompson, rebased and pushed up your suggestions. Think this should be good to go

JoshMarler:
FYI performance in our test case under a debug build is still pretty rough. I'll add an issue for this.

nick-thompson:
Thanks @JoshMarler! Yea, still plenty of work to do ;) But this is a huge win already

PR #252, opened 2021-04-03
Catch unhandled errors in EcmascriptEngine_Hermes

JoshMarler:
@nick-thompson hopefully this should catch the errors you've been seeing under hermes and trap under the usual red screen


nick-thompson:
Awesome, thanks @JoshMarler

PR #254, opened 2021-04-07
Made CMakeLists.txt CPM-compatible

sdeleon28:
Implemented the fix detailed in https://github.com/nick-thompson/react-juce/issues/253

In summary this PR:

* Makes it possible to include react-juce with CPM
* Makes it possible to modify `REACTJUCE_JS_LIBRARY` from client code
* Makes building JUCE optional
* Makes building the example optional
* Documents all of these changes

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -5,12 +5,17 @@ project(reactjuce VERSION 0.1.0)
 # (practical to debug the plugin processor directly from your IDE)
 option(JUCE_BUILD_EXTRAS "Build JUCE Extras" OFF)
 
-
 # Change this option to set the JS Interpreter/Engine you wish to run React-JUCE against.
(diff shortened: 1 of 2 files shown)


sdeleon28:
I'll fix that linting error in a sec

JoshMarler:
@sdeleon28 , apologies for the delay on this one. In crunch time at current shop. I'll do my utmost to take a first pass over this tonight!

mjmaurer:
Any updates here? Would love to have this merged :)

nick-thompson:
I'm for it, but we had originally held off on merging this because we had come up with a different way that we wanted to structure our CMake here. In particular, I remember @JoshMarler suggesting that the root level project shouldn't add juce, that we can delegate that to whomever calls `add_subdirectory(react_juce)`, and therefore we can move the juce inclusion here into the example project. I think that structure would clean things up a bit still, so I think I'd prefer that. Am I missing anything there @JoshMarler? Was there more to it?

PR #258, opened 2021-04-14
Fix mac build with strict warnings

JoshMarler:
@nick-thompson, this is rearing it's head with our warning settings.  Complains when you have no previous declaration of the non-static function inside the react_juce namespace. Moving into anonymous namespace resolves.


nick-thompson:
:+1: :D

PR #259, opened 2021-04-24
Fix #255: Passing a file URL to a .svg to ImageView renders nothing

### LINKED ISSUE #255: Passing a file URL to a .svg to ImageView renders nothing
I'm finding passing a file URL to a .svg as `ImageView`'s `source` prop results in nothing being rendered. 

I suspect we need to handle `.svg` file extensions explicitly and create a drawable from the files contents as a string or simlar.

JoshMarler:


nick-thompson:
Perfect, thanks @JoshMarler merging!

PR #272, opened 2021-07-10
Topic/stylesheet api

JoshMarler:
@nick-thompson , Opening this up as I go along so we can use it as discussion vehicle rather than me spamming you to death on Discord. 

I'll pop a few comments inline with questions. Overall everything is working but obviously this needs a bit of tidying. 

We'll get an A/B timing diff pre/post StyleSheet for initial app render to see how big a win this is. Suspect it will be pretty significant. 

Perhaps we also shove a print in `performRootShadowTreeLayout` and friends to see how much this reduces calls to that hot spot.

### DIFF
--- a/packages/react-juce/src/lib/Stylesheet.ts
+++ b/packages/react-juce/src/lib/Stylesheet.ts
@@ -13,9 +13,7 @@ const cssPropsMap = allCssProps
   .filter((s) => !s.startsWith("-") && s.includes("-"))
   .reduce((acc, v) => Object.assign(acc, { [camelCase(v)]: v }), {});
 
-//TODO: Add interface and type parameter of create function.
-//TODO: Add compose(), flatten().
-//TODO: Style prop on View must allow for an array of styles objects
+//TODO: Add interface types
 export default {
     create(stylesObj) {
         // TODO: Will be another instance of our
@@ -32,7 +30,7 @@ export default {
                 // of a css prop, first convert it to kebab-case.
                 propKey = cssPropsMap[propKey] || propKey;
 
-                // convert provided color string to alpha-hex code for JUCE
+                // Convert provided color string to alpha-hex code for JUCE.
                 if (Colors.isColorProperty(propKey)) {
                   value = Colors.colorStringToAlphaHex(value);
                   if (value.startsWith("linear-gradient")) {
@@ -47,7 +45,6 @@ export default {
                 }
 
                 nativeSubObj[propKey] = value;
-
             }
 
             nativeStyles[subObj] = nativeSubObj;
@@ -55,5 +52,11 @@ export default {
 
         //@ts-ignore
         return NativeMethods.createStyleSheet(nativeStyles);
+    },
+    cx(...args) {
+        return args.map((a) => {
+            if (a !== false)
+                return a;
+        })
     }
 }
--- a/packages/react-juce/src/lib/Stylesheet.ts
+++ b/packages/react-juce/src/lib/Stylesheet.ts
@@ -1,0 +1,59 @@
+import NativeMethods from './NativeMethods';
+
+import { all as allCssProps } from "known-css-properties";
+import camelCase from "camelcase";
+
+import Colors from "./MacroProperties/Colors";
+import { macroPropertyGetters } from "./MacroProperties";
+
+// Get any css properties not beginning with a "-",
+// and build a map from any camelCase versions to
+// the hyphenated version.
+const cssPropsMap = allCssProps
+  .filter((s) => !s.startsWith("-") && s.includes("-"))
+  .reduce((acc, v) => Object.assign(acc, { [camelCase(v)]: v }), {});
+
+//TODO: Add interface and type parameter of create function.
+//TODO: Add compose(), flatten().
+//TODO: Style prop on View must allow for an array of styles objects
+export default {
+    create(stylesObj) {
+        // TODO: Will be another instance of our
+        //       styles object type.
+        let nativeStyles = {};
+
+        for (const subObj in stylesObj) {
+            let nativeSubObj = {};
+
+            for (let propKey in stylesObj[subObj]) {
+                let value = stylesObj[subObj][propKey];
+
+                // If the supplied propkey is a camelCase equivalent
+                // of a css prop, first convert it to kebab-case.
+                propKey = cssPropsMap[propKey] || propKey;
+
+                // convert provided color string to alpha-hex code for JUCE
+                if (Colors.isColorProperty(propKey)) {
+                  value = Colors.colorStringToAlphaHex(value);
+                  if (value.startsWith("linear-gradient")) {
+                    value = Colors.convertLinearGradientStringToNativeObject(value);
+                  }
+                }
+                else if (macroPropertyGetters.hasOwnProperty(propKey)) {
+                    for (const [k, v] of macroPropertyGetters[propKey](value)) {
+                        nativeSubObj[k] = v;
+                    }
+                  return;
+                }
+
+                nativeSubObj[propKey] = value;
+
+            }
+
+            nativeStyles[subObj] = nativeSubObj;
+        }
+
+        //@ts-ignore
+        return NativeMethods.createStyleSheet(nativeStyles);
+    }
+}
--- a/react_juce/core/View.cpp
+++ b/react_juce/core/View.cpp
@@ -191,6 +191,10 @@ namespace reactjuce
             auto c = juce::Colour::fromString(colorString);
             float borderWidth = props[borderWidthProp];
 
+            auto myAreaRect = getLocalBounds();
+            auto xPos = myAreaRect.getX();
+            auto yPos = myAreaRect.getY();
+
             // Note this little bounds trick. When a Path is stroked, the line width extends
             // outwards in both directions from the coordinate line. If the coordinate
             // line is the exact bounding box then the component clipping makes the corners
--- a/react_juce/core/ViewManager.cpp
+++ b/react_juce/core/ViewManager.cpp
@@ -87,12 +87,9 @@ namespace reactjuce
     {
         const auto& [view, shadow] = getViewHandle(viewId);
 
-        if (name == "styles")
+        auto parseStyleObj = [=] (const juce::var &obj)
         {
-            const StyleSheetID id = value;
-
-            std::cout << "set styles using id: " << id << std::endl;
-
+            const StyleSheetID id = obj;
             auto it = styleSheets.find(id);
 
             if (it != styleSheets.end())
@@ -104,9 +101,25 @@ namespace reactjuce
                     //TODO: We'll likely change this to some internal
                     //      setStyleProperty on View which isn't directly
                     //      exposed to our react renderer implemenation.
-                    std::cout << "setViewProperty: " << nv.name.toString() << " : " << nv.value.toString() << std::endl;
                     setViewProperty(viewId, nv.name.toString(), nv.value);
                 }
+            }
+        };
+
+        if (name == "styles")
+        {
+            if (value.isArray())
+            {
+                auto *arr = value.getArray();
+
+                for (juce::var styleSheetObj : *arr)
+                {
+                    parseStyleObj(styleSheetObj);
+                }
+            }
+            else
+            {
+                parseStyleObj(value);
             }
         }
         else
--- a/react_juce/core/ViewManager.cpp
+++ b/react_juce/core/ViewManager.cpp
@@ -113,9 +113,7 @@ namespace reactjuce
                 auto *arr = value.getArray();
 
                 for (juce::var styleSheetObj : *arr)
-                {
                     parseStyleObj(styleSheetObj);
-                }
             }
             else
             {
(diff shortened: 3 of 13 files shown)


packages/react-juce/src/lib/Stylesheet.ts
```diff
@@ -0,0 +1,62 @@
...
+  .reduce((acc, v) => Object.assign(acc, { [camelCase(v)]: v }), {});
+
+//TODO: Add interface types
+export default {
+    create(stylesObj) {
+        // TODO: Will be another instance of our
+        //       styles object type.
+        let nativeStyles = {};
+
+        for (const subObj in stylesObj) {
```
JoshMarler, line 23: @nick-thompson , I suspect this block can be tidied further once I've added types etc. but functionally this is now doing the right thing.

packages/react-juce/src/lib/Stylesheet.ts
```diff
@@ -0,0 +1,62 @@
...
+                nativeSubObj[propKey] = value;
+            }
+
+            nativeStyles[subObj] = nativeSubObj;
+        }
+
+        //@ts-ignore
+        return NativeMethods.createStyleSheet(nativeStyles);
+    },
+    cx(...args) {
```
JoshMarler, line 56: @nick-thompson , I think you had a more elegant thing in mind for this little `cx` helper function when we want to conditionally apply a group of styles. See it's usage in `App.js`. Would you mind elaborating on what you had in mind?

react_juce/core/View.cpp
```diff
@@ -191,6 +191,10 @@ namespace reactjuce
             auto c = juce::Colour::fromString(colorString);
             float borderWidth = props[borderWidthProp];
 
+            auto myAreaRect = getLocalBounds();
```
JoshMarler, line 194: @nick-thompson , This is a mistake left over from debugging a thing. Will remove.

react_juce/core/ViewManager.cpp
```diff
@@ -87,11 +87,47 @@ namespace reactjuce
...
+            if (it != styleSheets.end())
+            {
+                juce::NamedValueSet &style = it->second;
+
+                for (auto &nv : style)
+                {
+                    //TODO: We'll likely change this to some internal
+                    //      setStyleProperty on View which isn't directly
+                    //      exposed to our react renderer implemenation.
+                    setViewProperty(viewId, nv.name.toString(), nv.value);
```
JoshMarler, line 104: @nick-thompson, so here is the main thing to decide on in this MR I think. To get things moving and test the overall approach I simply unpack the stored stylesheet props and call `setViewProperty` recursively. 

We don't want to support the old style props on `View` anymore and instead only want to support passing `styles` from React. So, how should we approach this?

Ideas, 

We have a `View::setStyleProperty` function which we call into here. `View::setStyleProperty` will only ever be called internally and never by the renderer/React. This would work fine in theory, then in `View::setProperty` we only handle the properties we expect/accept. We could also combine this with your earlier idea of filtering out unsupported properties in `Backend.ts`'s `setProperty`. 

Any alternative ideas/thoughts. Hopefully we can pair this up with actually typing all supported `ViewProps` etc. I'm happy to do that as a separate MR or just pack onto the top of this one.

JoshMarler:
@nick-thompson , Will fix linter checks before this goes in. Think I've got vim at 4 space tabs etc. Will trying running linter or edit.

react_juce/core/ViewManager.cpp
nick-thompson: Yea good question.

So, for one, I don't think we need to worry about any recursive behavior here. It's fine for style properties to deal with nested objects, like our `border` conversation. The user won't work with a nested object, but our macros will expand out into a border object. I think probably the only reason we might need recursive application here is just because thats how our current view painting expects it. But, we should just change that.

Second, I actually don't think the View even needs to hold these properties. It's fine to just hold a single `style` property in the `View` itself, then at paint time lookup the ViewManager and read the relevant styles from the ViewManager's map. That will prevent potentially a lot of value copying.

And then third, yea, I think filtering (and logging a warning message) at the js layer is definitely high on the priority list.

That make sense? This can also work out such that we can merge this without introducing a hgue breaking change: in our `View` paint routine, we check: do I have a `style` property? If so, lookup my styles from the ViewManager, and paint with them, then return early. If not, fall back to the current paint routine to support those properties (and perhaps log a deprecation warning). Then in a future PR we can remove support for the current way of doing things. It's a bit unclear to me how careful we have to be with backwards compat... I suspect we're still within a window of time where a breaking change like this won't affect too many people too drastically, so maybe I'm overthinking

react_juce/core/ViewManager.cpp
```diff
@@ -235,6 +271,27 @@ namespace reactjuce
...
+
+        juce::DynamicObject::Ptr res = new juce::DynamicObject();
+
+        auto obj = value.getDynamicObject();
+
+        for (auto &nv : obj->getProperties())
+        {
+            jassert(nv.value.isObject());
+
+            const StyleSheetID id = juce::DefaultHashFunctions::generateHash(juce::Uuid(), INT_MAX);
```
nick-thompson, line 286: What if we hashed the actual property set as the ID? That way if a user accidentally (or intentionally) built two style sheet objects that were identical to one another, our map would coalesce them and save some space

nick-thompson (commented):
Excellent work @JoshMarler this looks great. Left some discussion inline!

packages/react-juce/src/lib/Stylesheet.ts
```diff
@@ -0,0 +1,177 @@
...
+
+/**
+ *
+ */
+interface FontStyles {
+    color?: string,
+    fontSize?: number,
+    fontStyle?: string,
+    fontFamily?: string,
+    justification?: number, //TODO: How to actually type this? Use enum?
```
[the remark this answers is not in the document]
JoshMarler: Cool for me to just add the TODO for now and create associated issue?

packages/react-juce/src/lib/Stylesheet.ts
```diff
@@ -0,0 +1,177 @@
...
+    kerningFactor?: string,
+    lineSpacing?: number,
+    wordWrap?: number, //TODO: How to actually type this? Use enum?
```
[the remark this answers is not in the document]
JoshMarler: As before, good for me to alter TODO and created associated issue?

packages/react-juce/src/lib/Stylesheet.ts
```diff
@@ -0,0 +1,177 @@
...
+}
+
+/**
+ *
+ */
+interface ImageStyles {
+    placement?: number //TODO: How to actually type this? Use enum?
+}
+
+// TODO: How to add LayoutAnimated types. Do they belong in StyleSheet
```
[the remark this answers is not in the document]
JoshMarler: Cool, so we make this part of ViewProps ? In which case, we no longer support user specifying this as part of their stylesheet objects ?

packages/react-juce/src/lib/Stylesheet.ts
```diff
@@ -0,0 +1,177 @@
...
+                nativeSubObj[propKey] = value;
+            }
+
+            nativeStyles[subObj] = nativeSubObj;
+        }
+
+        //@ts-ignore
+        return NativeMethods.createStyleSheet(nativeStyles);
+    },
+    cx(...args: StyleSheetIdentifierSwitch[]) {
```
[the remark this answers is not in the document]
JoshMarler: Excellent thanks. I'll give this a go.

react_juce/core/ViewManager.cpp
JoshMarler: Riiiight ok. Approach makes total sense. So are we going with supporting the "deprecated" approach for now? Or do we make the breaking change? Think we have issue for filtering props. That should be able to be dealt with as a follow up to this I think. I'm making this ticket, full prop types and then props filtering my priorities (in that order).

react_juce/core/ViewManager.cpp
JoshMarler: Yeah doesn't seem a bad idea. Moderate concerns around collisions and performance after some recent findings. Could do in this PR or as enhancement/performance follow up ?

react_juce/core/ViewManager.cpp
nick-thompson: > I'm making this ticket, full prop types and then props filtering my priorities (in that order).

Sounds good!

> So are we going with supporting the "deprecated" approach for now? Or do we make the breaking change?

Let's do the easy version of deprecating: 
* In View::paint, if we have a style prop we look up our style props via the ViewManager and paint according to those, then return
* Else, we DBG("Deprecation warning: painting from top-level properties"); and then carry on with the code that's already written there

In a future PR we'll do prop filtering at the js level which will make the latter branch obsolete and we can remove it then

packages/react-juce/src/lib/Stylesheet.ts
```diff
@@ -0,0 +1,177 @@
...
+    right?: number | string,
+    top?: number | string,
+    width?: number | string,
+}
+
+/**
+ *
+ */
+interface ImageStyles {
+    placement?: number //TODO: How to actually type this? Use enum?
```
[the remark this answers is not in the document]
JoshMarler: Righty. So leave where it is for the moment and create ticket or move this up to `ImageProps` ?

react_juce/core/ViewManager.cpp
JoshMarler: @nick-thompson , just coming back to this. So quick question, if we have View holding a single `style` prop which back references to the style object map in `ViewManager` how do you envisage we should set `ShadowView` properties ? At the moment we have the below approach for this:

```
            if(!shadow->setProperty(name, value)) {
              view->setProperty(name, value);
              view->repaint();
            }
```

nick-thompson: Yea, I think they would need to. We could try to differentiate between layout props and others, but I think it's probably fine for both the view and its shadow to have a style prop

JoshMarler: Yep cool. I'll take that approach then. OK cheers. Try to get this finished off tomorrow.

packages/react-juce/src/lib/Stylesheet.ts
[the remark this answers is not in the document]
JoshMarler: Lol apologies as picking this up after leaving it stale for a long time. 

This is cool the only debatable issue with this approach for the `cx` function is you have to write something like this:

```
  mute_button: {
    justifyContent: "center",
    alignItems: "center",
    borderRadius: 5.0,
    borderWidth: 2.0,
    borderColor: "rgba(102, 253, 207, 1)",
    marginTop: 10,
    minWidth: 30.0,
    minHeight: 30.0,
    width: "20%",
    height: "17.5%",
  },
  mute_button_active: {
    backgroundColor: "hsla(162, 97%, 70%, 0)"
  },
  mute_button_inactive: {
    backgroundColor: "#66FDCF"
  },
  mute_button_text: {
    fontSize: 20.0,
    lineSpacing: 1.6,
    fontStyle: Text.FontStyleFlags.bold,
  },
  mute_button_text_inactive: {
      color: "#17191f"
  },
  mute_button_text_active: {
      color: "hsla(162, 97%, 70%, 1)"
  },
  
      const [muted, setMuted] = useState(false);

    const muteButtonStyles = StyleSheet.cx({[styles.mute_button]: true,
                                            [styles.mute_button_active]: muted,
                                            [styles.mute_button_inactive]: !muted});

    const muteTextStyles = StyleSheet.cx({[styles.mute_button_text]: true,
                                          [styles.mute_button_text_active]: muted,
                                          [styles.mute_button_text_inactive]: !muted});

```

In practice I'm not sure if it's a major issue but you can't rely on doing say:

```
    // Here we want to rely on positional override so that mute_button_text_active
    // overrides the default/base color in mute_button_text
    const muteTextStyles = StyleSheet.cx({[styles.mute_button_text]: true,
                                          [styles.mute_button_text_active]: muted});
```

Because you can't guarantee the ordering of the keys in the JS object/dictionary and thus you cant gurantee the order that the stylesheet objects/identifiers get passed to native in.

react_juce/core/ViewManager.cpp
JoshMarler: Just thinking about how we should give `View` access to the `ViewManager` to fetch a stylesheet object. Right now I can see two options, 

1. We add a getter for `ViewManager` to `ReactApplication` root and fetch it that way. 

2. We go for composition and have some sort of `StyleSheetFactory` or similar which `ViewManager` owns and then we pass an instance of `StyleSheetFactory` into `ViewFactory` when creating `View` instances so that it can be passed to `View`'s constructor. We might need to make that an overload used mostly internally to avoid the change to `ViewFactory` being a breaking change if possible. 

What do you think?

JoshMarler: One thought. I don't think we can avoid `ViewFactory` taking a `StyleSheetFactory` being a breaking change if we do go down that route. It's probably the nicest route I think but not sure how you feel about this one.

packages/react-juce/src/lib/Stylesheet.ts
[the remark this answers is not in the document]
JoshMarler: (Just coming back to this PR finally, well ... again) 

I actually hit this exact case hence the comment :-) Heh. I will checkout "classnames" though. That looks great.

PR #275, opened 2021-07-26
Fix issues with react-juce hot reload on recent npm versions

JoshMarler:
@nick-thompson, this fixes up some issues with our current bundle config so you can hot-reload the example app and react-juce package at the same time. There is one issue here where duktape doesn't appear to like the latest util/inspect. Can you suggest an alternative to our usage of `inspect` ? We could try some other polyfill for it now that it has been removed from webpack 5.


nick-thompson:
Yep, this looks great. Looks like two CI passes failing?

We could just remove the `util.inspect` line, or write our own `function inspect(x)` which prints all primitives as-is, prints the first few elements of an array (or all of the array if it's small enough), and prints like 2-levels deep on an object?

JoshMarler:
> Yep, this looks great. Looks like two CI passes failing?
> 
> We could just remove the `util.inspect` line, or write our own `function inspect(x)` which prints all primitives as-is, prints the first few elements of an array (or all of the array if it's small enough), and prints like 2-levels deep on an object?

Yep, needs package bump and lint to pass CI. 

Yeah I was debating just rolling our own thing. Will have to do quick grok to workout how to print object to arbitrary level of nesting. As I understood it inspect is used to avoid cycles also?

PR #276, opened 2021-07-26
Fix issues with react-juce hot reload on recent npm versions

JoshMarler:


JoshMarler:
@nick-thompson , if you're happy to give this a final check I can do the necessary npm bumps and merge. Will run lint step. Things now working under both Duktape and Hermes

nick-thompson (approved):
Yep, let's do it! Thanks @JoshMarler

nick-thompson:
@JoshMarler looks like 0.2.15 is live on npm– should be good to merge here then, yea?

JoshMarler:
> @JoshMarler looks like 0.2.15 is live on npm– should be good to merge here then, yea?

Hey,

It's working under Hermes but stupidly I'd not regenerated from cmake cache properly and something under Duktape is still broken. I'll guy this fixed over the weekend and then probably have to do a second bump. Apologies.

nick-thompson:
Ah got you, no problem! No rush

PR #278, opened 2021-08-03
Use source map (if present) to translate source file references in backtrace.

### LINKED ISSUE #277: If source map is present, use it to translate source file/line/column references in backtrace.

banburybill:
Establish framework for loading source maps if present and using them to translate references in a backtrace.

Implement this for Hermes. Use a regex to scan backtrace text and match source/line/col references. For sources where a source map is found, use the map to translate file/line/col references and substitute the translated reference into the backtrace text.

Should be easy to extend to DukTape given knowledge of DukTape backtrace formatting.

Fixes #277


nick-thompson:
Hey @banburybill apologies for not getting to this for so long.

Looking over the change, I'm not actually sure that this is code we want to introduce, because I could imagine it being somewhat finicky to maintain (regex parsing source map strings across different js engines seems dicey). My intention would be more that the uncaught exception handlers there in the EcmascriptEngine implementations should be hit quite rarely. For example, I would personally aggressively recommend using something like React's error boundaries to try to find and catch errors in js land before they ever make it to the native uncaught error handler. That way we can rely on the existing source map support in the tools we're using without having to maintain something ourselves.

Have you tried using the error boundaries or similar for preventing uncaught errors in js?

PR #279, opened 2021-08-16
Improve border properties support

### LINKED ISSUE #143: Improve support of  CSS based border-width properties
Our current `border-width` support is fairly basic and just strokes a rectangle around a view. It is difficult to add border lines and specific edges. i.e.:

------------------------------------------------------
Item 1
------------------------------------------------------
Item 2
------------------------------------------------------

We should implement an equivalent to CSS's border-width props: https://developer.mozilla.org/en-US/docs/Web/CSS/border-width

The implementation of this should follow suit with the like of ScrollView's properties where "meta-props" are split into their naive C++ equivalents. See ScrollView.ts and the unpacking of `scollbar-color` into `scrollbar-thumb-color` and `scrollbar-track-color`.

banburybill:
Break border drawing into separate drawing of top/right/bottom/left border components and add properties for border-(top|right|bottom|left)-(color|style-width) and border-(top-left|top-right|bottom-right|bottom-left)-radius. Also add styles 'dotted' and 'dashed'.

Fixes #143

### DIFF
--- a/packages/react-juce/src/lib/Backend.ts
+++ b/packages/react-juce/src/lib/Backend.ts
@@ -22,6 +22,78 @@ const cssPropsMap = allCssProps
   .filter((s) => !s.startsWith("-") && s.includes("-"))
   .reduce((acc, v) => Object.assign(acc, { [camelCase(v)]: v }), {});
 
+// known CSS border styles for React Native
+const cssBorderStyles = ["dotted", "dashed", "solid"];
+
+// Parse combination border properties.
+function parseBorderSideProp(name: string, val: string | number, info: object) {
+  invariant(
+    typeof val === "string" || typeof val === "number",
+    name + " must be a string or a number"
+  );
+
+  if (typeof val === "number")
+  {
+    info[0] = info[1] = info[2] = info[3] = val;
+    return;
+  }
+
+  const values = val.split(" ");
+
+  invariant(
+    values.length >= 1 && values.length <= 4,
+    "border-" + name + " should be a space separated string with 1 to 4 values."
+  );
+
+  switch (values.length) {
+    case 1:
+      info[0] = info[1] = info[2] = info[3] = values[0];
+      break;
+    case 2:
+      info[0] = info[2] = values[0];
+      info[1] = info[3] = values[1];
+      break;
+    case 3:
+      info[0] = values[0];
+      info[1] = info[3] = values[1];
+      info[2] = values[2];
+      break;
+    default:
+      info[0] = values[0];
+      info[1] = values[1];
+      info[2] = values[2];
+      info[3] = values[3];
+  }
+}
+
+function parseBorderProp(val: string, info: object) {
+  /*
+    Parameters can be in any order. So we need to recognised
+    which might be which.
+  */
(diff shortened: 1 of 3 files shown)


packages/react-juce/src/lib/Backend.ts
```diff
@@ -32,6 +104,12 @@ export class ViewInstance {
   constructor(id: string, type: string, props?: any, parent?: ViewInstance) {
     this._id = id;
     this._type = type;
+    this._border = {
```
nick-thompson, line 107: No need to carry this state on the ViewInstance, let's have `parseBorderProp` return a value which looks like this

packages/react-juce/src/lib/Backend.ts
```diff
@@ -144,6 +222,100 @@ export class ViewInstance {
         NativeMethods.setViewProperty(this._id, k, v);
       return;
     }
+
+    // Look for border properties and translate into our internal
+    // border-info property.
+    if (propKey.startsWith("border")) {
```
nick-thompson, line 228: I think we can probably extract a border property match/normalize utility out of this. We did something like that for colors originally, but after a bit of searching through RN's source I found this: https://github.com/facebook/react-native/tree/1465c8f3874cdee8c325ab4a4916fda0b3e43bdb/packages/normalize-color. I think we should probably replace our own color parser with their `normalize-color` and try to refactor the code below into our own `normalize-border` package. What do you think?

react_juce/core/View.cpp
```diff
@@ -12,6 +12,166 @@
...
+                                  juce::jmin(endRadius, borderHeight / 2 - pathOffset) + pathOffset,
+                                  0,
+                                  1.5 * Pi, 1.75 * Pi);
+                break;
+            }
+        }
+        else
+        {
+            // When joining no-radius corners, a browser will mitre the corners.
+            // For simplicity, we're taking an easier route - if we have a corner
```
nick-thompson, line 151: :+1:! Happy with the easier route here :)

nick-thompson (commented):
Thanks @banburybill! On the whole looks quite nice. I left a couple comments below, and I wonder as well if it'd be worth pulling some of the cpp code out of `View.cpp`? Just to keep the code well organized. What do you think?

packages/react-juce/src/lib/Backend.ts
banburybill: (Sorry for delay returning to this)

I could be not understanding something here, but don't we need to carry state around in the ViewInstance? There's no longer necessarily a single border property, but potentially a series of border- properties that may cumulatively specify what actually needs to be drawn. Surely we need to accumulate that state somewhere?

banburybill: Hmm. Guess we could have parseBorder(Side)Prop() take a border object and return a transmuted one. Still need to have something accumulating the values, though.

packages/react-juce/src/lib/Backend.ts
banburybill: Using `normalize-color` would certainly improve colour specification flexibility. No harm in putting border parsing into its own module.

